PR #3, opened 2025-03-08
chore: migrate data source interface doc

Gozala:


adr/001-data-source-interface.md
```diff
@@ -0,0 +1,155 @@
...
+// less than 32
+type Digest<T, Size extends Number = 32> = Bytes<T, Size>
+
+type Namsepace = string
+type Path = string
+type The = `${Namespace}/${Path}`
+
+// Attribute / predicate is represented as concatenation
+// of two hashes
+type Attribute = [...Digest<Namespace>, ...Digest<Path>]
```
Gozala, line 48: Mostly I've being thinking here that attributes consist of two components and I imagine scenarios where pull may be able to narrow down by namespace, but not know the path component.

adr/001-data-source-interface.md
```diff
@@ -0,0 +1,155 @@
...
+
+// Entity is the hash of arbitrary URI representing a subject in the
+// fact
+type Of = Digest<URI>
```
Gozala, line 52: I'm not sure about the whole hash of the URI thing, but I did came to conclusion that entities addressed by hash is counterintuitive & confusing in practice because point of hashes is to uniquely identify content while point of entities is to give mutable references. If you can mutate content associated with a hash that is very misleading.

Instead of reinventing a wheel, I thought URIs are good enough identifiers. As an extra benefit I could use did:key:zSpace to reference database itself. You could also do URIs like `of:hash`.

Motivation for `Digest<URI>` is basically to give entity a fixed length.

adr/001-data-source-interface.md
```diff
@@ -0,0 +1,155 @@
...
+
+type Min<T extends { length: number }, U extends { length: number }> =
+    // If T.length <= U.length ? T : U
+    T | U
+
+// Value reference is either byte represenation of value o
+type ValueReference<T> = Min<Bytes<T>, Digest<T>>
```
Gozala, line 59: I was not sure how to convey this idea, but if value you're trying is more compact in raw format perhaps using hash to refer to it is overkill. Hence this. Now that I think about it is possible to represent such values with [identity multihash](https://github.com/multiformats/multicodec/blob/master/table.csv#L2) so maybe that is a better way to go about it.

adr/001-data-source-interface.md
```diff
@@ -0,0 +1,155 @@
...
+// way to avoid having to load value, although I'm not sure it's
+// going to be a win in practice. It's only useful when value is
+// another entity in all other instances you probably will need
+// value itself. Furthermore we may want to consider encryption in
+// which case value would be encrypted and will require decryption
+// before it could be used.
+type FactAddress =
+    | [index:0, ...Entity, ...Attribute, ...Is]
+    | [index:1, ...Attribute, ...Entity, ...Is]
+    | [index:2, ...Is, ...Value, ...Entity]
```
Gozala, line 102: I imagine this roughly corresponds to keys you use, but maybe I'm misprojecting. This also more or less how [synopsys projected facts to keyspace](https://www.tldraw.com/s/v2_c_uG9xxguw-dwNAeCwJhj3U?d=v513.-1162.3626.2494.page)

adr/001-data-source-interface.md
cdata: I like this. In my practice, I did hash of unique u128 because I wasn't sure what else to do. Hash of a URI is a little more concrete to me. And yeah, something like a `did:key` or even just something dumb like `ulid:$u128` would be possible if we just said it's a URI.

adr/001-data-source-interface.md
```diff
@@ -0,0 +1,155 @@
...
+
+// @cdata - I remember considering not to capture all three
+// components in the key but just 2 out of three as third could
+// be the value, however having a hash of the value is interesting
+// way to avoid having to load value, although I'm not sure it's
+// going to be a win in practice. It's only useful when value is
+// another entity in all other instances you probably will need
+// value itself. Furthermore we may want to consider encryption in
+// which case value would be encrypted and will require decryption
+// before it could be used.
```
cdata, line 98: How do you construct the VAE index if you don't encode some reference to the value in the key?

adr/001-data-source-interface.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Context
+
+We would like to define an interface that can be used across separate query engine implementations. Specifically we would like to use [datalogia](https://github.com/gozala/datalogia)  able to use implementation implementing this intreface as data store. At the same time we would like to create another query engine implementation in Rust that would be able to use this interface for the data store.
+
+By considering two languages we can make an interface that is not too coupled with the query engine.
+
+## Decision
+
+> Interfaces are typed in [typescript] for compactness.
```
Gozala, line 15: Now I wish I did this in [wit](https://component-model.bytecodealliance.org/design/wit.html) instead.

adr/001-data-source-interface.md
cdata: Actually, the keys I came up with are meaningfully different! But, I am still learning and it was just a naive strategy without a lot of thought put into it.

EAV key:

https://github.com/meiosys/staging/blob/3765236f6c48844f916a60bcbcb31e8ee92ecfca/rust/x-query/src/fact/key.rs#L92-L96

AEV key:

https://github.com/meiosys/staging/blob/3765236f6c48844f916a60bcbcb31e8ee92ecfca/rust/x-query/src/fact/key.rs#L155-L159

VAE key:

https://github.com/meiosys/staging/blob/3765236f6c48844f916a60bcbcb31e8ee92ecfca/rust/x-query/src/fact/key.rs#L208-L212

Regardless of practicality, one thing I liked about this was that any key could be constructed from any other key.

All keys in my case implement a common trait for convenience, also:

https://github.com/meiosys/staging/blob/3765236f6c48844f916a60bcbcb31e8ee92ecfca/rust/x-query/src/fact/key.rs#L3-L87

adr/001-data-source-interface.md
Gozala: I think of it as

```sh
/of/the → {is, cause}
/the/of → {is, cause}
/is/the → {of, cause}
```

adr/001-data-source-interface.md
```diff
@@ -0,0 +1,155 @@
...
+type Is = 
+    | [type: 0]                                            // Null
+    | [type: 1, ...ValueReference<Uint8Array>]             // Raw bytes
+    | [type: 2, ...Of]                                     // Entity
+    | [type: 3, value: 1|0]                                // Boolean
+    | [type: 4, ...ValueReference<string>]                 // UTF8
+    | [type: 5, ...ValueReference<UnsignedInt>]            // ULeb128
+    | [type: 6, ...ValueReference<SignedInt>]              // Leb128
+    | [type: 7, ...ValueReference<Float>]                  // FP64
+    | [type: 8, ...ValueReference<{[key:string]:}>]        // CBOR
```
cdata, line 73: This seems fine to me. I'm grateful that it is a finite set of types (which I didn't realize until we spoke today).

adr/001-data-source-interface.md
cdata: Identity multihash SGTM

adr/001-data-source-interface.md
cdata: sickos-yes.jpeg

It's not too late!

adr/001-data-source-interface.md
cdata: Sorry, I might need more help understanding :sweat:

adr/001-data-source-interface.md
```diff
@@ -0,0 +1,155 @@
...
+// another entity in all other instances you probably will need
+// value itself. Furthermore we may want to consider encryption in
+// which case value would be encrypted and will require decryption
+// before it could be used.
+type FactAddress =
+    | [index:0, ...Entity, ...Attribute, ...Is]
+    | [index:1, ...Attribute, ...Entity, ...Is]
+    | [index:2, ...Is, ...Value, ...Entity]
+    
+type FactReader = {
```
cdata, line 104: Very similar in purpose to my `IndexKey` (linked in another comment).

I'm not sure this needs to be specified since it is kind of a convenience.

One note about `read` is in my use case it should be presumed to be async unless you think in practice it will always have data in memory by the time it is used (which seems like it isn't a given since I may know the key without ever having touched the database or storage layer).

Issue #4, opened 2025-03-26
Prolly Tree TODO


cdata:
This issue tracks discussion and remaining work to be done as we prepare our work for Local First Conference.

PR #10, opened 2025-03-28
feat: Implement basic storage crate

cdata:
This is the first PR to kick off the porting of @jsantell 's prolly tree implementation with an eye towards improvements we wish to make as sketched out in #4 .

This change carries over the basic storage facilities and factors them into their own crate. The design is very similar to @jsantell 's original design, except that the storage is made more generalized to arbitrary types (where the ancestor implementation was coupled to the prolly tree's notion of a block).

At a high-level, a `Storage` can be constructed with any combination `Encoder` and `StorageBackend`, and sufficiently compatible combinations will result in `Storage` expressing the `ContentAddressedStorage` API via a blanket trait implementation. This allows us to mix and match encoding strategies and raw storage backends rather easily.

### DIFF
--- a/rust/x-storage/src/encoder.rs
+++ b/rust/x-storage/src/encoder.rs
@@ -1,0 +1,24 @@
+use async_trait::async_trait;
+use x_common::ConditionalSync;
+
+use crate::{HashType, XStorageError};
+
+/// An [Encoder] converts to and from content-addressable bytes
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait Encoder<const HASH_SIZE: usize> {
+    /// The in-memory representation of a block
+    type Block: ConditionalSync;
+    /// The encoded byte representation of a block
+    type Bytes: AsRef<[u8]> + 'static + ConditionalSync;
+    /// The hash type produced by this [Encoder]
+    type Hash: HashType<HASH_SIZE> + ConditionalSync;
+    /// The error type produced by this [Encoder]
+    type Error: Into<XStorageError>;
+
+    /// Encode a serializable item into its referencable [`Hash`] and its bytes.
+    async fn encode(&self, block: &Self::Block) -> Result<(Self::Hash, Self::Bytes), Self::Error>;
+
+    /// Decode bytes into a `Block`.
+    async fn decode(&self, bytes: &[u8]) -> Result<Self::Block, Self::Error>;
+}
--- a/rust/x-storage/src/hash.rs
+++ b/rust/x-storage/src/hash.rs
@@ -1,0 +1,20 @@
+use x_common::ConditionalSend;
+
+/// A trait that can be implemented for types that represent a hash. A blanket
+/// "unchecked" implementation is provided for any type that matches
+/// `AsRef<[u8]>` (this might be an antipattern; more investigation required).
+pub trait HashType<const SIZE: usize>: Clone + ConditionalSend {
+    /// Get the raw bytes of the hash
+    fn bytes(&self) -> [u8; SIZE];
+}
+
+impl<const SIZE: usize, T> HashType<SIZE> for T
+where
+    T: Clone + AsRef<[u8]> + ConditionalSend,
+{
+    fn bytes(&self) -> [u8; SIZE] {
+        let mut bytes = [0u8; SIZE];
+        bytes.copy_from_slice(self.as_ref());
+        bytes
+    }
+}
--- a/rust/x-storage/src/storage/backend.rs
+++ b/rust/x-storage/src/storage/backend.rs
@@ -1,0 +1,25 @@
+use async_trait::async_trait;
+use x_common::{ConditionalSend, ConditionalSync};
+
+use crate::XStorageError;
+
+mod memory;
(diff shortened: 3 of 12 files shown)


rust/x-storage/src/encoder.rs
```diff
@@ -0,0 +1,24 @@
...
+use x_common::ConditionalSync;
+
+use crate::{HashType, XStorageError};
+
+/// An [Encoder] converts to and from content-addressable bytes
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait Encoder<const HASH_SIZE: usize> {
+    /// The in-memory representation of a block
+    type Block: ConditionalSync;
```
Gozala, line 11: I think it would help if `ConditionalSync` had some comment explaining what the implication of "Conditional". My current interpretation is that it's thread safe, transferable and can target wasm.

https://github.com/meiosys/staging/blob/fc1897155af626b92f05173db392664ab2ec81b3/rust/x-common/src/sync.rs#L13-L33

rust/x-storage/src/hash.rs
```diff
@@ -0,0 +1,20 @@
+use x_common::ConditionalSend;
+
+/// A trait that can be implemented for types that represent a hash. A blanket
+/// "unchecked" implementation is provided for any type that matches
+/// `AsRef<[u8]>` (this might be an antipattern; more investigation required).
+pub trait HashType<const SIZE: usize>: Clone + ConditionalSend {
```
Gozala, line 6: Does blanket implementation simply uses first `SIZE` bytes ?

rust/x-storage/src/storage/backend.rs
```diff
@@ -0,0 +1,25 @@
...
+use crate::XStorageError;
+
+mod memory;
+pub use memory::*;
+
+/// A [StorageBackend] is a facade over some generalized storage substrate that
+/// is capable of storing and/or retrieving values by some key
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait StorageBackend {
```
Gozala, line 13: Is this meant to be an abstraction for a blob store storing prolly tree nodes or a kv abstraction implemented by a prolly tree itself ?

It makes sense for former, but for the later I think we care about range scans a lot more than individual key lookups.

rust/x-storage/src/storage/content_addressed.rs
```diff
@@ -0,0 +1,63 @@
...
+impl<const HASH_SIZE: usize, Bl, By, H, Ee, Es, T> ContentAddressedStorage<HASH_SIZE> for T
+where
+    H: HashType<HASH_SIZE> + ConditionalSync,
+    Bl: ConditionalSync,
+    By: AsRef<[u8]> + 'static + ConditionalSync,
+    Ee: Into<XStorageError>,
+    Es: Into<XStorageError>,
+    T: Encoder<HASH_SIZE, Block = Bl, Bytes = By, Hash = H, Error = Ee>
+        + StorageBackend<Key = H, Value = By, Error = Es>
+        + ConditionalSync,
```
Gozala, line 39: It took me a while to figure out what the acronyms were, I think it would have being a lot simpler if full names were used

```suggestion
impl<const HASH_SIZE: usize, Block, Bytes, Hash, EncoderError, StorageError, T> ContentAddressedStorage<HASH_SIZE> for T
where
    Hash: HashType<HASH_SIZE> + ConditionalSync,
    Block: ConditionalSync,
    Bytes: AsRef<[u8]> + 'static + ConditionalSync,
    EncoderError: Into<XStorageError>,
    StorageError: Into<XStorageError>,
    T: Encoder<HASH_SIZE, Block = Block, Bytes = Bytes, Hash = Hash, Error = EncoderError>
        + StorageBackend<Key = Hash, Value = Bytes, Error = StorageError>
        + ConditionalSync,
```

rust/x-storage/src/encoder.rs
cdata: Thanks for calling it out. This is an annoying but necessary Rust boilerplate when you target both browser-facing Wasm and other targets (and it will disappear as `async fn` in trait becomes more mature), so I tend to gloss over it.

rust/x-storage/src/hash.rs
cdata: Yes, it fills a `[u8; SIZE]` array; if there aren't enough bytes the remaining are zero.

The alternative would be to have a checked conversion, which is probably the more sane thing to do. But, this blanket implementation is nice because it makes it trivial to experiment with different byte containers in the encoder.

rust/x-storage/src/storage/backend.rs
cdata: It's the former

rust/x-storage/src/storage/content_addressed.rs
cdata: I love this feedback because I prefer readable names myself; it is stylistically common in Rust to have abbreviated names for generic bounds and non-abbreviated names for associated types (the former is not typically written by the API consumer, but the later is).

I'm happy to align on the style that both are full words. I strongly dislike reading abbreviations.

PR #11, opened 2025-03-28
feat: File system and IDB storage backends

cdata:
This change is a quick follow-up to #10 which ports the file system and IndexedDB storage backend implementations from the ancestor prolly tree implementation. A basic test is included which demonstrates changing the backend under test depending on the Rust target. Also, generic bounds are no longer abbreviated per feedback in #10 .


rust/x-storage/src/storage/backend/indexeddb.rs
```diff
@@ -0,0 +1,123 @@
...
+            .await
+            .map_err(|error| XStorageError::StorageBackend(format!("{error}")))?;
+
+        Ok(())
+    }
+
+    async fn get(&self, key: &Self::Key) -> Result<Option<Self::Value>, Self::Error> {
+        let tx = self
+            .db
+            .transaction(&[&self.store_name], TransactionMode::ReadWrite)
```
Gozala: Why not readonly transaction here ?

cdata: Good catch! Is a mistake

PR #13, opened 2025-04-03
feat: Port of ranked-prolly-tree crate

cdata:
This change ports over the `ranked-prolly-tree` crate originally designed and implemented by @jsantell . Largely the design remains the same. The following substantive changes were made:

- `Tree` and `Node` make use of the refactored `x-storage` constructs
- `Tree` and `Node` are generic over their `Distribution`; `GeometricDistribution` is the only implementation, but others may be added in the future
- Some domain concepts are renamed (e.g., `NodeRef` -> `Reference`, `Adoptable` -> `Adopter`)
- Additional documentation and tests have been added

In addition to the port, the `x-query` crate has been revised to work under the `wasm32-unknown-unknown` target. All tests across the workspace now run and pass under both native and `wasm32-unknown-unknown` targets.

### DIFF
--- a/rust/x-prolly-tree/README.md
+++ b/rust/x-prolly-tree/README.md
@@ -1,0 +1,8 @@
+# x-prolly-tree
+
+This crate provides a key-value store implemented as a [prolly tree], utilizing
+a flexible content-addressed block storage backend. This prolly tree is designed
+to be the foundation of a passive database, providing on demand partial
+replication.
+
+[prolly tree]: https://www.dolthub.com/blog/2024-03-03-prolly-trees/
--- a/rust/x-prolly-tree/src/adopter.rs
+++ b/rust/x-prolly-tree/src/adopter.rs
@@ -1,0 +1,92 @@
+use crate::{Block, Entry, KeyType, Node, Reference, XProllyTreeError};
+use async_trait::async_trait;
+use nonempty::NonEmpty;
+use x_common::ConditionalSync;
+use x_storage::{ContentAddressedStorage, HashType};
+
+/// A helper trait implemented by [`Entry`], [`Reference`] and [`Node`] to
+/// create new [`Node`]s.
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait Adopter<const BRANCH_FACTOR: u32, const HASH_SIZE: usize, Key, Value, Hash>:
+    Sized
+where
+    Key: KeyType,
+    Value: Clone + ConditionalSync,
+    Hash: HashType<HASH_SIZE>,
+{
+    /// Adopt a collection of `children` into a new [`Node`]. Children data must
+    /// be ordered and follow rank rules.
+    async fn adopt(
+        children: NonEmpty<Self>,
+        storage: &mut impl ContentAddressedStorage<
+            HASH_SIZE,
+            Block = Block<HASH_SIZE, Key, Value, Hash>,
+            Hash = Hash,
+        >,
+    ) -> Result<Node<BRANCH_FACTOR, HASH_SIZE, Key, Value, Hash>, XProllyTreeError>;
+}
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+impl<const BRANCH_FACTOR: u32, const HASH_SIZE: usize, Key, Value, Hash>
+    Adopter<BRANCH_FACTOR, HASH_SIZE, Key, Value, Hash> for Entry<Key, Value>
+where
+    Key: KeyType + 'static,
+    Value: Clone + ConditionalSync,
+    Hash: HashType<HASH_SIZE>,
+{
+    async fn adopt(
+        children: NonEmpty<Self>,
+        storage: &mut impl ContentAddressedStorage<
+            HASH_SIZE,
+            Block = Block<HASH_SIZE, Key, Value, Hash>,
+            Hash = Hash,
+        >,
+    ) -> Result<Node<BRANCH_FACTOR, HASH_SIZE, Key, Value, Hash>, XProllyTreeError> {
+        Node::segment(children, storage).await
+    }
+}
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+impl<const BRANCH_FACTOR: u32, const HASH_SIZE: usize, Key, Value, Hash>
+    Adopter<BRANCH_FACTOR, HASH_SIZE, Key, Value, Hash> for Reference<HASH_SIZE, Key, Hash>
+where
+    Key: KeyType + 'static,
+    Value: Clone + ConditionalSync,
+    Hash: HashType<HASH_SIZE>,
+{
+    async fn adopt(
+        children: NonEmpty<Self>,
+        storage: &mut impl ContentAddressedStorage<
+            HASH_SIZE,
+            Block = Block<HASH_SIZE, Key, Value, Hash>,
+            Hash = Hash,
+        >,
+    ) -> Result<Node<BRANCH_FACTOR, HASH_SIZE, Key, Value, Hash>, XProllyTreeError> {
+        Node::branch(children, storage).await
+    }
+}
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+impl<const BRANCH_FACTOR: u32, const HASH_SIZE: usize, Key, Value, Hash>
+    Adopter<BRANCH_FACTOR, HASH_SIZE, Key, Value, Hash>
+    for Node<BRANCH_FACTOR, HASH_SIZE, Key, Value, Hash>
+where
+    Key: KeyType + 'static,
+    Value: Clone + ConditionalSync,
+    Hash: HashType<HASH_SIZE>,
+{
+    async fn adopt(
+        children: NonEmpty<Self>,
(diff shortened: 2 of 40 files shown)


rust/x-prolly-tree/src/adopter.rs
```diff
@@ -0,0 +1,92 @@
...
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait Adopter<const BRANCH_FACTOR: u32, const HASH_SIZE: usize, Key, Value, Hash>:
+    Sized
+where
+    Key: KeyType,
+    Value: Clone + ConditionalSync,
+    Hash: HashType<HASH_SIZE>,
+{
+    /// Adopt a collection of `children` into a new [`Node`]. Children data must
+    /// be ordered and follow rank rules.
```
Gozala, line 19: Not sure what rank rules are, perhaps pointer here would be useful.

rust/x-prolly-tree/README.md
```diff
@@ -0,0 +1,8 @@
+# x-prolly-tree
```
Gozala, line 1: What does `x` stand for ?

rust/x-prolly-tree/src/error.rs
```diff
@@ -0,0 +1,40 @@
...
+    #[error("Invalid tree construction: {0}")]
+    InvalidConstruction(String),
+
+    /// There was a problem when accessing storage
+    #[error("Storage error: {0}")]
+    Storage(XStorageError),
+
+    /// A required block is missing from storage
+    #[error("Block not found in storage: {0}")]
+    MissingBlock(String),
```
Gozala, line 21: 😭

rust/x-prolly-tree/src/error.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    /// The tree did not match the expected shape
+    #[error("Tree did not match expected shape: {0}")]
+    UnexpectedTreeShape(String),
```
Gozala, line 25: Perhaps adding some info showing path where tree is misshaped would be good idea.

rust/x-prolly-tree/src/error.rs
Gozala: Perhaps this should include a hash of the block that is missing ?

rust/x-prolly-tree/src/node.rs
```diff
@@ -0,0 +1,532 @@
...
+use futures_core::Stream;
+use nonempty::NonEmpty;
+use x_common::ConditionalSync;
+use x_storage::{ContentAddressedStorage, HashType};
+
+use crate::{Block, Entry, KeyType, Rank, Reference, XProllyTreeError};
+
+/// Primary representation of tree nodes.
+///
+/// The common error type used by this crate Each [`Node`] stores its children
```
Gozala, line 14: I think beginning of the sentence "The common error type used by this crate" is perhaps leftover from something that was here before or otherwise it misses a part.

rust/x-prolly-tree/src/node.rs
```diff
@@ -0,0 +1,532 @@
...
+    /// Whether this node is a segment.
+    pub fn is_segment(&self) -> bool {
+        self.block.is_segment()
+    }
+
+    /// Create a new branch [`Node`] given [`Reference`] children, storing
+    /// the new [`Node`] in the provided [`ContentAddressedStorage`]
+    pub async fn branch<Storage>(
+        children: NonEmpty<Reference<HASH_SIZE, Key, Hash>>,
+        storage: &mut Storage,
```
Gozala, line 51: Just wondering why storage is passed as opposed to be held by `Node` itself. It seems to imply that you could use different stores for different branches which seems counter intuitive. Perhaps you're supposed to pass same storage, but it would be good idea to leave some explanation behind as code comment.

rust/x-prolly-tree/src/node.rs
```diff
@@ -0,0 +1,532 @@
...
+        let block = Block::segment(children);
+        let bound = block.upper_bound().clone();
+        let hash = storage.write(&block).await.map_err(|error| error.into())?;
+        let reference = Reference::new(bound, hash);
+
+        Ok(Node { block, reference })
+    }
+
+    /// Hydrates a [`Node`] from [`ContentAddressedStorage`] given a [`Reference`].
+    pub async fn from_reference<Storage>(
```
Gozala, line 90: This may be conventional naming in rust, but I would personally called it `load` or something along those lines.

rust/x-prolly-tree/src/node.rs
```diff
@@ -0,0 +1,532 @@
...
+        let Some(block) = storage.read(&hash).await.map_err(|error| error.into())? else {
+            return Err(XProllyTreeError::MissingBlock(format!(
+                "#{}",
+                hash.bytes().to_base58()
+            )));
+        };
+        let reference = Reference::new(block.upper_bound().clone(), hash);
+
+        Ok(Node { block, reference })
+    }
```
Gozala, line 130: I'm not yet sure if there is a good rational here for having both `from_hash` and `from_reference`, but naively I feel like perhaps everything in this layer should just work with Reference's, perhaps even store which could instead of taking a hash could take anything that has `hash(&self) → &Hash`.

rust/x-prolly-tree/src/node.rs
```diff
@@ -0,0 +1,532 @@
...
+        &self.reference.hash()
+    }
+
+    /// Return all [`Entry`]s from this [`Node`] into a [`Entry`] collection.
+    ///
+    /// The result is an error if this is not a segment [`Node`].
+    pub fn into_entries(self) -> Result<NonEmpty<Entry<Key, Value>>, XProllyTreeError> {
+        if !self.is_segment() {
+            return Err(XProllyTreeError::IncorrectTreeAccess(
+                "Cannot convert branch into entries".into(),
```
Gozala, line 149: Ah I see know what incorrect tree access meant which I was confused about perhaps more clear error names could be used `BranchIsNotASegment`, I imagine there's also `SegmentIsNotABranch`.

rust/x-prolly-tree/src/node.rs
```diff
@@ -0,0 +1,532 @@
...
+    ) -> Result<Option<Entry<Key, Value>>, XProllyTreeError>
+    where
+        Storage: ContentAddressedStorage<
+                HASH_SIZE,
+                Block = Block<HASH_SIZE, Key, Value, Hash>,
+                Hash = Hash,
+            >,
+    {
+        #[allow(unused_assignments)]
+        let mut current_node_holder: Option<Self> = None;
```
Gozala, line 171: I'm sure this has something to do with borrowing stuff, but still a comment might help.

rust/x-prolly-tree/src/node.rs
```diff
@@ -0,0 +1,532 @@
...
+                    current_node = current_node_holder.as_ref().unwrap();
+                }
+                false => return current_node.entry_by_key(key),
+            }
+        }
+    }
+
+    /// Inserts a new [`Entry`] into the hierarchy represented by this [`Node`]
+    /// as its root. If successful, returns the new root [`Node`] representing
+    /// the hierarchy.
```
Gozala, line 190: Perhaps there should be a separate trait representing a Tree root where this method lives ? Seems like a footgun to have a method on node that you should only call on the root.

rust/x-prolly-tree/src/node.rs
```diff
@@ -0,0 +1,532 @@
...
+                HASH_SIZE,
+                Block = Block<HASH_SIZE, Key, Value, Hash>,
+                Hash = Hash,
+            >,
+    {
+        if !self.is_branch() {
+            return Err(XProllyTreeError::IncorrectTreeAccess(
+                "Cannot descend through segment".into(),
+            ));
+        }
```
Gozala, line 449: Why not have separate traits for branch nodes and segment nodes so this won't be necessary ?

rust/x-prolly-tree/src/tree.rs
```diff
@@ -0,0 +1,188 @@
...
+        }
+        Ok(())
+    }
+
+    /// Returns an async stream over all entries.
+    pub fn stream<'a>(
+        &'a self,
+    ) -> impl Stream<Item = Result<Entry<Key, Value>, XProllyTreeError>> + 'a {
+        self.stream_range(..)
+    }
```
Gozala, line 129: Seems like an antipattern so perhaps not having this method to iterate a full range is better.

rust/x-prolly-tree/src/tree.rs
```diff
@@ -0,0 +1,188 @@
...
+            let entries = collection
+                .into_iter()
+                .map(|(key, value)| {
+                    let node = Entry { key, value };
+                    let rank = Distribution::rank(&node.key);
+                    (node, rank)
+                })
+                .collect();
+            let entries = NonEmpty::from_vec(entries).ok_or_else(|| {
+                XProllyTreeError::InvalidConstruction("Tree must have at least one child".into())
```
Gozala, line 166: We have an empty tree representation so it seems odd to not be able to create an empty tree from empty collection.

rust/x-prolly-tree/tests/tree.rs
```diff
@@ -0,0 +1,206 @@
...
+}
+
+#[cfg(target_arch = "wasm32")]
+use wasm_bindgen_test::wasm_bindgen_test;
+#[cfg(target_arch = "wasm32")]
+wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_dedicated_worker);
+
+#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+#[cfg_attr(not(target_arch = "wasm32"), tokio::test)]
+async fn basic_set_and_get() -> Result<()> {
```
Gozala, line 18: Would be good to test random insertion order also.

rust/x-prolly-tree/src/error.rs
cdata: It does, via the inner string

PR #18, opened 2025-04-11
notes: structural datalog idea

Gozala:


adr/003-structural-triples.md
```diff
@@ -0,0 +1,311 @@
...
+  :composer artist
+  :name string
+  :duration duration
+  :album album
+  :rating rate
+  :genre string
+  :year uint8)
+```
+
+Above rule now defines a logic by which `#ba4jcao2nbln4ui7v2ff` can be derived from the `song` records. In this case it a simply renames `title`
```
Gozala:
```suggestion
```

Deleting leftover from previous example

notes/structural-triples.md
```diff
@@ -0,0 +1,310 @@
...
+type TrackAlbum = {
+  album: Album
+}
+
+type TrackV2 = Track & TrackAlbum
+```
+
+🤔 However it is good to ponder how things would have turned out if we have started track that accounted for `album`  from the begining, most likely all tracks would have had `artist`, `title`, `album`, and `duration`, but than how would we have stored information about songs that we did not know an `album` or a `duration` for ?
+
+#### Subetyping
```
just-be-dev:
```suggestion
#### Subtyping
```

notes/structural-triples.md
```diff
@@ -0,0 +1,310 @@
...
+
+In type systems you could define a supertype containing subset of the subtype in fact that is what we end up defining in our `song` rule in datalog example. In other words we broke down model into smaller parts and then composed `track` from them.
+
+That is effectively what triples do they just take it to a logical extreme. This effectively reduces need for schema migration because our tables end up just maps of `entity -> value` where name of the table serves as an `attribute`.
+
+Everything else is just composition (or a query over those tables), therefor instead of migartion we can simply define a new query.
```
just-be-dev:
```suggestion
Everything else is just composition (or a query over those tables), therefor instead of migration we can simply define a new query.
```

notes/structural-triples.md
```diff
@@ -0,0 +1,310 @@
...
+
+
+#### Optional fields
+
+Alterantive way to handle our delema would be to make all fields optional _(explains why that is a default in SQL)_ that way you can always have a partial information. In some way this in almost what semantic triples are with a slight difference that record with no fields is possible in a design with all optional fields and impossible in semantic triples, but then again what is the point of entity if it has no relation to anything else ?
```
just-be-dev:
```suggestion
Alternative way to handle our dilemma would be to make all fields optional _(explains why that is a default in SQL)_ that way you can always have a partial information. In some way this in almost what semantic triples are with a slight difference that record with no fields is possible in a design with all optional fields and impossible in semantic triples, but then again what is the point of entity if it has no relation to anything else ?
```

PR #42, opened 2025-05-05
feat: Switch to serde

cdata:
This change swaps out our custom encoding facility for `serde`. The old design was tightly coupled to specific structures, which left us with the requirement of defining (de)serialization at every point of interest. Serde enables us to decouple things somewhat - all data structures can derive serializability, and the encoder can now be generic over the serde traits. An IPLD-compatible CBOR encoder has been implemented, and replaces the previous bespoke encoder.

Serde isn't without trade-offs, but this will allow us to defer worrying about how encoding should work for a little while.

A bonus inclusion in this change is a `StorageBackend` layer that can perform brotli compression and decompression on blobs.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
-    "rust-analyzer.cargo.target": null,
+    "rust-analyzer.cargo.target": "wasm32-unknown-unknown",
     "rust-analyzer.procMacro.ignored": {
         "async-trait": [
             "async_trait"
--- a/rust/dialog-storage/src/storage/compress.rs
+++ b/rust/dialog-storage/src/storage/compress.rs
@@ -8,15 +8,17 @@ use crate::DialogStorageError;
 
 use super::StorageBackend;
 
+const BUFFER_SIZE: usize = 4096;
+
 /// A layer over a [`StorageBackend`] that brotli-compresses incoming writes,
 /// and decompresses outgoing reads.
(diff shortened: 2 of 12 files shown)


.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-    "rust-analyzer.cargo.target": "wasm32-unknown-unknown",
+    "rust-analyzer.cargo.target": null,
     "rust-analyzer.procMacro.ignored": {
```
Gozala: I assume this was unintentional change ?

rust/dialog-storage/src/storage/compress.rs
```diff
@@ -0,0 +1,69 @@
...
+    type Key = Backend::Key;
+    type Value = Backend::Value;
+    type Error = DialogStorageError;
+
+    async fn set(&mut self, key: Self::Key, value: Self::Value) -> Result<(), Self::Error> {
+        let mut compressed = Cursor::new(Vec::new());
+
+        {
+            let mut writer = CompressorWriter::new(&mut compressed, 4096, 9, 20);
+            writer.write(value.as_ref()).map_err(|error| {
```
Gozala, line 48: I'd suggest defining these as constants so they are referenced by name and provide some insight what they stand for.

cdata: Addressed in #43

PR #48, opened 2025-05-09
feat: port experimental package to ts

Gozala:
I think this works now, I have not tried to publish to verify it all works but will try in a moment.


typescript/dialog-experimental/src/artifacts.ts
```diff
@@ -0,0 +1,252 @@
...
+ */
+class ArtifactsStore implements Querier, Transactor {
+  /**
+   * Open an artifacts store
+   * @param address The store address configuration
+   * @returns A task that resolves to a Querier and Transactor interface
+   */
+  static *open(address: Address) {
+    if (!initialized) {
+      yield* Task.wait(init())
```
cdata, line 59: I _think_ if we can assume top-level await then the Wasm can be auto-init'd in a wrapper module. No change requested, but perhaps something to think about.

Gozala: I had that assumption not too long ago, and it broke things in browsers. Unless you feel very strongly about this I suggest we just avoid potential issues in various runtimes.

.github/workflows/test.yml
```diff
@@ -4,16 +4,16 @@ on:
...
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
   lint_nix:
-    name: 'Lint Nix'
-    runs-on: 'ubuntu-latest'
+    name: "Lint Nix"
+    runs-on: "ubuntu-latest"
```
cdata: Do you have a formatter configuration that we can both use so that these files don't thrash back and forth across PRs?

typescript/dialog-experimental/src/artifacts.ts
cdata: Like I said: no change requested.

.github/workflows/test.yml
```diff
@@ -22,31 +22,68 @@ jobs:
...
+        with:
+          project: typescript/${{matrix.package}}/tsconfig.json
+      - name: Build
+        run: pnpm --if-present --dir typescript/${{matrix.package}} build
+      - name: Test (Node)
+        run: pnpm run --if-present --dir typescript/${{matrix.package}} test:node
+      - name: Test (Web)
+        run: pnpm run --if-present --dir typescript/${{matrix.package}} test:web
+
+  wasm_artifacts:
```
cdata: TL;DR this should be `npm_packages`

This was previously the entrypoint for running `npm test`, and the substance of the job remains the same. My preference would be for a single task to embody running all NPM-style package tests (and the newly introduced job will almost certainly want to depend on these artifacts down the road if not today). Also, there are already tests of the wasm artifacts (see the `wasm32-unknown-unknown` matrix variations under `rust_crates`).

.github/workflows/test.yml
```diff
@@ -22,31 +22,68 @@ jobs:
...
+        run: pnpm install --dir typescript/${{matrix.package}}
+      - run: pnpm run --if-present --dir typescript/${{matrix.package}} install:vendor
+      - name: Typecheck
+        uses: gozala/typescript-error-reporter-action@v1.0.8
+        with:
+          project: typescript/${{matrix.package}}/tsconfig.json
+      - name: Build
+        run: pnpm --if-present --dir typescript/${{matrix.package}} build
+      - name: Test (Node)
+        run: pnpm run --if-present --dir typescript/${{matrix.package}} test:node
```
cdata: I'm hesitant about this for a couple of reasons:

- The Node.js use case hasn't been spelled out
- I know with certainty that the Wasm artifacts do not work under Node.js today
- If we were not testing Node.js, we could align on the test runner that is already in use

.github/workflows/test.yml
Gozala: Added config to vscode asking it to use single quotes.

.github/workflows/test.yml
Gozala: It only runs `pnpm run test:node` if package has `test:node` if package does not this will just skip it. This way packages can opt-in or out without having to update workflows. If you still dislike this I can remove this

.github/workflows/test.yml
Gozala: Although it looks like .yml file was inconsistent in it's use of quotes. Do you mind installing prettier which will format things. Otherwise I can ask mine to ignore .yml files in this project.

.github/workflows/test.yml
Gozala: reverted rename

.github/workflows/test.yml
cdata: Let's have an F2F discussion about it, I'm worried about nuances being missed in text.

typescript/dialog-experimental/tsconfig.json
```diff
@@ -40,17 +40,17 @@
     // "allowArbitraryExtensions": true,                 /* Enable importing files with any extension, provided a declaration file is present. */
     // "noResolve": true,                                /* Disallow 'import's, 'require's or '<reference>'s from expanding the number of files TypeScript should add to a project. */
     /* JavaScript Support */
-    "allowJs": true /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */,
-    "checkJs": true /* Enable error reporting in type-checked JavaScript files. */,
+    // "allowJs": true /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */,
+    // "checkJs": true /* Enable error reporting in type-checked JavaScript files. */,
```
cdata, line 44: Can you reconcile this with [the existing tsconfig.json](https://github.com/dialog-db/staging/blob/main/rust/dialog-artifacts/javascript_tests/tsconfig.json) and create a common one that both packages make use of?

(there is nothing precious about the existing one, I just want to avoid having two)

typescript/dialog-experimental/package.json
```diff
@@ -1,18 +1,19 @@
 {
   "name": "@dialog-db/experimental",
-  "version": "0.0.4",
+  "version": "0.0.7",
```
cdata, line 3: I briefly looked into pnpm build support under Nix and there is a combination of three things:

- "official" support is not fully baked https://github.com/NixOS/nixpkgs/issues/317927
- "community" support is updated infrequently ([1](https://github.com/Ninlives/pnpm2nix), [2](https://github.com/hallettj/pnpm-lock-export), [3](https://github.com/nix-community/pnpm2nix))
- [pnpm's lockfile format seems to change frequently](https://github.com/NixOS/nixpkgs/issues/231513#issuecomment-2062028180) (probably causing headaches for tool maintainers)

So, I'm inclined to say that we should align on `package-lock.json` (and, we should make sure we check it in).

cdata: Also, parsing this discussion: https://github.com/orgs/pnpm/discussions/3367

It seems like there is available tooling to go from `package-lock.json` to `pnpm-lock.yaml`, but not the other way around.

I would like to propose that we assume a vanilla Node.js/NPM toolchain for automation's sake, and enable bespoke tooling as a local development detail.

cdata: I did find a couple of fly-by-night options ([1](https://github.com/ndom91/pnpm-lock-to-package-lock), [2](https://github.com/ndom91/pnpm-lock-to-package-lock)) that go the other way....

.github/workflows/test.yml
cdata: I'm happy to pick a formatter. My preference is that formatting rules are enforced in CI.

PR #49, opened 2025-05-09
feat: react hook

Gozala:
Implements a react hook for ease of integration.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -182,7 +182,7 @@
             src = ./typescript/dialog-experimental/.;
 
             # npmDepsHash = lib.fakeHash;
-            npmDepsHash = "sha256-dOk2f9/l+PDHQAwolCip7zgVsmmjO7wasUBfCQCNi9U=";
+            npmDepsHash = "sha256-qcnrYVltgUUXWQRFT9TzYfHOcdUswfEI/j6WkZ41HmU=";
 
             nativeBuildInputs = common-build-inputs "stable" ++ [
               playwright-driver
--- a/typescript/dialog-experimental/package.json
+++ b/typescript/dialog-experimental/package.json
@@ -38,7 +38,6 @@
   },
   "dependencies": {
     "@dialog-db/query": "^0.11.1",
-    "@vitejs/plugin-react-refresh": "1.3.6",
     "react": ">=19"
   },
   "devDependencies": {
@@ -49,7 +48,7 @@
     "@types/mocha": "^10.0.10",
     "@types/react": "19.1.3",
     "@types/react-dom": "^19.1.5",
-    "@vitejs/plugin-react-refresh": "^1.3.6",
+    "@vitejs/plugin-react-swc": "^3.9.0",
     "@web/test-runner": "^0.20.1",
     "c8": "8.0.1",
     "mocha": "^11.2.2",
--- a/typescript/dialog-experimental/web-test-runner.config.mjs
+++ b/typescript/dialog-experimental/web-test-runner.config.mjs
@@ -1,10 +1,13 @@
 import { vitePlugin } from '@remcovaes/web-test-runner-vite-plugin'
-import reactRefresh from '@vitejs/plugin-react-refresh'
+import react from '@vitejs/plugin-react-swc'
+
+react.preambleCode
 export default {
   plugins: [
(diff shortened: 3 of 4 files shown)


typescript/dialog-experimental/test/react.spec.web.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  'test hook with provider': async (assert) => {
+    const db = await Artifacts.open({ name: 'test-provider' })
+    const Todo = fact({
+      title: String,
+      done: Boolean,
+    })
+
+    function TodoList() {
+      const todos = useQuery(Todo())
```
Gozala: @bfollington You could also parametrize rules e.g. instead we could have said 

```ts
useQuery(Todo.match({ done: false }))
```

rust/dialog-artifacts/src/artifacts.rs
Gozala, line 1: Changes in this file were merged from #65 and will go away when it lands.

typescript/dialog-experimental/package.json
```diff
@@ -23,35 +24,47 @@
...
-    "@open-wc/testing": "^4.0.0",
-    "@web/dev-server-esbuild": "^1.0.4",
+    "@types/react": "19.1.3",
+    "@types/react-dom": "^19.1.5",
+    "@vitejs/plugin-react-refresh": "^1.3.6",
     "@web/test-runner": "^0.20.1",
     "c8": "8.0.1",
     "mocha": "^11.2.2",
-    "ts-mocha": "^11.1.0",
+    "playwright-test": "^14.1.9",
```
cdata, line 55: Are we making use of it?

typescript/dialog-experimental/web-test-runner.config.mjs
```diff
@@ -1,6 +1,10 @@
-import { esbuildPlugin } from '@web/dev-server-esbuild';
-
+import { vitePlugin } from '@remcovaes/web-test-runner-vite-plugin'
+import reactRefresh from '@vitejs/plugin-react-refresh'
```
cdata: My understanding is that this plugin was deprecated (it hasn't been updated in 4 years) and replaced by https://www.npmjs.com/package/@vitejs/plugin-react (which is what I used in #53 ).

flake.nix
```diff
@@ -128,6 +128,7 @@
             installPhase = ''
               mkdir -p $out/@dialog-db
               cp -r ./rust/dialog-artifacts/pkg $out/@dialog-db/dialog-artifacts
+              rm $out/@dialog-db/dialog-artifacts/.gitignore
```
cdata, line 131: It's not that important, but since this is happening in an ephemeral build workspace you can safely `rm ./.gitignore` _before_ copying into `$out` and have the same effect.

typescript/dialog-experimental/web-test-runner.config.mjs
Gozala: Yeah except it caused this problem on first run and I was unable to figure out why or how to fix it. https://github.com/dialog-db/staging/pull/49#issuecomment-2888835819

while this plugin seems to have being superseded it actually works without mentioned problem

typescript/dialog-experimental/package.json
Gozala: It has been helpful for me to debug things via `npm:test:debug` script that I run manually when needed

flake.nix
Gozala: @cdata do you mean something like this

```diff
diff --git a/flake.nix b/flake.nix
index cb32393..082acf1 100644
--- a/flake.nix
+++ b/flake.nix
@@ -124,11 +124,11 @@
               export HOME=`pwd`
 
               wasm-pack build --release --scope dialog-db --target web --weak-refs -m no-install ./rust/dialog-artifacts
+              rm ./rust/dialog-artifacts/.gitignore
             '';
             installPhase = ''
               mkdir -p $out/@dialog-db
               cp -r ./rust/dialog-artifacts/pkg $out/@dialog-db/dialog-artifacts
-              rm $out/@dialog-db/dialog-artifacts/.gitignore
             '';
 
             nativeBuildInputs = common-build-inputs "stable";
```

PR #52, opened 2025-05-12
feat: CSV export and import

cdata:
This change implements CSV import / export. This should be helpful when constructing demos, sparing us from creating data from scratch every time.

While implementing this feature, I discovered from bugs related to prolly tree traversal and how the value index was constructed. These have been fixed and tests have been added to cover them.

Additionally:

- The Rust implementation of `Artifacts` now receives an async stream when committing facts
  - An "ext" trait has been added that exposes the iterator-based API (and internally converts to a stream)
- Serialization now has different characteristics depending on where it occurs, in some cases serialization leads to a UTF-8-compatible output, in others a compact binary output

### DIFF
--- a/rust/dialog-artifacts/src/artifacts.rs
+++ b/rust/dialog-artifacts/src/artifacts.rs
@@ -261,6 +261,11 @@ where
         let value_index = self.value_index.clone();
 
         try_stream! {
+            // We clone to "pin" the indexes at a version for the lifetime of the stream
+            let entity_index = entity_index.read().await.clone();
+            let attribute_index = attribute_index.read().await.clone();
+            let value_index = value_index.read().await.clone();
+
             let entity = selector.entity();
             let attribute = selector.attribute();
             let value = selector.value();
@@ -274,8 +279,7 @@ where
                     end = end.set_attribute(attribute.into());
                 }
 
-                let index = entity_index.read().await;
-                let stream = index.stream_range(Range { start, end });
+                let stream = entity_index.stream_range(Range { start, end });
 
                 tokio::pin!(stream);
 
@@ -304,8 +308,7 @@ where
                     end = end.set_attribute(attribute.into());
                 }
 
-                let index = value_index.read().await;
-                let stream = index.stream_range(Range { start, end });
+                let stream = value_index.stream_range(Range { start, end });
 
                 tokio::pin!(stream);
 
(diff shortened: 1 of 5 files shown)


rust/dialog-artifacts/src/artifacts.rs
```diff
@@ -103,6 +109,63 @@ where
         }
     }
 
+    #[cfg(feature = "csv")]
+    /// Export the database in CSV format. Each row will consist of the data
+    /// from a single [`Artifact`], laid out in {attribute, entity, value,
+    /// cause} order.
+    // TODO: It would be cool if we could export (and maybe even import?) based
+    // on pattern matching.
```
Gozala, line 117: As in provide a selector and get CSV of the matches ?

Gozala: It feels like perhaps `select` should return `Selector` which could then have `archive` function that could be used to serialize it to CSV.

rust/dialog-artifacts/src/artifacts.rs
```diff
@@ -103,6 +109,63 @@ where
...
+            }
+        }
+
+        Ok(())
+    }
+
+    #[cfg(feature = "csv")]
+    /// Import data from a CSV laid out like the one produced by
+    /// [`Artifacts::export`]
+    pub async fn import<Read>(&mut self, read: &mut Read) -> Result<(), DialogArtifactsError>
```
Gozala, line 147: Nit: I wonder if it would make sense to have more generic "Archive" thing that could be imported, which in turn could be implemented for CSV and potentially other formats

rust/dialog-artifacts/src/artifacts.rs
cdata: Yeah, that's a good way to think about it.

cdata: https://github.com/dialog-db/staging/issues/56

rust/dialog-artifacts/src/artifacts/value.rs
```diff
@@ -315,7 +395,13 @@ impl From<&u8> for ValueDataType {
...
             7 => ValueDataType::Float,
             8 => ValueDataType::Record,
-            _ => ValueDataType::Null,
+            9 => ValueDataType::Symbol,
+            _ => {
+                println!(
+                    "WARNING! Encountered unsupported value tag '{value}'; defaulting to null (but this is undefined behavior)..."
+                );
+                ValueDataType::Null
+            }
```
Gozala, line 404: Nit: Seems to me that unknown values should be treated as bytes.

rust/dialog-artifacts/src/artifacts.rs
cdata: I think the door is open for that, for now let's just stick with CSV.

Strictly speaking, our serialization (`serde`) is universal and is actually what the CSV exporter is built upon.

cdata: https://github.com/dialog-db/staging/issues/57

PR #58, opened 2025-05-14
feat: Persisted artifacts revision

cdata:
This change implements automatic persistence of the latest revision of `Artifacts`. Previously, we asked the user to keep the books. They had to persist the revision of their database, and then provide the desired revision when opening it again. Now, the revision is automatically tracked against an "identifier" - a unique name - for the `Artifacts`. In the tests this is just a random string, but you could imagine it might be a DID.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
-    "rust-analyzer.cargo.target": null,
+    "rust-analyzer.cargo.target": "wasm32-unknown-unknown",
     "rust-analyzer.procMacro.ignored": {
         "async-trait": [
             "async_trait"


rust/dialog-artifacts/src/artifacts/store.rs
```diff
@@ -36,7 +38,7 @@ pub trait ArtifactStoreMut: ArtifactStore {
     async fn commit<Instructions>(
         &mut self,
         instructions: Instructions,
-    ) -> Result<(), DialogArtifactsError>
+    ) -> Result<Revision, DialogArtifactsError>
```
Gozala, line 41: Nice I wanted to ask for this other day actually.

.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-    "rust-analyzer.cargo.target": "wasm32-unknown-unknown",
+    "rust-analyzer.cargo.target": null,
```
Gozala:
```suggestion
    "rust-analyzer.cargo.target": "wasm32-unknown-unknown",
```

I assume this was unintentional change

PR #61, opened 2025-05-15
fix: Revision should return hash

cdata:
Also: make reset public, and make JS-facing commit "immutable" to bypass wasm-bindgen warnings.

Also also: fix a cache invalidation bug where we were not invalidating the name->root pointers in raw storage.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,6 +56,8 @@ jobs:
 
           nix develop --command bash -c "cat << 'EOF' | bash
 
+          set -eufo pipefail
+
           pushd typescript/dialog-artifacts-web-tests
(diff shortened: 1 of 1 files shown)


rust/dialog-storage/src/storage/cache.rs
```diff
@@ -60,6 +60,7 @@ where
     type Error = Backend::Error;
 
     async fn set(&mut self, key: Self::Key, value: Self::Value) -> Result<(), Self::Error> {
+        self.cache.lock().await.insert(key.clone(), value.clone());
```
Gozala, line 63: Should we instead be removing records from the cache ? Because this may end up evicting something from cache while not necessarily be something we intend to read at all.

Issue #60, opened 2025-05-14
API to fork & branch dbs


Gozala:
It would be nice to be able to fork and branch DBs instances. Per our discussion two would have slightly different semantics.

1. `artifacts.branch(name?: string)` would create a new instance that share same block store with the origin and has different identifier derived from the given name. Assuming we use `did:key`s I would imagine we'd use private key of the `artifacts` and `name` seed phrase to derive another `did:key` and store it all in the the same IDB database.

2. `artifacts.fork(as: CryptoKeyPair)` would create a new instance with a separate IDB instance copying all the relevant blocks.

I also suggest that default branch to be no different other than it using some canonical name like "main" or "default" at creation time, this will ensure single key-pair for the database where all the branch key-pairs are derived.

PR #63, opened 2025-05-16
feat: implement branching support

Gozala:
This partially implements idea described in the #60, but more important to me this addresses the problem that I was running into which is, if revision record is updated in the IDB `.reset` will not pick it up because it will hit a cache, what's worse it is not very obvious when it happens because if you did not have commits locally it will work as expected because there will be nothing in cache.

This changes address this problem by introducing another `branches` store to track the root. By default branch named "main" is used, but from rust side you could create different branches. In browser context this creates two stores in IDB one for blocks and other for branches.

Public API for branching and switching branches was intentionally left out, because it was not necessary to address problems I was running into. We can discuss how to proceed with public API under #60

### DIFF
--- a/rust/dialog-artifacts/src/artifacts.rs
+++ b/rust/dialog-artifacts/src/artifacts.rs
@@ -563,10 +563,10 @@ mod tests {
     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
     #[cfg_attr(not(target_arch = "wasm32"), tokio::test)]
     async fn it_commits_and_selects_facts() -> Result<()> {
-        let (storage_backend, _temp_directory) = make_target_storage().await?;
+        let (blocks_storage, branches_storage, _temp_directory) = make_target_storage().await?;
         let entity_order = |l: &Artifact, r: &Artifact| l.of.cmp(&r.of);
         let mut facts =
-            Artifacts::anonymous(storage_backend.clone(), storage_backend.clone()).await?;
+            Artifacts::anonymous(blocks_storage.clone(), branches_storage.clone()).await?;
 
         let mut data = vec![
             Artifact {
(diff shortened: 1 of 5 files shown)


rust/dialog-artifacts/src/artifacts.rs
```diff
@@ -86,7 +89,8 @@ where
         + 'static,
 {
     identifier: String,
-    storage: Storage<HASH_SIZE, CborEncoder, Backend>,
+    branch: String,
+    branches: Storage<HASH_SIZE, CborEncoder, Backend>,
```
cdata, line 93: We went down this route in Noosphere, and I ended up on the pattern where there is a factory of namespaced storage instances. In such a scenario, you would give `Artifacts` the factory and then it would allocate the namespaced storages that it cared about.

rust/dialog-artifacts/src/artifacts.rs
```diff
@@ -105,19 +109,29 @@ where
...
+
+        let branches = Storage {
             encoder: CborEncoder,
-            backend: backend.clone(),
+            backend: branches.clone(),
         };
 
         // TODO: We probably want to enforce some namespacing within storage so
         // that generic K/V storage can go e.g., in a different IDB store or a
         // different folder on the FS
```
cdata, line 130: Word to the wise

rust/dialog-artifacts/src/artifacts.rs
cdata: The subtext of my comment is that it won't stop at "one storage for blocks, one for branches."

cdata: https://github.com/subconsciousnetwork/noosphere/blob/94b0c5e8a91437e1486e02e29bc1aba1b7b225a8/rust/noosphere-storage/src/storage.rs#L8-L27

Gozala: Are you suggesting we do the same here ? Unfortunately in this context it is more complicated thanks to IDB, specifically you can't create new store on demand it has to happen when opening a DB through schema migration route. It is made even more complicate by the fact that not only you need to open DB you also need to close all open DBs with prior version otherwise open request will remain pending.

In the light of these complications I'm not sure how to go about implementing API linked above, best I can think of is to prefix keys with the name, but that does not seem that great. Could we perhaps wait until we have for more and then deal with that ?

That said, block store and branch store does seem to mirror noosphere's block store, kv store.

PR #64, opened 2025-05-16
feat: CI integration for radicle patches

Gozala:
This sets up github CI integration for radicle patches (equivalent of pull requests) as [per instructions here]( https://app.radicle.xyz/nodes/seed.radicle.garden/rad:zyMFFgrGTswVFj1X9uJrVwEV8EWu/tree/docs/project_setup.md).

⚠️ This is not a complete solution, we need hosted node that will do the coordination and report updates into patches in radicle, which radicle team offered to do for us, however repo needs to be configured along these ways for the integration to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,28 +12,24 @@ What if...
 
 ...you could host the whole database in a plain-old browser tab?
 
-## Contributing
+## Development
 
-Prerequisites:
+### Setup with Nix
 
-- You must have the [Radicle CLI](https://radicle.xyz/download) installed (`rad` command)
+This project uses Nix to provide a consistent development environment with all required dependencies.
 
-After cloning this repository, run the setup script to configure the Git remotes and aliases
+1. Install [Nix](https://nixos.org/download.html) if you don't have it already
+2. Enter the development environment:
 
 ```bash
-./scripts/setup.sh
+nix develop
 ```
 
-This script will:
+This automatically:
+- Sets up all development dependencies (Rust, WebAssembly tools, etc.)
+- Configures Git remotes and aliases for Radicle integration 
+- Prepares your environment for building and testing
 
-1. ✅ Check if the `rad` command is available
-2. 👾 Set up Git aliases to work with radicle patches (eqivalent of pull requests)
-3. ⛓️ Configure Git remotes for CI jobs.
 
-If you want to be core contributor with push access, reach out with a request to get your radicle node to be added to the delegates. You will need to provide your node id by running following command
-
-```sh
-rad self --nid
-```
-
-[datalog]: https://en.wikipedia.org/wiki/Datalog
+[datalog]:https://en.wikipedia.org/wiki/Datalog
+[radicle]:https://radicle.xyz/
--- a/flake.nix
+++ b/flake.nix
@@ -84,6 +84,8 @@
           cargo-nextest
           playwright-test
           nodejs
+          git
+          radicle-node  # Provides the 'rad' CLI tool
         ];
 
         interactive-dev-tools =
@@ -249,6 +251,26 @@
                 export PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS=1
                 export PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1;
                 export PLAYWRIGHT_BROWSERS_PATH=${playwright-driver.browsers}
+                
+                # Get repository root directory using git
+                REPO_ROOT="$(git rev-parse --show-toplevel)"
+                
+                # Check if radicle is available
+                if command -v rad >/dev/null 2>&1; then
+                  # Run the setup script if it exists
+                  if [ -f "$REPO_ROOT/scripts/setup-radicle.sh" ]; then
+                    echo "🛠️  Running Radicle setup script..."
+                    . "$REPO_ROOT/scripts/setup-radicle.sh"
+                  else
+                    echo "⚠️  Radicle setup script not found at $REPO_ROOT/scripts/setup-radicle.sh"
+                  fi
+                else
+                  echo "⛓️‍💥 Radicle toolchain ('rad' command) not found in PATH."
+                  echo "  💡 This development environment should include radicle-node,"
+                  echo "      but the 'rad' command is not accessible."
+                  echo "  💡 Ensure you have run 'nix develop' rather than using a legacy shell."
+                  echo "  💡 If the problem persists, try running 'which radicle-node' to debug."
+                fi
               '';
             };
         };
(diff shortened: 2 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -2,7 +2,12 @@ on:
   workflow_dispatch:
   pull_request:
   push:
-    branches: [main]
+    branches:
+      - main
+      # Run CI for brannches with this prefix so that radicle integration
+      # can pick those CI jobs and report them in patches (equivalent of pull
+      # requests)
+      - patches/*
```
Gozala, line 10: @cdata this is a controversial piece, so let me try to elaborate. The way integration works is in git we can configure remote with two upstreams one for github and one for radicle, something along the following lines

```sh
git remote add upstream rad://z941zp6YisjQUk4Q43qpaHJnLWzA
git remote set-url --push upstream rad://z941zp6YisjQUk4Q43qpaHJnLWzA/z6Mkej6T5F7YpZ5FgZ3sNf2ta9yeaUCxPQSTyFSiNuFMktBf
git remote set-url --add --push upstream https://github.com/dialog-db/staging.git
```

Then when you create patch (radicle pull request) push to upstream will create branch like `patches/....` which will push it both to radicle and github. Because branch name is prefixed with `patches/` it will kickoff the CI workflow and github integration will report updates to corresponding patch in the radicle.

Here I enable CI jobs for `patches/` prefixes to trigger CI jobs when that happens

cdata: The part I'm confused about is: why do you think this is controversial?

Gozala: I'm glad it's not, I thought it might be.

scripts/setup.sh
```diff
@@ -0,0 +1,163 @@
+#!/bin/sh
+# Don't use set -e to avoid terminal closing when sourced with . ./scripts/setup.sh
+# Instead, check return codes explicitly and report errors
+
+# Check if rad command is available
+if ! command -v rad >/dev/null 2>&1; then
+
+  echo "⛓️‍💥 This script requires the Radicle toolchain to be installed."
+  echo "  💡 Please install it from https://radicle.xyz/download and try again."
```
cdata: Speaking for myself, I will not be able to install Radicle this way (it will be a declared dependency in `flake.nix`, which is how it ends up on my system).

You've mentioned that a skill-gated barrier to entry is acceptable (and perhaps even desirable) for certain kinds of participation, so: why not assume that `flake.nix` is comprehensive and require that `nix develop` be the first step for bootstrapping a local development environment?

README.md
```diff
@@ -10,4 +10,30 @@ What if...
...
 
-...you could host the whole database in a plain-old browser tab?
\ No newline at end of file
+...you could host the whole database in a plain-old browser tab?
+
+## Contributing
+
+Prerequisites:
+
+- You must have the [Radicle CLI](https://radicle.xyz/download) installed (`rad` command)
```
cdata: More or less the same comment as https://github.com/dialog-db/dialog-db/pull/64/files#r2096558870

README.md
```diff
@@ -10,4 +10,30 @@ What if...
...
+
+After cloning this repository, run the setup script to configure the Git remotes and aliases
+
+```bash
+./scripts/setup.sh
```
cdata: Nit: this should probably be called `setup-radicle.sh`

README.md
```diff
@@ -10,4 +10,30 @@ What if...
 
 ...it had better transfer and caching properties than actual websites?
 
-...you could host the whole database in a plain-old browser tab?
\ No newline at end of file
+...you could host the whole database in a plain-old browser tab?
+
+## Contributing
```
cdata: I'll echo my sentiment that I don't think Radicle should be the low bar to contributing. It solves a problem, specifically that GitHub is a single point of failure and authority. But, where contributing is concerned it seems inferior on almost every metric compared to GitHub (and also most self-hosted options).

cdata: I'll paste my bottom line comment from our other conversation:

> I'll meet you half way on Radicle if you are willing to meet me half way on low-barriers to entry / participation / contribution.

scripts/setup.sh
```diff
@@ -0,0 +1,163 @@
...
+    echo "  ⚠️  No Radicle push URL is configured for upstream."
+    echo "      Expected: $RAD_PUSH_URL"
+    echo "  🚫 Aborting to avoid overwriting your customized remote configuration."
+    exit 1
+  fi
+
+  # upstream remote is properly configured
+else
+  echo "  ⛓️ Adding 'upstream' remote"
+  git remote add upstream "$RAD_FETCH_URL"
```
cdata: Is `upstream` required, or can this be customized?

README.md
Gozala: Should I remove this section from readme all together or should I change title so it does not feel like it's only way to contribute ?

scripts/setup.sh
Gozala: Would this basically imply adding job `develop` along the side of ones like `dialog-artifacts-web` that basically does more or less what this script does ? 

There is also whole https://app.radicle.at/nodes/seed.radicle.at/rad:zTGn5tnYJgF8qd6KDuodURvUvwLe but I'm not sure I understand how that is supposed to be used. 

Alternatively I can remove this script all-together and assume that git configuration is either done by the user or they don't care to use git aliases and remotes setup and do their own thing.

scripts/setup.sh
Gozala: To make CI workflow integration we basically will be pushing to both rad remote and github remote simultaneously, which is what "upstream" remote is configured here to do.

Added git alias assumes "upstream" remote to to automate all the steps via `git patch "description"` alias. Having joint remote also helps when updating patches (pull requests) because you can just "git push upstream".


So short answer is no `upstream` is not required and steps on line 20 can be run manually and pushes to both remotes could also be run manually. It is also possible to name multi endpoint remote something else, and I'm open to suggestions. Trying to figure out what remote names were used seems more complicated and I'm not sure it would be worth it.

scripts/setup.sh
cdata: If you have Nix installed, you can run `nix develop` and it will drop you into a shell environment with all the build tools this project needs set up and ready to go (including Node.js, Rust toolchain and on Linux even the web browser and webdriver).

I like the script, since it seems like setting things up to work with Radicle is pretty involved :shrug:

README.md
cdata: I don't mind it in the README, but it would be nice if basic contributions took a more conventional approach.

Gozala: I removed section about contributions and instead added instructions about development setup with nix

README.md
Gozala: Renamed to setup-radicle.sh

scripts/setup.sh
Gozala: I think I've made changes to nix.flake that will do what we want, but I'm not exactly sure this is what you meant.

PR #65, opened 2025-05-16
fix: reset before commit

Gozala:
This fixes following error that I was encountering if reset was called before transaction

```sh
Tree operation failed: Block not found in storage: #11111111111111111111111111111111
```

### DIFF
--- a/rust/dialog-artifacts/src/artifacts.rs
+++ b/rust/dialog-artifacts/src/artifacts.rs
@@ -117,7 +117,7 @@ where
             let revision_block = storage.get(&make_reference(identifier.as_bytes())).await?;
             let revision = if let Some(revision_hash_bytes) = revision_block {
                 // Check if the revision is NULL_REVISION_HASH
-                if revision_hash_bytes == NULL_REVISION_HASH.to_vec() {
+                if revision_hash_bytes == NULL_REVISION_HASH {
                     None
                 } else {
                     // For actual revisions, read the revision from storage
@@ -262,7 +262,6 @@ where
         revision_hash: Option<Blake3Hash>,
     ) -> Result<(), DialogArtifactsError> {
         // Determine target revision we are resetting to
-        // Step 1: Figure out needed revision_hash
         let required_hash = match revision_hash {
             // If a specific revision hash is provided, use it
             Some(hash) => hash,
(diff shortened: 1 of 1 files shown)


rust/dialog-artifacts/src/artifacts.rs
```diff
@@ -264,44 +264,78 @@ where
             self.value_index.write()
         );
 
-        let revision = if let Some(revision) = revision {
-            self.storage
+        if let Some(revision) = revision {
```
cdata: It seems that by changing this conditional, you now have duplicate spots where you set the hash (L288 and L321). I suspect you could get away with only doing L321.

rust/dialog-artifacts/src/artifacts.rs
```diff
@@ -1026,4 +1060,15 @@ mod tests {
 
         Ok(())
     }
+
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+    #[cfg_attr(not(target_arch = "wasm32"), tokio::test)]
+    async fn it_can_reset_before_commit() -> Result<()> {
```
cdata, line 1079: It would be nice to see a test for the other thing that was fixed as well: `it_updates_the_mutable_pointer_when_resetting`

rust/dialog-artifacts/src/artifacts.rs
```diff
@@ -1026,4 +1073,267 @@ mod tests {
...
+    #[cfg_attr(not(target_arch = "wasm32"), tokio::test)]
+    async fn it_stores_null_revision_hash_directly() -> Result<()> {
+        // Use memory storage backend to avoid file system errors
+        let storage_backend = MemoryStorageBackend::<[u8; 32], Vec<u8>>::default();
+
+        // Create an anonymous artifacts instance
+        let mut artifacts = Artifacts::anonymous(storage_backend).await?;
+
+        // Reset to NULL_REVISION_HASH explicitly
+        artifacts.reset(Some(NULL_REVISION_HASH)).await?;
```
cdata, line 1098: :thought_balloon: should we be populating the store with a pointer to `NULL_REVISION_HASH` when we first initialize `Artifacts`? (no action required, just occurred to me that this is what we seem to expect)

Gozala: I actually only ended up having to store `NULL_REVISION_HASH` because there was no way for me to delete stored hash if artifacts was reset to `NULL_REVISION_HASH`. I have considered adding delete to store, but hesitated to make larger changes.

That said we could probably store it, don't know if there's much value in it though

PR #68, opened 2025-05-19
chore: first draft of how privacy & encryption can be managed

Gozala:
[Preview of rendered markdown](https://github.com/dialog-db/dialog-db/blob/patches/a9fbcd8afcbeb3e40bc20deb345f7e82b5d4d30d/notes/privacy.md)


notes/privacy.md
```diff
@@ -0,0 +1,257 @@
...
+        L2["Level 2: Range Access<br/>Tree Navigation"]
+        L3["Level 3: Data Access<br/>Content Visibility"]
+    end
+    
+    L0 --> L1 --> L2 --> L3
+    
+    style L0 fill:#ddd,stroke:#333,stroke-width:1px
+    style L1 fill:#bfb,stroke:#333,stroke-width:1px
+    style L2 fill:#bbf,stroke:#333,stroke-width:1px
+    style L3 fill:#f9f,stroke:#333,stroke-width:1px
```
cdata: Nit: color choices here make the text difficult to read

image

notes/privacy.md
```diff
@@ -0,0 +1,257 @@
...
+### Level 0: No Access
+
+The base infrastructure level:
+
+- **Blob Store**: Stores fully encrypted, opaque blobs without any knowledge of their content or relationships
+- **Mutable Pointer**: Manages references without knowledge of what is being referenced
+- **No Data Visibility**: Cannot see or derive any value from user data
+- **Authorization Only**: Access only through proper authorization credentials
+
+At this level, infrastructure components are completely unaware of the data, only handling completely opaque, encrypted content. This ensures that infrastructure providers cannot derive any value from user data or metadata.
```
cdata, line 44: This isn't entirely true. Infrastructure knows the hashes of blobs, right? That's scant information for an attacker to work with, but it qualifies as metadata regardless.

notes/privacy.md
```diff
@@ -0,0 +1,257 @@
...
+Limited structural visibility:
+
+- **Blob Traversal**: Authorized actors can follow links between blobs
+- **Sync Assistance**: Can help with synchronization by providing missing connected nodes
+- **No Content Insight**: Still cannot see data content or key ranges
+- **Minimal Metadata**: Only visibility is blob connectivity
+
+This level enables efficient synchronization help - an L1 actor can bundle all connected nodes a user is missing into a single compressed package, improving transfer efficiency without revealing data content.
+
+### Level 2: Range Access
```
cdata, line 57: How do you enable L1 without giving L2 also?

cdata: Nevermind, I think I got to it eventually.

notes/privacy.md
ar9onaut: I assume that it's a hash of the encrypted blob and not of the blob before encryption. In which case it would not be a leak because you could derive it from the blob itself

PR #73, opened 2025-05-28
chore: DBSP hypothesis

Gozala:


notes/dbsp.md
```diff
@@ -0,0 +1,267 @@
...
+- Exploiting **prolly tree structure** and **EAV/AEV/VAE indexes** to determine which subtrees need replication
+- Using **query structure knowledge** (attributes, entities) to identify specific subtree ranges in the indexes
+- After evaluating conjuncts, using **variable bindings** to further constrain subsequent subtree replication
+- Leveraging **partial replication and caching hierarchy** to minimize data transfer and storage overhead
+- Achieving the **same selective loading** benefits through selective subtree replication instead of selective fact loading
+
+The key question is whether this approach can provide a **unified evaluation strategy** that works for both initial queries and incremental maintenance while:
+- Preserving current performance characteristics through selective subtree replication
+- Leveraging prolly tree properties for efficient change detection and partial replication
+- Minimizing blob store access through intelligent caching and replication strategies
```
Gozala, line 267: I got some interesting insights from @expede yesterday, trying to capture here before it escapes

Unless I've misunderstood something, we don't actually need a fancy approach of deriving insight after zsets propagate from first node to a next. That is because join can only filter out some facts, it can not add more facts.

notes/dbsp.md
```diff
@@ -0,0 +1,267 @@
...
+}
+```
+
+The exploration should validate whether we can:
+
+1. **For Incremental Maintenance**: When the store root pointer changes (indicating new facts), use our query planner's analysis to:
+   - Determine which prolly tree subtrees could contain facts affecting existing materialized views
+   - Leverage EAV/AEV/VAE index structure to identify relevant subtree ranges
+   - Replicate only those subtrees through our partial replication system
+   - Process the replicated fact changes through DBSP operators
```
cdata, line 245: We don't currently encode any metadata beyond the state of each fact. So IIUC the proposed strategy correctly... 

If you want a diff that tells you what was asserted and what was retracted across two revisions, you would actually need to do a fine-grained diff of each changed segment (it isn't sufficient just to use the coarse-grained tree diff). This needs to be observed because fine-grained is significantly more expensive than coarse-grained in this context.

Gozala: I imagine that we would diff locally after replication before we feed changes into DBSP operators. In fact I imagine that selections returned by `select` would maintain some local state so when you call `.pull()` on them they will be able to compute delta and return it as ZSet.

PR #76, opened 2025-05-30
Divergence clock idea

Gozala:
[Rendered version](https://github.com/dialog-db/dialog-db/blob/patches/2c40176539076238940d60ec0ba41d8b82dce40d/notes/divergence-clock.md)


notes/images/divergence-clock.md
```diff
@@ -0,0 +1,196 @@
...
+I think we could synthesize ideas from various logical clocks that would offer the best tradeoffs given our propblem domain. Here are desired properties for our clock:
+
+1. Ability to identify concurrent changes
+2. Ability to compare any two events (without having to read arbitrary tree branches)
+
+We could accomplish first goal we must capture synchronization points e.g. tree root a.k.a revision from where changes where made. This way if two changes have same a same revision they could be considered concurrent. However, since revision is a cryptogrphic hash we would not be able to compare two changes with different revisions as it would not be clear which one is older.
+
+We can build upon this general idea but instead of using revision hash we could use monotonically growing tree size as synchronization point instead. Specifically we could use `{ since, drift, at }` tuple to represent our logical clock where:
+
+- `since` - Is count of commits in the database
```
Gozala: Alternatively we could just increment this on every merge since that would still increase monotonically and it would be more space efficient. I don’t think number of commits is that useful for us anyway

notes/images/divergence-clock.md
```diff
@@ -0,0 +1,196 @@
...
+- `drift` - Is the number of commits made since updating `time`
+- `at` - Is the unique identifier of the site that produced change
```
Gozala: I really would like us not to need this perhaps we could instead have section in the tree for capturing sites so this could be a varint as opposed to something really longs as UUID

notes/images/divergence-clock.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Idea
+
+I think there are two potential ways we could address such limitation that I'll try to describe below
+
+### Partially extending causal reference space
+
+We could lift imposed limitation on causal references e.g instead of requiring that all causal references share same `{ the, of }` we could require that all causal references share attribute namespace (slice of `the` before `/`).
+
+This approach would extend consistency guartees from fact to schema boundary, while still keeping relevant causal information in cousin branches so close by in the tree.
```
cdata: My intuition is that this is too generalized. The whole reason e.g., PouchDB gives you manual conflict resolution is that you need domain knowledge to say when one version of a structure is in conflict with another.

Gozala: Are you referring to the whole document or a [Partially extending causal reference space](https://github.com/dialog-db/dialog-db/blob/93b98172632f70b01109593d98a5d28229012842/notes/divergence-clock.md#partially-extending-causal-reference-space) segment itself ?


If later I think I agree although I would describe it as not general enough 🤔. It is more general than what we have (or what you get with CRDTs) yet not general enough like transactional guarantees you get in relational databases so it'll like be confusing if nothing else. I did put it down mostly because we have discussed it in person prior as a possibility.

If you suggest the whole idea is too general I'd like to discuss more so I understand your position better here. Specifically my thinking is that:

1. Concurrent changes (same `since` different `at`) that overlap (same `the`, `of`) can be considered conflicts. 
2. Conflicting changes form alternate realities that user can choose from while system selects a default in deterministic manner.

I am still not sure how user overriding a default choice could manifest with minimal overhead, but I think we can pun on that for now unless we have some interesting idea.

notes/divergence-clock.md
```diff
@@ -0,0 +1,196 @@
...
+| Commit | 4 | C | 1 |
+| Push | 7 | C |  |
+
+## Alternative Approach
+
+Instead of trying to captureing complete commit history inside a tree we could take completely different approach and instead have upstream tree and local workspace tree for local changes between. Reconciliation can take place on-push by merging local changes into upstream tree. This way tree would represent reconciled merge state so at query time no reconciliation would be required.
+
+This approach has a tradeoff making it impossible to surface conflicts at query time as they would need to be resolved before changes are pushed and once pushed information about prior conflict will be gone.
+
+Perhaps there are more lessons here to learn from datomic, as far as I understand in datomic tree represents current state, yet history is persisted in it's own index from which everything could be recovered. In other words datomic is can be though of index over the transaction log which is persisted to optimized for querying current state.
```
ar9onaut, line 188: I believe datomic simply has `/t/e/a → [value, retract:boolean]` index along with other three and when things are retracted entries are removed from the other indexes but retained in this one. This way historical data is retained and can still be queried in a less efficient way while primary indexes can be used to query latest state.

💡 Perhaps if user overrides conflict we could prune primary indexes from the concurrent changes that were not selected which will not incur overhead but rather reduce it. Although eventual consistency does imply that conflicts can still get resurfaced later as other replicas sync, furthermore two different users may choose different winners that will need to be reconciled also 😵‍💫

notes/images/divergence-clock.md
cdata: No, not the whole idea, just that section of the document

PR #91, opened 2025-06-27
feat: Introduce `dialog-diagnose` crate and CLI tool

cdata:
This change introduces a new inspection tool for Dialog databases: a CLI / TUI that supports navigating facts and the internal structure of the database's indexes.

This tool is incomplete in some areas, and also non-comprehensive. But, it has reached a threshold of minimum viable usefulness.

Some things it can do:

- Load a CSV into a one-off Dialog database
- Lazily present all facts in the database
- Visualize the distribution of segment entries in the entity index
- Navigate the entity index prolly tree and inspect its nodes

I started with "just" the entity index, because we have plans to explore using a single prolly tree to embody all indexes. If we go that route, only one tree will need to be inspected in the UI. If we don't go that route, we may need to add some way to cycle through indexes.

Here is an example of it running:

[Screencast From 2025-06-26 15-31-40.webm](https://github.com/user-attachments/assets/96b15fec-db67-464b-815b-2d3521c2b096)


rust/dialog-diagnose/src/widgets/facts/table.rs
```diff
@@ -0,0 +1,113 @@
...
+    where
+        Self: Sized,
+    {
+        let header_style = Style::default();
+        let selected_row_style = Style::default();
+
+        let selected_col_style = Style::default();
+        let selected_cell_style = Style::default().add_modifier(Modifier::REVERSED);
+
+        let header = ["Entity", "Attribute", "Value", "Cause"]
```
ar9onaut, line 32: I would go with `The`, `Of`, `Is`, `Cause` to stick to the same naming convention

ar9onaut (approved):
I just skimmed through it mostly to learn. My general suggestions would be for us to stick to consistent naming convention. Specifically I think we should either embrace `{ the, of, is, cause }` or drop it. I also think we should adopt index nodes and segment nodes as terms as branches is not very clear.

ar9onaut:
This is really awesome bravo both @bfollington and @cdata

PR #92, opened 2025-06-28
feat: Universal index

cdata:
This change switches us from using three distinct indexes to using a single, universal index with tagged keys. The observation was made that by tagging the keys, a single prolly tree may naturally balance segments by key type. One trade-off is that some segments that embody index boundaries will likely cut across two (or more, for small data) indexes.

Additionally, `diagnose` has been refined with:

- Tree level indicator
- Scrolling for tree view
- Highlighting for key bytes

[Screencast From 2025-06-27 23-02-54.webm](https://github.com/user-attachments/assets/829f22bb-f000-45d0-b34e-0680d6c01c97)

### DIFF
--- a/rust/dialog-artifacts/src/key.rs
+++ b/rust/dialog-artifacts/src/key.rs
@@ -1,5 +1,12 @@
+//! Key structures for indexing artifacts in prolly trees.
+//!
+//! This module provides the key layout and manipulation utilities for creating
+//! efficient indexes over semantic triples. Keys are structured to enable fast
+//! range queries over different access patterns (by entity, attribute, or value).
+
 use std::ops::{Deref, DerefMut};
 
+use dialog_prolly_tree::KeyType;
 use serde::{Deserialize, Serialize};
 use serde_big_array::BigArray;
 
@@ -15,6 +22,10 @@ pub use value::*;
 mod part;
 pub use part::*;
 
+/// Helper macro for creating mutable slices from byte arrays at compile time.
+///
+/// This macro is used internally for efficient manipulation of key byte layouts
+/// without runtime bounds checking.
 macro_rules! mutable_slice {
     ( $array:expr, $index:expr, $run:expr ) => {{
         const START: usize = $index;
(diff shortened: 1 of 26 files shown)


rust/dialog-artifacts/src/key.rs
```diff
@@ -20,19 +36,25 @@ macro_rules! mutable_slice {
...
+/// Length of the key tag field in bytes
+pub(crate) const TAG_LENGTH: usize = 1;
+/// Length of the entity field in key bytes
 pub(crate) const ENTITY_LENGTH: usize = 64;
+/// Length of the attribute field in key bytes
 pub(crate) const ATTRIBUTE_LENGTH: usize = 64;
+/// Length of the value data type field in key bytes
 pub(crate) const VALUE_DATA_TYPE_LENGTH: usize = 1;
+/// Length of the value reference field in key bytes
 pub(crate) const VALUE_REFERENCE_LENGTH: usize = 32;
```
ar9onaut, line 50: 🤔 So if we have `[entity, attribute, entity]` we would not be able to capture it in just key then because we entity is twice as long 😕

I think we need to capture tradeoffs somewhere regarding decisions about entity representation size limit and implications so we can have a better decision making matrix.

ar9onaut: What I have previously imagined which I think is slightly different from what's here is that each data type we support has tag including `entity: 0, attribute: 1, reference: 3, bytes: 4, ....` this way key would be just triple and data type in the head would tell us what index it's is

1. `[entity, attribute, value]` would start with `0`
2. `[attribute, entity, value]` would start with `1`.
3. `[value, attribute, entity]` would start with anything other than 0 or 1.

That said, I'm not sure shaving one byte here provides significant value, although I did like the fact that for `[entity, attribute, entity]` relations 1st and 3rd indexes converged into one.

ar9onaut: On the flip side if we add some other index e.g. `[entity, value, attribute]` it may no longer work out so current design is more flexible.

ar9onaut (approved):
Looks good. Simplification feel really nice.

rust/dialog-artifacts/src/key.rs
cdata: The advantages of this approach are clear enough:

- All keys have a fixed size (which gives us optimization opportunities)
- Any key type can be converted to any other key type (spares us from book keeping and cross-referencing values)

> That said, I'm not sure shaving one byte here provides significant value, although I did like the fact that for [entity, attribute, entity] relations 1st and 3rd indexes converged into one.

This is interesting but it's just a curiosity until we demonstrate to ourselves that it's useful. We have an analysis tool now, so perhaps you can experiment for yourself to decide if this would be useful to do. In the mean time, every new element we introduce to the keyspace undermines the "palindromic" property you're describing here.

PR #94, opened 2025-06-30
fix: outdated and invalid statements

ar9onaut:
- Fix invalid description of vector clocks
- Describe `since` as counter as opposed to commit count.
- Add note about convergence preference

### DIFF
--- a/notes/divergence-clock.md
+++ b/notes/divergence-clock.md
@@ -2,22 +2,22 @@
 
 ## Goal
 
-Supporting concurrent writes into database implies that multiple actors can create concurrent commits _(containing assertions / retractions)_ than need to be reconciled by all partial replicas. Reconciliation requirements following:
+Supporting concurrent writes into database implies that multiple actors can create concurrent commits _(containing assertions / retractions)_ that need to be reconciled by all partial replicas. Reconciliation requirements following:
 
 1. Identifying concurrent changes.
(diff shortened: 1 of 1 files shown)


notes/divergence-clock.md
```diff
@@ -108,14 +114,16 @@ I think we could synthesize ideas from various logical clocks that would offer t
 1. Ability to identify concurrent changes
 2. Ability to compare any two events (without having to read arbitrary tree branches)
 
-We could accomplish first goal we must capture synchronization points e.g. tree root a.k.a revision from where changes where made. This way if two changes have same a same revision they could be considered concurrent. However, since revision is a cryptogrphic hash we would not be able to compare two changes with different revisions as it would not be clear which one is older.
+We could accomplish first goal we must capture synchronization points e.g. tree revision from where changes where made. This way if two changes have same a same revision they could be considered concurrent. However, since revision is a cryptogrphic hash we would not be able to compare two changes with different revisions as it would not be clear which one is older.
```
Gozala:
```suggestion
To could accomplish first goal we must capture synchronization points e.g. tree revision from where changes where made. This way if two changes have same a same revision they could be considered concurrent. However, since revision is a cryptogrphic hash we would not be able to compare two changes with different revisions as it would not be clear which one is older.
```

PR #95, opened 2025-07-02
feat: Efficient encoding for columnar data

cdata:
This change cleans up the encoding experiment in [stub/column-compression](https://github.com/dialog-db/dialog-db/tree/stub/column-compression) and demonstrates how it may be used to enable zero-copy access to encoded buffers.

The intention is to eventually encode the nodes in our prolly tree with this strategy, so that general access to the tree is cache friendly, memory efficient and minimizes the time spent converting between representations. It may be useful for other data structures down the road, but its virtues as an encoding are highly dependent on how repetitive the "cells" in the input data are.

I won't belabor the details in the PR description - please check out the comments in the code!


ar9onaut (approved):
Looks great to me, thanks for all the detailed comments they are super helpful when trying to read through the code.

One question that occured to me while reading this was why not call it `Columnar` as and `columns` as opposed to `Cellular` and `cells`.

PR #98, opened 2025-07-07
feat: Describe Query Syntax Schema

Gozala:


notes/query.md
```diff
@@ -0,0 +1,813 @@
+# DialogDB Query Syntax Schema
+
+DialogDB is a database that uses **facts** and **rules** to deduce conclusions. It can be queried using Abstract Syntax Trees (ASTs) in the [IPLD data model](https://ipld.io/docs/data-model/kinds/) and encoded as JSON using [DAG-JSON codec](https://ipld.io/docs/codecs/known/dag-json/).
+
+All knowledge in DialogDB is represented as facts in form of `{the, of, is, cause}` relations:
+- **the**: Attribute/predicate
+- **of**: Entity/subject
+- **is**: Value/object
+- **cause**: Causal relationship (timestamp)
```
just-be-dev, line 9: I don't have the best understanding of what the `cause` relationship here actually is. I presume it's _not_ an equivalent of `CREATED_AT`. Is this timestamp supposed to point to the timestamp of some other facts is somehow related to this fact? 

So is it something like the below example?

`{ the: Effect, of: Checkbox, is: Clicked, ts: 0001 }`
`{ the: CheckState, of: Checkbox: is: Checked, cause: 0001 }`

Gozala: It is moving target, but currently it is hash of the fact it succeeds. Kind of like in git every commit has a head which is the previous commit.

I do say it is moving target, because most likely we're going to replace it with either

https://github.com/dialog-db/dialog-db/blob/main/notes/divergence-clock.md

or more generalized version described here https://github.com/dialog-db/dialog-db/pull/93/files

PR #99, opened 2025-08-17
query engine in rust

Gozala:
Port of query engine that also improves on it's design

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -677,6 +677,7 @@ dependencies = [
  "thiserror 2.0.12",
  "tokio",
  "tokio-test",
+ "wasm-bindgen-futures",
 ]
 
 [[package]]
--- a/rust/dialog-query/Cargo.toml
+++ b/rust/dialog-query/Cargo.toml
@@ -27,11 +27,19 @@ dialog-common = { workspace = true }
 futures-util = { workspace = true }
 futures-core = { workspace = true }
 async-stream = { workspace = true }
-tokio = { workspace = true, features = ["rt", "rt-multi-thread", "macros"] }
 
 # Macro utilities
 paste = "1.0"
 dialog-query-macros = { workspace = true }
+
+# Tokio with platform-specific features
+# wasm32 only supports: sync, macros, io-util, rt, time
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+tokio = { workspace = true, features = ["rt", "rt-multi-thread", "macros"] }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+tokio = { workspace = true, features = ["rt", "macros", "sync"] }
+wasm-bindgen-futures = "0.4"
 
 
 [dev-dependencies]
--- a/rust/dialog-query/src/artifact.rs
+++ b/rust/dialog-query/src/artifact.rs
@@ -6,7 +6,7 @@ pub use dialog_artifacts::{
 };
 pub use futures_util::stream::Stream;
 
-pub use dialog_common::ConditionalSend;
+pub use dialog_common::{ConditionalSend, ConditionalSync};
 // For testing, we can access MemoryStorageBackend through dialog-artifacts
 // to ensure version consistency
 pub trait Store: ArtifactStore + Clone + Send + 'static {}
--- a/rust/dialog-query/src/query.rs
+++ b/rust/dialog-query/src/query.rs
@@ -1,7 +1,7 @@
 //! Query trait for polymorphic querying across different store types
 
 use async_stream::try_stream;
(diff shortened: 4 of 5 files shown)


rust/dialog-query/notes/schema.md
```diff
@@ -0,0 +1,133 @@
...
+    #[many]
+    Address(String),
+}
+```
+
+This expands to following form
+
+```rs
+#[derive(Attribute::one("microshaft.employee"))]
+struct Name(String);
```
Gozala:
```rs
impl Attribute for Name {
  namespace() → &'static str {
    "microshaft.employee"
  }
}
```

rust/dialog-query/notes/schema.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+You should be able to query using schema by writing something like this:
+
+```rs
+async fn demo() -> Result<()> {
+    let store = MemoryStorageBackend::default();
+    let artifacts = Artifacts::anonymous(store).await?;
+
+    let employees = Employee.query(&artifacts)?.collect().await;
```
Gozala:
```suggestion
    let employees = Employee::match(Name("Alice"), _)
      .query(&artifacts)?.collect().await;
```

Gozala:
```rs
let employees = Employee::match(Name("Alice"), Job::_)
      .query(&artifacts)?.collect().await;
```

Gozala:
```rs
Employee::match(Name::match("Alice"), Job::any())
  .query(&artifacts)
```

Gozala:
```rs
Employee::find()
  .filter(Name("Alice"))
  .query(&artifacts);
```

rust/dialog-query/notes/schema.md
```diff
@@ -0,0 +1,133 @@
...
+Would be even better if we could annotate it inside the composition, but not sure how could this be manifested.
+
+```rs
+#[schema]
+struct Employee(
+  Name,
+  #[implicit("Earth")]
+  Address
+);
+```
```
Gozala:
```rs
enum MicroshaftEmployee {
    Name(String),
    Job(String),
    Salary(u32),
    #[many]
    Address(String),
}

enum Familiar {
  UserName(String)
}
```

Gozala:
```rs
#[schema]
struct Person(
  MicroshaftEmployee::Name | Familiar::UserName
);
```

Gozala:
```prolog
person(Name) :-
  MicroshaftEmployee(Name)

person(Name) :-
  Familiar(Name)
```

Gozala:
```rs
#[relation]
struct Name(String)
#[schema]
struct Person(Name)

impl Rule for Person {
  where() → {
  }
}

impl Rule for Person {
  where() → {
  }
}
```

Gozala: @cdata so here is one very contrived example in datalog where we define name as a thing that can be read from some entity or something that is join away. This is a way we can define person that can be resolved in two different ways, but not as trivial just resolving different attribute name

```prolog
person(Name, Address) :-
   the("person/name", Of, Name),
   the("person/address", Of, Address).

person(Name) :-
   the("user/account", User, Account),
   the("account/title", Account, Name),
   the("user/address", User, Address),
   not the("account/closed", Account, true).
```

Ideally one would define schemas in an abstract way like set of components it must have, but then have various ways it could be derived.

Gozala: I imagine something like this, but I would prefer things like `self.name()` and `self.address()` would instead be parameters passed into `where` however I'm not sure it could be done without excessive annotations. Ideally I can also bend compiler to my will and make it complain if some of the parameters Don't get bound.

```rs
#[rule]
struct Person(Name, Address)

impl Rule for Person {
  where(&self) → Where {
    [
       Name::of(self).is(self.name()),
       Address::of(self).is(self.address())
    ]
  }
}

impl Rule for Person {
  where(&self) → Where {
    let account = Variable::new("account")
    [
       User::Account::of(self).is(account),
       Account::Title::of(self).is(self.name()),
       User::Address::of(self).is(self.address()),
       Account::Closed::of(self).not(true)
    ]    
  }
}
```

Gozala: Also one thing that I'm arriving to is that through various conversation is that I think app authors should basically just write down

```rs
#[rule]
struct Person(Name, Address)
```

And let system fill in the `impl Rule` parts, or at least allow system to write the second one if not the first one

PR #112, opened 2025-12-04
Port of the rest storage backend from the sync branch

Gozala:
Cleans up s3 storage backend from the sync branch to target main branch.

### DIFF
--- a/rust/dialog-storage/src/storage/backend/s3/helpers.rs
+++ b/rust/dialog-storage/src/storage/backend/s3/helpers.rs
@@ -7,9 +7,9 @@ use bytes::Bytes;
 use hyper::server::conn::http1;
 use hyper_util::rt::TokioIo;
 use s3s::dto::{
-    DeleteObjectInput, DeleteObjectOutput, ETag, GetObjectInput, GetObjectOutput,
-    HeadObjectInput, HeadObjectOutput, ListObjectsV2Input, ListObjectsV2Output, Object,
-    PutObjectInput, PutObjectOutput, StreamingBlob, Timestamp,
+    DeleteObjectInput, DeleteObjectOutput, ETag, GetObjectInput, GetObjectOutput, HeadObjectInput,
+    HeadObjectOutput, ListObjectsV2Input, ListObjectsV2Output, Object, PutObjectInput,
+    PutObjectOutput, StreamingBlob, Timestamp,
 };
 use s3s::service::S3ServiceBuilder;
 use s3s::{S3, S3Request, S3Response, S3Result, s3_error};


rust/dialog-storage/Cargo.toml
```diff
@@ -7,6 +7,17 @@ license.workspace = true
 
 [features]
 helpers = ["tempfile", "anyhow", "rand"]
+http_tests = []
+r2_tests = []
+s3_integration_tests = []
+test-utils = [
```
cdata: Let's pick one consistent casing for features. I have a slight preference for kebab but I could go either way.

rust/dialog-storage/README.md
```diff
@@ -1,4 +1,230 @@
...
+use std::path::Path;
+
+let backend = FileSystemStorageBackend::<Vec<u8>, Vec<u8>>::new(
+    Path::new("/path/to/storage")
+).await?;
+```
+
+**Use cases**: Local persistence, desktop applications, server-side storage
+
+**Platform**: Not available on WASM
```
cdata: The layout of these sections is weird. It would be more helpful to have a table at the top with columns "use-case" and "Wasm" and the "Wasm" column can have checkmarks or no checkmarks.

Then, separately you could have an example usage section. Probably even better, it would be good to have an `examples` directory to contain usage examples (because those can be checked in CI easily). Or, barring that, moving these examples into code docs so that they run as doc tests.

rust/dialog-storage/README.md
```diff
@@ -1,4 +1,230 @@
...
+backend.delete(&key).await?;
+
+// List objects (returns S3 keys with prefix)
+let result = backend.list(None).await?;
+for key in result.keys {
+    println!("{}", key);
+}
+```
+
+## Testing
```
cdata: Ideally for testing we can run `nix check` and cover everything that isn't side-effect-y.

rust/dialog-storage/README.md
```diff
@@ -1,4 +1,230 @@
 # dialog-storage
 
-Generalized API for constructing content addressed storage from different
-backends and encoding schemes.
\ No newline at end of file
+Generalized API for constructing content addressed storage from different backends and encoding schemes.
```
cdata, line 3: In general regarding this README update: the code examples seem like they could easily fall out of date, and the accompanying prose is not especially helpful (when it is present at all). Consider moving substantially all of it into inline code documentation and expanding on the literal explanation of how to use the crate.

rust/dialog-storage/src/storage/backend.rs
```diff
@@ -14,6 +14,14 @@ mod indexeddb;
 #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
 pub use indexeddb::*;
 
+/// S3/R2-compatible storage backend.
+pub mod s3;
+pub use s3::{S3, S3StorageError};
```
cdata, line 21: Should we feature-gate this part of the crate? E.g., so that someone who doesn't care about remote storage doesn't incur the related dependencies.

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+pub use access::{Acl, Credentials, Invocation, Service, Session, SigningError};
+
+mod checksum;
+pub use checksum::{Checksum, Hasher};
+
+use crate::{DialogStorageError, StorageBackend, StorageSink, StorageSource};
+
+// ============================================================================
+// Request Types
+// ============================================================================
```
cdata: This isn't an idiomatic documentation format for Rust.  Module-level documentation should be:

```rust
//! The module docs
```

If you find you need dividing lines within a single file, it's an indicator to me that the file should be broken out into discrete modules.

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+}
+
+/// S3-safe key encoding that preserves path structure.
+///
+/// Keys are treated as `/`-delimited paths. Each path component is checked:
+/// - If it contains only safe characters (alphanumeric, `-`, `_`, `.`), it's kept as-is
+/// - Otherwise, it's base58-encoded and prefixed with `!`
+///
+/// The `!` character is used as a prefix marker because it's in AWS S3's
+/// "safe for use" list and unlikely to appear at the start of path components.
```
cdata: An external link to the related docs would be nice here.

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+// ============================================================================
+// Tests
+// ============================================================================
+
+#[cfg(all(test, not(target_arch = "wasm32")))]
+mod unit_tests {
+    use super::*;
+
+    #[test]
+    fn test_url_without_prefix() {
```
cdata: I have a personal preference of behavior-style `it_does_the_thing`-style test names. I'm happy to revisit this, but the existing body of unit tests follow this style and I would like to give privilege to consistency.

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+///
+/// The `!` character is used as a prefix marker because it's in AWS S3's
+/// "safe for use" list and unlikely to appear at the start of path components.
+///
+/// Examples:
+/// - `remote/main` → `remote/main` (all components safe)
+/// - `remote/user@example` → `remote/!<base58>` (@ is unsafe, encode component)
+/// - `foo/bar/baz` → `foo/bar/baz` (all safe)
+pub fn encode_s3_key(bytes: &[u8]) -> String {
+    use base58::ToBase58;
```
cdata: Nit: imports at the top of the module unless you have a mechanical reason to do otherwise.

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+
+    let key_str = String::from_utf8_lossy(bytes);
```
cdata: Are we avoiding a legitimate failure case here?

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+    let components: Vec<&str> = key_str.split('/').collect();
```
cdata: Why collect only to map it and collect into a vec again?

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+///
+/// Path components starting with `!` are base58-decoded.
+/// Other components are used as-is.
+pub fn decode_s3_key(encoded: &str) -> Result<Vec<u8>, S3StorageError> {
+    use base58::FromBase58;
+
+    let components: Vec<&str> = encoded.split('/').collect();
+    let mut decoded_components: Vec<Vec<u8>> = Vec::new();
+
+    for component in components {
```
cdata: You should be able to collect into a `Result` type to simplify this (and also note that we avoid the second `Vec` allocation).

```rs
encoded.split('/').map(|component| {
  if let Some(encoded_part) = component.strip_prefix('!') {
    encoded_part.from_base58().map_err(|e| {
      S3StorageError::SerializationFailed(format!(
          "Invalid base58 encoding in component '{}': {:?}",
          component, e
      ))
    })
  } else {
    Ok(component.as_bytes().to_vec())
  }
}).collect::<Result<Vec<u8>, S3StorageError>>()
```

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+                ))
+            })?;
+            decoded_components.push(decoded);
+        } else {
+            // Use as-is
+            decoded_components.push(component.as_bytes().to_vec());
+        }
+    }
+
+    // Join with /
```
cdata: Can this be accomplished with [`Vec::join`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.join)?

rust/dialog-storage/src/storage/backend/s3/access.rs
```diff
@@ -0,0 +1,705 @@
+//! AWS S3 Signature Version 4 signing implementation.
+//!
+//! This module provides presigned URL generation for S3-compatible storage services
```
cdata, line 3: A link to relevant external documentation would be nice here (I recall struggling to find it when you first mentioned this pattern).

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait Request: Invocation + Sized {
+    /// The request body, if any. Used when sending the request.
+    fn body(&self) -> Option<&[u8]> {
+        None
+    }
+
+    /// Perform this request against the given S3 backend.
+    async fn perform<Key, Value>(
```
cdata: What is the case for overriding this trait method?

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+
+        Ok(builder.send().await?)
+    }
+}
+
+/// S3/R2-compatible storage backend.
+///
+/// # Example
+///
+/// ```ignore
```
cdata: Prefer `no_run` where possible so that the test is still compiled.

rust/dialog-storage/src/storage/backend/s3/access.rs
```diff
@@ -0,0 +1,705 @@
...
+
+#[cfg(not(target_arch = "wasm32"))]
+use std::time::SystemTime;
+
+#[cfg(target_arch = "wasm32")]
+use web_time::{SystemTime, web::SystemTimeExt};
+
+// ============================================================================
+// Public Types
+// ============================================================================
```
cdata: See comment at the top of `s3.rs`

rust/dialog-storage/src/storage/backend/s3/access.rs
```diff
@@ -0,0 +1,705 @@
...
+}
+
+/// Cloud storage service configuration for request signing.
+///
+/// Contains the signing parameters for different cloud storage services.
+/// Does NOT include bucket, endpoint, or session duration - those are
+/// specified when creating a session.
+#[derive(Debug, Clone)]
+pub enum Service {
+    /// AWS S3 or S3-compatible service (including Cloudflare R2).
```
cdata, line 41: Do you have any other backends in mind that may warrant alternate variants?

rust/dialog-storage/src/storage/backend/s3/access.rs
```diff
@@ -0,0 +1,705 @@
...
+        let secret = format!("AWS4{}", credentials.secret_access_key);
+        // Derive signing key: HMAC chain of date -> region -> service -> "aws4_request"
+        let k_date = Self::hmac(secret.as_bytes(), date);
+        let k_region = Self::hmac(&k_date, service.region().as_bytes());
+        let k_service = Self::hmac(&k_region, service.name().as_bytes());
+        Self(Self::hmac(&k_service, b"aws4_request"))
+    }
+
+    fn hmac(key: &[u8], data: &[u8]) -> Vec<u8> {
+        let mut mac = Hmac::<Sha256>::new_from_slice(key).expect("HMAC accepts any key size");
```
cdata: In general I do not like to ship code that may panic at runtime. If it's infallible in practice, an inline comment justifying the panic-inducing call should be included.

rust/dialog-storage/src/storage/backend/s3/access.rs
```diff
@@ -0,0 +1,705 @@
...
+    /// # Arguments
+    /// * `credentials` - AWS credentials for signing
+    /// * `service` - The cloud storage service configuration
+    /// * `duration` - URL signature expiration in seconds (e.g., 86400 = 24 hours)
+    pub fn new(credentials: &Credentials, service: &Service, duration: u64) -> Self {
+        Self::new_at(credentials, service, duration, current_time())
+    }
+
+    /// Create a new authorized session with a specific timestamp (useful for testing).
+    pub fn new_at(
```
cdata: `new_at_time`

rust/dialog-storage/src/storage/backend/s3/access.rs
```diff
@@ -0,0 +1,705 @@
...
+
+        // Build signed headers
+        let mut headers = vec![("host".to_string(), host.clone())];
+        if let Some(checksum) = request.checksum() {
+            let header_name = format!("x-amz-checksum-{}", checksum.name());
+            headers.push((header_name, checksum.to_string()));
+        }
+        headers.sort_by(|a, b| a.0.cmp(&b.0));
+
+        let signed_headers_str: String = headers
```
cdata: It seems like this string could be built up in the same iteration loop that builds the `canonical_headers` via some kind of aggregation.

Also, nit: don't repeat obvious type information in variable names.

rust/dialog-storage/src/storage/backend/s3/access.rs
```diff
@@ -0,0 +1,705 @@
...
+/// Errors that can occur during signing.
+#[derive(Debug)]
+pub enum SigningError {
+    /// The endpoint URL is invalid (e.g., missing host).
+    InvalidEndpoint(String),
+    /// Failed to parse a URL.
+    UrlParse(url::ParseError),
+}
+
+impl std::fmt::Display for SigningError {
```
cdata: please derive `thiserror::Error` instead

rust/dialog-storage/src/storage/backend/s3/access.rs
```diff
@@ -0,0 +1,705 @@
...
+
+        Ok(Authorized { url, headers })
+    }
+}
+
+/// Trait for request invocations that can be authorized and sent.
+///
+/// Implementors describe the HTTP method, URL, and optional checksum/ACL
+/// for S3 requests. The [`Request`](super::Request) trait extends this
+/// with a body and the ability to perform the request.
```
cdata: It's not clear to me why these are separate traits (or why they should live in different modules given their proximate relevance).

rust/dialog-storage/src/storage/backend/s3/access.rs
```diff
@@ -0,0 +1,705 @@
...
+
+impl From<url::ParseError> for SigningError {
+    fn from(e: url::ParseError) -> Self {
+        Self::UrlParse(e)
+    }
+}
+
+// ============================================================================
+// Helper Functions
+// ============================================================================
```
cdata: See comment at the top of `s3.rs`

rust/dialog-storage/src/storage/backend/s3/access.rs
```diff
@@ -0,0 +1,705 @@
...
+    {
+        DateTime::<Utc>::from(SystemTime::now().to_std())
+    }
+    #[cfg(not(target_arch = "wasm32"))]
+    {
+        DateTime::<Utc>::from(SystemTime::now())
+    }
+}
+
+fn hex_encode(bytes: &[u8]) -> String {
```
cdata, line 431: Some human-readable documentation would be nice even though this is private API.

rust/dialog-storage/src/storage/backend/s3/access.rs
```diff
@@ -0,0 +1,705 @@
...
+    result
+}
+
+fn percent_encode_path(path: &str) -> String {
+    percent_encode(path).replace("%2F", "/")
+}
+
+// ============================================================================
+// Tests
+// ============================================================================
```
cdata: See comment at the top of `s3.rs`

rust/dialog-storage/src/storage/backend/s3/access.rs
```diff
@@ -0,0 +1,705 @@
...
+            Self::AuthenticatedRead => "authenticated-read",
+            Self::BucketOwnerRead => "bucket-owner-read",
+            Self::BucketOwnerFullControl => "bucket-owner-full-control",
+        }
+    }
+}
+
+/// Pre-derived signing key for request signing.
+#[derive(Debug, Clone)]
+struct SessionKey(Vec<u8>);
```
cdata, line 107: Is there coverage for this? (direct or indirect)

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+    bucket: String,
+    /// Optional prefix for all keys
+    prefix: Option<String>,
+    /// Session for authorizing requests
+    session: Session,
+    /// Hasher for computing checksums
+    hasher: Hasher,
+    /// HTTP client
+    client: reqwest::Client,
+    _key: PhantomData<Key>,
```
cdata: Nit: drop the `_` prefix in field names

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+            // WASM doesn't have is_connect() or is_request() methods
+            S3StorageError::RequestFailed(error.to_string())
+        }
+    }
+}
+
+/// Extension trait for requests that can be performed against an S3 backend.
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait Request: Invocation + Sized {
```
cdata: This seems like it should be logically grouped nearby to `Invocation`, and a bit more documented rationale for the use case of these traits would be nice.

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+            return Err(S3StorageError::OperationFailed(format!(
+                "Failed to list objects: {}",
+                response.status()
+            )));
+        }
+
+        let body = response
+            .text()
+            .await
+            .map_err(|e| S3StorageError::RequestFailed(e.to_string()))?;
```
cdata: What is the semantic distinction between `S3StorageError::OperationFailed` and `S3StorageError::RequestFailed`?

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+
+    /// Parse the S3 ListObjectsV2 XML response.
+    fn parse_list_response(xml: &str) -> Result<ListResult, S3StorageError> {
+        // Simple XML parsing for ListObjectsV2 response
+        // Format:
+        // <ListBucketResult>
+        //   <IsTruncated>false</IsTruncated>
+        //   <Contents><Key>...</Key></Contents>
+        //   <NextContinuationToken>...</NextContinuationToken>
+        // </ListBucketResult>
```
cdata: This seems like it belongs in the doc comment

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+            .text()
+            .await
+            .map_err(|e| S3StorageError::RequestFailed(e.to_string()))?;
+
+        // Parse the XML response
+        Self::parse_list_response(&body)
+    }
+
+    /// Parse the S3 ListObjectsV2 XML response.
+    fn parse_list_response(xml: &str) -> Result<ListResult, S3StorageError> {
```
cdata: Given the fact that we are manually implementing XML parsing here, it seems prudent to have some test cases that check this code against malformed (or unexpectedly divergent) responses.

Otherwise, I would be more comfortable if we _didn't_ implement parsing ourselves.

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+}
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+impl<Key, Value> StorageSink for S3<Key, Value>
+where
+    Key: AsRef<[u8]> + Clone + ConditionalSync,
+    Value: AsRef<[u8]> + From<Vec<u8>> + Clone + ConditionalSync,
+{
+    async fn write<S>(&mut self, source: S) -> Result<(), Self::Error>
```
cdata, line 344: Ideally this method is used to implement bulk updates to the backend. Is there some way we can take advantage of the stream to put all the objects in storage at once e.g., with a multipart request?

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+        // S3 doesn't support draining, so just read
+        self.read()
+    }
+}
+
+// ============================================================================
+// Tests
+// ============================================================================
+
+#[cfg(all(test, not(target_arch = "wasm32")))]
```
cdata: At least some (all?) of these tests seem like they should run under the `wasm32-unknown-unknown` target (via `wasm_bindgen_test`). Relevant boilerplate: https://github.com/dialog-db/dialog-db/blob/bf88f2c3313f54e3bd2d89f394f2c4aaf1f1d6c1/rust/dialog-artifacts/src/artifacts.rs#L551-L558

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+            S3::<Vec<u8>, Vec<u8>>::open("https://s3.amazonaws.com", "bucket", Session::Public);
+
+        let url = backend.bucket_url().unwrap();
+        assert_eq!(url.as_str(), "https://s3.amazonaws.com/bucket");
+    }
+}
+
+/// Local S3 server tests using s3s for end-to-end testing
+#[cfg(all(test, not(target_arch = "wasm32")))]
+#[allow(unused_imports, unused_variables, unused_mut, dead_code)]
```
cdata: I suspect that all of these lints should be fixed (so please remove the `#[allow(...)]`)

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+mod local_s3_tests {
```
cdata, line 745: We don't need to implement this in this change, but I would like to get an issue filed to implement an integration test regime similar to what I implemented at Common. 

Macro to generate integration tests: https://github.com/commontoolsinc/system/blob/main/rust/ct-macros/src/testing.rs
Practical usage: https://github.com/commontoolsinc/system/blob/78d5b3ee4d9621fd84821ed0758110b9ac2666dc/rust/common-integration-tests/tests/remote_functions.rs#L27-L91

Effectively the macro (which wants for documentation) generates a native "outer" test that sets up the integration environment by setting up necessary servers and assigning related connection information as environment variables. Then, it invokes the "inner" client test. The client test can read test-specific connection information from the environment. The "outer" test passes or fails on the outcome of the "inner" test.

In practice we used this for running `wasm32-unknown-unknown` integration tests with a live, local backend. But, the same idea could be extended to support arbitrary targets for the "inner" test.

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+    use super::*;
+    use crate::CborEncoder;
+    use s3s::dto::*;
+    use s3s::{S3 as S3Trait, S3Request, S3Response, S3Result};
+    use std::collections::HashMap;
+    use std::sync::{Arc, RwLock};
+    use tokio::net::TcpListener;
+
+    #[tokio::test]
+    async fn test_local_s3_set_and_get() -> anyhow::Result<()> {
```
cdata: Same comment about BDD vs TDD-style test names

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+        let keys: Vec<&[u8]> = items.iter().map(|(k, _)| k.as_slice()).collect();
+        assert!(keys.contains(&b"a".as_slice()));
+        assert!(keys.contains(&b"b".as_slice()));
+
+        Ok(())
+    }
+}
+
+#[cfg(all(any(test, feature = "test-utils"), not(target_arch = "wasm32")))]
+#[allow(unused_imports, unused_variables, unused_mut, dead_code)]
```
cdata: Same comment about lints as above

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+        // Verify the items (order may vary)
+        let keys: Vec<&[u8]> = items.iter().map(|(k, _)| k.as_slice()).collect();
+        assert!(keys.contains(&b"a".as_slice()));
+        assert!(keys.contains(&b"b".as_slice()));
+
+        Ok(())
+    }
+}
+
+#[cfg(all(any(test, feature = "test-utils"), not(target_arch = "wasm32")))]
```
cdata: In other places we move this out into a top-level `helpers.rs` module and have a feature called `helpers`. It would be nice to align with that pattern unless we have a good reason not to.

See: https://github.com/dialog-db/dialog-db/blob/bf88f2c3313f54e3bd2d89f394f2c4aaf1f1d6c1/rust/dialog-artifacts/src/lib.rs#L65-L68

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+                        chunks.extend_from_slice(&bytes);
+                    }
+                }
+                chunks
+            } else {
+                Vec::new()
+            };
+
+            // Calculate MD5 for ETag
+            let e_tag_str = format!("{:x}", md5::compute(&data));
```
cdata: Nit: avoid including obvious type information in variable names

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -0,0 +1,2482 @@
...
+            endpoint,
+            shutdown_tx,
+            storage,
+        })
+    }
+}
+
+// ============================================================================
+// Integration Tests (requires s3_integration_tests feature and env vars)
+// ============================================================================
```
cdata: See comment at top of the file.

.github/workflows/test.yml
```diff
@@ -92,10 +92,10 @@ jobs:
...
       matrix:
         include:
-          - flags: ''
-          - flags: '--release'
-          - flags: '--target wasm32-unknown-unknown'
-          - flags: '--target wasm32-unknown-unknown --release'
+          - flags: "--features s3"
+          - flags: "--features s3 --release"
+          - flags: "--features s3 --target wasm32-unknown-unknown"
+          - flags: "--features s3 --target wasm32-unknown-unknown --release"
```
Gozala:
```suggestion
          - flags: '--features s3'
          - flags: '--features s3 --release'
          - flags: '--features s3 --target wasm32-unknown-unknown'
          - flags: '--features s3 --target wasm32-unknown-unknown --release'
```

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: `Invocation` lives in the different `access` module though and I was trying to avoid circular dependence that is access module should not know or care about `Request`. And it does not require `body` which is why `body` lives here as opposed to in the `Invocation`.

If this does not seem rational maybe suggest what would make more sense. Otherwise I'll just add comments explaining this.

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: I'm not sure there is one. Does that imply I should be a method on `s3` instead or a static function ? Not to say overriding may not be needed in the future, it's more like I don't need it now, but I think it may change when I port transactional memory stuff

cdata: At a high level I don't get why it's a different trait from `Invocation`. At a level below that, it seems like maybe this should be some kind of utility function that works on the Rust ecosystem `http` types.

rust/dialog-storage/src/storage/backend/s3.rs
cdata: It seems like both traits belong in their own module. Broadly, these many-thousand-line modules are a code smell unto themselves and need to be broken up, and these are great candidates for splitting out.

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: Intention was that RequestFailed meant transport level error wheres `OperationFailed` meant S3 operation failed e.g. due to authorization or malformed request etc...

Perhaps it's not used consistently and  naming could be improved. I'll take a look

rust/dialog-storage/src/storage/backend/s3/access.rs
Gozala: [AWS STS](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html) service can be used to generate temporary session credentials. Which I expect we'd want to try as a way of obtaining credentials from the server. It is not for the storage backend though, but this module also not for the storage backend it's for authorizations which happens to be needed by the backend.

rust/dialog-storage/src/storage/backend/s3/access.rs
Gozala: Yes it is how all of the authorization system works and struct is private because it's only used by the Authority::authorize which is tested in this module.

More end to end tests that verify that expected signatures come out are in the s3 tests that check exact URLs. And we also run s3s test server which actually performs validation of pre-signed URLs.

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: I'm not sure I understand what are you implying here ? Are you suggesting that use of `from_utf8_lossy` does that ? What I'm trying to do is take keys that are arbitrary bytes (although I prefer if they were UTF8 strings)  and try to make valid S3 key that is also readable to a degree possible so that when we use blob store we could still have those blobs under say `blob/...` namespace as opposed to unreadable hashes at the root. This is especially helpful when we also store mutable records in the same bucket that have human readable names.

If there is a better way to go about it, I'd be happy to adopt it.

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: I don't believe S3 supports multi-file uploads. Multi-part uploads are for uploading large files in chunks so that uploads could be resume across sessions.

That said we could probably do sets in parallel though as opposed to sequentially.

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: Created https://github.com/dialog-db/dialog-db/issues/113 and will work on that as followup change

rust/dialog-storage/src/storage/backend/s3/access.rs
Gozala: Added comments explaining why it's actually not going to panic.

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: Updated naming and captured semantic differences in the code comments

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: switched to using qick_xml and added bunch more tests

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: Updated all but local_s3_tests because those require running server which will be addressed as followup for https://github.com/dialog-db/dialog-db/issues/113

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: ✅

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: ✅

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: ✅

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: ✅

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: Do you mind if I do that in as a followup pr ?

rust/dialog-storage/src/storage/backend/s3.rs
cdata: `from_utf8_lossy` (as the name implies) is a lossy conversion to UTF-8. Maybe that's acceptable in this case (hence my question).

If you want to convert arbitrary bytes to somewhat readable UTF-8, I recommend using `ToBase58::to_base58` (or else just require UTF-8 strings as the input).

If you want a human-readable prefix, add that separately.

Not knowing the requirements of the key name it's difficult to speculate about the failure conditions of something like this. But, we can reason that:

- The resulting string may not round trip back to the same bytes
- Two different sequences of bytes could become the same key

At the very least I would like to see a clearly articulated comment explaining why that's okay here. But, actually I would just not have to worry about these kinds of problems at all.

PR #114, opened 2025-12-06
feat: dialog_common:test infrastructure

### LINKED ISSUE #113: Test infrastructure for tests with local servers
As per https://github.com/dialog-db/dialog-db/pull/112/files#r2590222699

> We don't need to implement this in this change, but I would like to get an issue filed to implement an integration test regime similar to what I implemented at Common.
>
> Macro to generate integration tests: https://github.com/commontoolsinc/system/blob/main/rust/ct-macros/src/testing.rs
Practical usage: https://github.com/commontoolsinc/system/blob/78d5b3ee4d9621fd84821ed0758110b9ac2666dc/rust/common-integration-tests/tests/remote_functions.rs#L27-L91
>
> Effectively the macro (which wants for documentation) generates a native "outer" test that sets up the integration environment by setting up necessary servers and assigning related connection information as environment variables. Then, it invokes the "inner" client test. The client test can read test-specific connection information from the environment. The "outer" test passes or fails on the outcome of the "inner" test.
>
> In practice we used this for running wasm32-unknown-unknown integration tests with a live, local backend. But, the same idea could be extended to support arbitrary targets for the "inner" test.

Gozala:
fixes #113

### DIFF
--- a/rust/dialog-common/src/lib.rs
+++ b/rust/dialog-common/src/lib.rs
@@ -16,46 +16,82 @@ pub use hash::*;
 #[cfg(feature = "helpers")]
 pub mod helpers;
 
-/// Test macro for cross-platform testing with resource provisioning.
+/// Test macro for cross-platform testing with optional service provisioning.
 ///
-/// # Usage
+/// # Unit Tests
+///
+/// Tests **without parameters** are unit tests. They run on both native and wasm targets.
+///
+/// ```rust
+/// #[dialog_common::test]
+/// fn it_works() {
+///     assert_eq!(2 + 2, 4);
+/// }
+///
+/// #[dialog_common::test]
+/// async fn it_works_async() {
+///     assert_eq!(2 + 2, 4);
+/// }
+/// ```
+///
+/// Run unit tests with:
+/// - `cargo test` - native
+/// - `cargo test --target wasm32-unknown-unknown` - wasm
+///
+/// # Integration Tests
+///
+/// Tests **with a parameter** are integration tests. The parameter type determines
+/// which service is provisioned, and it **must** implement [`helpers::Provisionable`].
+/// Use [`#[dialog_common::provider]`](macro@provider) to generate this implementation.
+///
+/// Service providers (databases, servers, etc.) run on native, so integration tests
+/// require the `integration-tests` or `web-integration-tests` feature flag. With
+/// `web-integration-tests`, the provider runs natively while the test code itself
+/// executes in wasm, allowing you to test browser-specific behavior against real services.
 ///
 /// ```rust
 /// use dialog_common::helpers::Service;
 /// use serde::{Deserialize, Serialize};
 ///
+/// // 1. Define the address type (passed to tests, must be serializable)
 /// #[derive(Debug, Clone, Serialize, Deserialize)]
 /// pub struct Host {
 ///     pub url: String,
 /// }
 ///
-/// #[dialog_common::test]
-/// async fn it_starts_server(host: Host) -> anyhow::Result<()> {
-///     assert!(host.url.starts_with("http://127.0.0.1:"));
-///     Ok(())
-/// }
-///
-/// // Test with custom settings
-/// #[dialog_common::test(port = 8080)]
-/// async fn it_uses_custom_port(host: Host) -> anyhow::Result<()> {
-///   assert!(host.url.contains(":8080"));
-///   Ok(())
-/// }
-///
+/// // 2. Define settings for the provider (must impl Default)
 /// #[derive(Debug, Clone, Default)]
 /// pub struct Settings {
-///     pub port: u16,   // 0 = random available port
+///     pub port: u16,
 /// }
 ///
+/// // 3. Define the provider using #[dialog_common::provider]
+/// //    This generates `impl Provisionable for Host`
 /// #[dialog_common::provider]
 /// async fn tcp_server(settings: Settings) -> anyhow::Result<Service<Host, (std::net::TcpListener,)>> {
 ///     let listener = std::net::TcpListener::bind(format!("127.0.0.1:{}", settings.port))?;
 ///     let addr = listener.local_addr()?;
 ///     Ok(Service::new(Host { url: format!("http://{}", addr) }, (listener,)))
 /// }
+///
+/// // 4. Write integration tests - the parameter makes it an integration test
+/// #[dialog_common::test]
+/// async fn it_starts_server(host: Host) -> anyhow::Result<()> {
+///     assert!(host.url.starts_with("http://127.0.0.1:"));
+///     Ok(())
+/// }
+///
+/// // Provider settings can be customized via macro attributes
+/// #[dialog_common::test(port = 8080)]
+/// async fn it_uses_custom_port(host: Host) -> anyhow::Result<()> {
+///     assert!(host.url.contains(":8080"));
+///     Ok(())
+/// }
 /// ```
 ///
-/// See the macro documentation for the full CI test matrix.
+/// Run integration tests with:
+/// - `cargo test --features integration-tests` - native only
+/// - `cargo test --features web-integration-tests` - spawns wasm subprocess for each test
 #[cfg(feature = "helpers")]
 pub use dialog_macros::test;
 


rust/dialog-common/src/lib.rs
```diff
@@ -1,11 +1,36 @@
...
 
 //! This crate constitutes a library of light weight helpers that are shared
-//! across multiple other crates. Their cheif quality is that they have
+//! across multiple other crates. Their chief quality is that they have
 //! virtually zero dependencies.
 
+// Allow the crate to refer to itself as `dialog_common` in generated macro code.
+// This is needed because our test macros expand to `::dialog_common::helpers::...`.
+#[cfg(feature = "helpers")]
+extern crate self as dialog_common;
```
Gozala:
```suggestion
// Allow the crate to refer to itself as `dialog_common`
extern crate self as dialog_common;
```

Gozala:
> Actually I'm realizing this isn't right outer tests aren't going to load on `--target wasm32-unknown-unknown` so I need a different way to signal target I think.

I have completely redesigned solution in this PR to address this & I think it's ready for the review.

cdata:
Back-linking the original conversation for posterity: https://github.com/dialog-db/dialog-db/pull/112#discussion_r2590222699

rust/dialog-common-macros/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+///
+/// ```rs
+/// #[dialog_common::test(port = 9000, tls = true)]
+/// async fn it_uses_custom_port(server: ServerAddress) -> anyhow::Result<()> {
+///     assert_eq!(server.port, 9000);
+///     Ok(())
+/// }
+/// ```
+#[proc_macro_attribute]
+pub fn test(attr: TokenStream, item: TokenStream) -> TokenStream {
```
cdata: It's a nuanced thing, but I think it would be better to call this a `integration_test`. It makes it immediately clear why this is different from other tests (and implies that it is a most costly test to add, which it is). Also: we may want to add a `test` macro at some point to spit out all the boilerplate that we currently add to most unit tests, and at that point the distinction will be important.

rust/dialog-common-macros/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+/// use serde::{Deserialize, Serialize};
+///
+/// #[derive(Debug, Clone, Serialize, Deserialize)]
+/// pub struct ServerAddress {
+///     pub endpoint: String,
+///     pub port: u16,
+/// }
+/// ```
+///
+/// ### Step 2: Define the provider (native only)
```
cdata: :gem:

rust/dialog-common-macros/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+///     Ok(())
+/// }
+/// ```
+#[proc_macro_attribute]
+pub fn test(attr: TokenStream, item: TokenStream) -> TokenStream {
+    test::generate(attr, item)
+}
+
+// disabling because we don't want to add crate dependencies just for this
+#[cfg(not(doctest))]
```
cdata: Is there a reason you wanted to do this rather than use `ignore` as the code block tag?

rust/dialog-common-macros/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+///
+/// // Async test
+/// #[dialog_common::test]
+/// async fn it_works_async() -> anyhow::Result<()> {
+///     assert_eq!(2 + 2, 4);
+///     Ok(())
+/// }
+/// ```
+///
+/// Unit tests are gated with `#[cfg(not(dialog_test_wasm))]` so they don't
```
cdata: Can we make the gate more explicit? `dialog_integration_test_wasm`. Also: what if we wanted to run integration tests in not-Wasm?

rust/dialog-common-macros/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+#[cfg(not(doctest))]
+/// A cross-platform test macro with automatic service provisioning.
+///
+/// # CI Test Matrix
+///
+/// The macro generates code that supports these CI configurations:
+///
+/// 1. `cargo test` - All tests run natively (unit tests + integration tests inline)
+/// 2. `cargo test --target wasm32-unknown-unknown` - Unit tests run in wasm
+/// 3. `RUSTFLAGS="--cfg dialog_test_wasm" cargo test` - Integration tests run in wasm
```
cdata: Even if we only support integration tests with `wasm32-unknown-unknown` inner tests today, in the fullness of time I think we want to be able to run the inner test as a native target test as well. Our names should reflect that, which is to say we shouldn't bake-in the assumption that `wasm` is a critical detail.

rust/dialog-common-macros/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+///
+/// When a function takes an address parameter, the macro generates:
+///
+/// - **Native inline provider**: Starts service, runs test via `tokio::spawn`, stops service
+/// - **Native wasm-spawn provider**: Starts service, spawns `cargo test --target wasm32...`
+/// - **Wasm inner**: Deserializes address from env var, runs test
+///
+/// ```rs
+/// #[dialog_common::test]
+/// async fn it_connects_to_server(server: ServerAddress) -> anyhow::Result<()> {
```
cdata: Can you make it explicit in the docs that the parameter is what turns this into an integration test? It took me a a fair amount of squinting to infer that.

rust/dialog-macros/src/test.rs
```diff
@@ -0,0 +1,624 @@
...
+//!
+//! This module provides the implementation for cross-platform test macros
+//! that handle service provisioning and multi-target testing.
+//!
+//! # CI Test Matrix
+//!
+//! The macro is designed to support this CI workflow:
+//!
+//! 1. `cargo test` - Run all tests natively (unit tests + integration tests inline)
+//! 2. `cargo test --target wasm32-unknown-unknown` - Run unit tests in wasm
```
cdata, line 11: Can you explain here what it will do if I pass an explicit target that isn't `wasm32-unknown-unknown` or the native target e.g., `wasm-wasip3` or `aarch64-unknown-linux-gnu`?

rust/dialog-common-macros/src/test.rs
```diff
@@ -0,0 +1,624 @@
...
+//! This module provides the implementation for cross-platform test macros
+//! that handle service provisioning and multi-target testing.
+//!
+//! # CI Test Matrix
+//!
+//! The macro is designed to support this CI workflow:
+//!
+//! 1. `cargo test` - Run all tests natively (unit tests + integration tests inline)
+//! 2. `cargo test --target wasm32-unknown-unknown` - Run unit tests in wasm
+//! 3. `RUSTFLAGS="--cfg dialog_test_wasm" cargo test` - Run integration tests in wasm
```
cdata: Just staring at this, I think it's more obvious/idiomatic for someone to invoke the test command with a feature e.g., `integration-tests` than to set a rustc flag. The downside is that they need to add it to their `Cargo.toml`. But, maybe it's not such a bad trade-off.

rust/dialog-common-macros/src/test.rs
```diff
@@ -0,0 +1,624 @@
...
+///     Ok(())
+/// }
+///
+/// // 3. Wasm integration test - runs when `RUSTFLAGS="--cfg dialog_test_wasm" cargo test`
+/// #[cfg(all(dialog_test_wasm, not(target_arch = "wasm32")))]
+/// #[tokio::test]
+/// async fn it_connects() -> anyhow::Result<()> {
+///     let service = ServerAddress::start(Default::default()).await?;
+///     let json = serde_json::to_string(&service.address)?;
+///     // Spawns: RUSTFLAGS="--cfg dialog_test_wasm_inner" \
```
cdata: Setting the `rustc` flag here isn't so bad because it's a generated test command.

rust/dialog-common-macros/src/test.rs
```diff
@@ -0,0 +1,624 @@
...
+/// async fn it_connects(server: ServerAddress) -> anyhow::Result<()> {
+///     assert!(!server.endpoint.is_empty());
+///     Ok(())
+/// }
+/// ```
+///
+/// Generates (simplified, with hash `abc123`):
+/// ```rs
+/// // 1. Integration logic - the actual test body (allow dead_code when dialog_test_wasm)
+/// #[cfg_attr(dialog_test_wasm, allow(dead_code))]
```
cdata: If you follow my suggestion above re: using a feature instead of a `rustc` flag for the outer, then this codegen would be:

```rs
#[cfg(feature = "integration-tests", allow(dead_code))]
```

rust/dialog-common-macros/src/test.rs
```diff
@@ -0,0 +1,624 @@
...
+/// async fn it_connects_logic_abc123(server: ServerAddress) -> anyhow::Result<()> {
+///     assert!(!server.endpoint.is_empty());
+///     Ok(())
+/// }
+///
+/// // 2. Native test - runs when `cargo test` (no flags)
+/// #[cfg(all(not(dialog_test_wasm), not(target_arch = "wasm32")))]
```
cdata: Similarly:

```rs
#[cfg(all(not(feature = "integration-tests"), not(target_arch = "wasm32")))]
```

rust/dialog-common-macros/src/test.rs
```diff
@@ -0,0 +1,624 @@
...
+/// async fn it_connects() -> anyhow::Result<()> {
+///     let service = ServerAddress::start(Default::default()).await?;
+///     let result = tokio::spawn(it_connects_logic_abc123(service.address.clone())).await;
+///     service.stop().await?;
+///     result?;
+///     Ok(())
+/// }
+///
+/// // 3. Wasm integration test - runs when `RUSTFLAGS="--cfg dialog_test_wasm" cargo test`
+/// #[cfg(all(dialog_test_wasm, not(target_arch = "wasm32")))]
```
cdata: Similarly:

```rs
#[cfg(all(feature = "integration-tests", not(target_arch = "wasm32")))]
```

rust/dialog-common-macros/src/lib.rs
Gozala: ignore puts things in tests that make it feel like there are ignored tests, so this seemed like more appropriate, but happy to put ignore if you prefer.

cdata: > ignore puts things in tests that make it feel like there are ignored tests

That is a feature. Ideally these code samples would be run as doctests (at the very least, the compiler should check them). The test output is a good way to note headroom for future improvements.

Gozala:
> nit: Let's just call the macro crate `dialog-macros` until we have a reason for multiple macro crates to exist.

I do have dialog-query-macros in the sync branch, would they also go into `dialog-macros` or would that be a reason enough not to do `dialog-macros` ?

rust/dialog-common-macros/src/lib.rs
Gozala: We actually do run inner test on native already, and this flag is to disable all tests but the ones that need to run inner tests in wasm.

That said do you have a better name suggestion ? I'm not happy with this one either, but don't know what would be a good one

rust/dialog-common-macros/src/lib.rs
Gozala: I added a comment on line 126 why I though it was better also in this crate specifically examples creates circular dependency, that is to say I think making this example compile introduced complication that I was not sure was worth it.

I guess it would be good to understand general preference on when to fallback to

1. no_run
2. ignore
3. something else

rust/dialog-common-macros/src/lib.rs
cdata: Thanks, I misconstrued that.

I guess with the added context I would go with something more like `web-integration-tests` (I assume that the `wasm-bindgen-test` dependency is implied anyway, which only has meaning for Wasm targeting web browsers).

cdata:
> I do have dialog-query-macros in the sync branch, would they also go into dialog-macros or would that be a reason enough not to do dialog-macros ?

I guess it's a question of our discretion. The primary reason to split the macros into different crates is that it might reduce compile time depending on the weight of the different macros. But, you can also address that with feature flags.

Personally I would rather start with a single crate until it gets too unwieldy or confusing to group everything up that way.

rust/dialog-common-macros/src/test.rs
Gozala: I think `integration-tests` is bit misleading as I would expect it to run all integration tests on the target runtime in this specific case I'd expect it to run only tests that have a parameter on the native target as opposed to running them on wasm target.

My intention of having `wasm` in the name was to make that clear

rust/dialog-common-macros/src/lib.rs
cdata: My preference is almost always: if there is a code sample in docs, it should at least be checked by the compiler (`no_run`). I can't think of an exception to that off of the top of my head. The reason is: at some point the example will fall out of date, and no-one will know unless we have some kind of check for it. Better not to have an example at all than a misleading, out-of-date one.

PR #117, opened 2025-12-08
port differential

Gozala:
Ports difference computation and integration logic from sync branch and bases it on top of #116

P.R: Introduces fairly complex `tree_spec!` macro, probably most complicated piece, to make it easier to express desired behavior of differentiation, visualize it and also capture what subtrees should be skipped vs traversed.

### DIFF
--- a/.github/workflows/s3-integration-tests.yml
+++ b/.github/workflows/s3-integration-tests.yml
@@ -1,0 +1,106 @@
+name: 'S3 Integration Tests'
+
+on:
+  workflow_dispatch:
+    inputs:
+      provider:
+        description: 'Storage provider to test'
+        required: true
+        default: 'both'
+        type: choice
+        options:
+          - r2
+          - s3
+          - both
+  push:
+    branches:
+      - main
+      - patches/*
+    paths:
+      - 'rust/dialog-storage/src/storage/backend/s3.rs'
+      - 'rust/dialog-storage/src/storage/backend/s3/**'
+  pull_request:
+    paths:
+      - 'rust/dialog-storage/src/storage/backend/s3.rs'
+      - 'rust/dialog-storage/src/storage/backend/s3/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event.inputs.provider || 'auto' }}
+  cancel-in-progress: true
+
+jobs:
+  integration-tests:
+    name: '${{ matrix.provider }} (${{ matrix.name }})'
+    runs-on: 'ubuntu-latest'
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - provider: r2
+            name: native
+            target: ''
+            endpoint_var: R2_ENDPOINT
+            region_var: R2_REGION
+            bucket_var: R2_BUCKET
+            access_key_var: R2_ACCESS_KEY_ID
+            secret_key: R2_SECRET_ACCESS_KEY
+          - provider: r2
+            name: wasm
+            target: '--target wasm32-unknown-unknown'
+            endpoint_var: R2_ENDPOINT
+            region_var: R2_REGION
+            bucket_var: R2_BUCKET
+            access_key_var: R2_ACCESS_KEY_ID
+            secret_key: R2_SECRET_ACCESS_KEY
+          - provider: s3
+            name: native
+            target: ''
+            endpoint_var: S3_ENDPOINT
+            region_var: S3_REGION
+            bucket_var: S3_BUCKET
+            access_key_var: S3_ACCESS_KEY_ID
+            secret_key: S3_SECRET_ACCESS_KEY
+          - provider: s3
+            name: wasm
+            target: '--target wasm32-unknown-unknown'
+            endpoint_var: S3_ENDPOINT
+            region_var: S3_REGION
+            bucket_var: S3_BUCKET
+            access_key_var: S3_ACCESS_KEY_ID
+            secret_key: S3_SECRET_ACCESS_KEY
+    steps:
+      - name: 'Check if should run'
+        id: should-run
+        run: |
+          if [[ "${{ github.event_name }}" != "workflow_dispatch" ]]; then
+            echo "run=true" >> $GITHUB_OUTPUT
+          elif [[ "${{ github.event.inputs.provider }}" == "both" ]]; then
+            echo "run=true" >> $GITHUB_OUTPUT
+          elif [[ "${{ github.event.inputs.provider }}" == "${{ matrix.provider }}" ]]; then
+            echo "run=true" >> $GITHUB_OUTPUT
+          else
+            echo "run=false" >> $GITHUB_OUTPUT
+          fi
+      - uses: actions/checkout@v4
+        if: steps.should-run.outputs.run == 'true'
+      - uses: ./.github/actions/setup-nix
+        if: steps.should-run.outputs.run == 'true'
+      - name: 'Remove system Chrome (for wasm tests)'
+        if: steps.should-run.outputs.run == 'true' && matrix.target != ''
+        run: |
+          sudo apt-get remove google-chrome-stable -y || true
+          sudo rm -f /usr/bin/chromium /usr/bin/chromium-browser
+      - name: 'Run integration tests'
+        if: steps.should-run.outputs.run == 'true'
+        env:
+          R2S3_ENDPOINT: ${{ vars[matrix.endpoint_var] }}
+          R2S3_REGION: ${{ vars[matrix.region_var] }}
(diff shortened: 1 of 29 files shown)


rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -0,0 +1,1907 @@
...
+//!
+//! # Core Concepts
+//!
+//! - [`Change`]: Represents an addition or removal of an entry
+//! - [`Differential`]: A stream of changes between two trees
+//! - [`Delta`]: Internal structure for computing differences
+//!
+//! # Usage
+//!
+//! Use [`Tree::differentiate`](crate::Tree::differentiate) to compute changes between trees,
```
cdata, line 19: It's a little confusing that this documentation refers to `Tree` (which lives in a different module). Maybe it belongs in that other module (and this documentation can point the user there for usage instructions).

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -0,0 +1,1907 @@
...
+    Key: KeyType + 'static,
+    Value: ValueType,
+    T: Stream<Item = Result<Change<Key, Value>, DialogProllyTreeError>>,
+{
+}
+
+/// Helper struct to convert Vec<Change<Key, Value>> into a Differential stream.
+///
+/// This allows using a pre-computed vector of changes as a differential.
+pub struct VecDifferential<Key, Value>
```
cdata: Can this whole thing be replaced with [`futures_util::stream::iter`](https://docs.rs/futures-util/latest/futures_util/stream/fn.iter.html)?

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -0,0 +1,1907 @@
...
+where
+    Key: KeyType + 'static,
+    Value: ValueType,
+    Hash: HashType,
+{
+    /// Get the upper bound of this node or reference
+    fn upper_bound(&self) -> &Key {
+        match self {
+            SparseTreeNode::Node(node) => node.upper_bound(),
+            SparseTreeNode::Ref(ref_) => ref_.upper_bound(),
```
cdata: I'm not sure what this `ref_` convention is. If you want to use a Rust keyword as an identifier, use `r#ref`. But, my personal preference would be to simply use the full word `reference`.

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -0,0 +1,1907 @@
...
+    /// Get the hash of this node or reference
+    fn hash(&self) -> &Hash {
+        match self {
+            SparseTreeNode::Node(node) => node.hash(),
+            SparseTreeNode::Ref(ref_) => ref_.hash(),
+        }
+    }
+
+    /// Load this as a node (returns the node if already loaded, or loads from storage)
+    async fn load<Storage>(
```
cdata: Nit: a name like `ensure_loaded` would make it a little more clear that the IO is conditional.

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -0,0 +1,1907 @@
...
+        self,
+        storage: &Storage,
+    ) -> Result<Node<Key, Value, Hash>, DialogProllyTreeError>
+    where
+        Storage: ContentAddressedStorage<Hash = Hash>,
+    {
+        match self {
+            SparseTreeNode::Node(node) => Ok(node),
+            SparseTreeNode::Ref(ref_) => Node::from_reference(ref_, storage).await,
```
cdata: Same comment as above regarding naming convention

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -0,0 +1,1907 @@
...
+            SparseTreeNode::Node(node) => node.upper_bound(),
+            SparseTreeNode::Ref(ref_) => ref_.upper_bound(),
+        }
+    }
+
+    /// Get the hash of this node or reference
+    fn hash(&self) -> &Hash {
+        match self {
+            SparseTreeNode::Node(node) => node.hash(),
+            SparseTreeNode::Ref(ref_) => ref_.hash(),
```
cdata: Same comment as above regarding naming convention

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -0,0 +1,1907 @@
...
+/// A sparse view of a tree containing only the nodes that differ between two trees.
+#[derive(Debug, Clone)]
+pub(crate) struct SparseTree<'a, Key, Value, Hash, Storage>
+where
+    Key: KeyType + 'static,
+    Value: ValueType,
+    Hash: HashType,
+    Storage: ContentAddressedStorage<Hash = Hash>,
+{
+    storage: &'a Storage,
```
cdata, line 204: FWIW It tends to be the case that our storage abstractions are cheaply cloneable; you may be able to forgo the lifetime here by assuming owned `Storage`.

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -0,0 +1,1907 @@
...
+}
+
+impl<Key, Value, Hash, Storage> SparseTree<'_, Key, Value, Hash, Storage>
+where
+    Key: KeyType + 'static,
+    Value: ValueType,
+    Hash: HashType,
+    Storage: ContentAddressedStorage<Hash = Hash>,
+{
+    /// Expands loaded branch nodes and Refs whose upper bound falls within the
```
cdata: Nit: weird that 'nodes' is not capitalized but 'Refs' is. Make it consistent!

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -0,0 +1,1907 @@
...
+    Key: KeyType + 'static,
+    Value: ValueType,
+    Hash: HashType,
+    Storage: ContentAddressedStorage<Hash = Hash>,
+{
+    storage: &'a Storage,
+    nodes: Vec<SparseTreeNode<Key, Value, Hash>>,
+}
+
+impl<Key, Value, Hash, Storage> SparseTree<'_, Key, Value, Hash, Storage>
```
cdata, line 213: We don't have to cover it in this change, but I suspect that this could more clearly be expressed as a `Difference` struct that aggregates the delta between two trees, as opposed to a differently-named `Tree` with an implied alternative use case. Perhaps we can revisit the abstraction as we integrate this capability into the `dialog-search-tree`.

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -0,0 +1,1907 @@
...
+                storage: right.storage(),
+                nodes: right
+                    .root()
+                    .map(|root| vec![SparseTreeNode::Node(root.clone())])
+                    .unwrap_or(vec![]),
+            },
+        )
+    }
+
+    /// Expands the difference by intelligently expanding only the nodes that
```
cdata: nit:

"Expands ... by intelligently expanding ..."

Repetitive use of "expand" that doesn't add to the reader's understanding; I would prefer "efficiently" to "intelligently".

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -0,0 +1,1907 @@
...
+        }
+
+        Ok(())
+    }
+
+    /// Returns a stream of changes between the two trees.
+    ///
+    /// This performs a two-cursor walk over the entry streams from both sparse
+    /// trees, yielding Add and Remove changes as appropriate.
+    pub fn stream(&'a self) -> impl Differential<Key, Value> + 'a {
```
cdata: Per feedback you gave me in #129 , why not use a more contextual verb here?

cdata (approved):
Almost all nits, but I would like to revisit these abstractions as we integrate them into the `dialog-search-tree` crate proposed in #129

rust/dialog-prolly-tree/src/helpers.rs
```diff
@@ -0,0 +1,887 @@
...
+/// let tree = spec_a.tree().clone();
+/// let delta = tree.differentiate(spec_b.tree());
+/// ```
+///
+/// In this example:
+/// - Tree A has keys: a, b, c, d, e, f, g, h, i, j, k, l
+/// - Tree B has keys: f, g, h, i, j, k, l, m, n, o, p, q, r, s
+/// - Trees overlap in keys f-l, differ in a-e (only in A) and m-s (only in B)
+#[macro_export]
+macro_rules! tree_spec {
```
cdata, line 1068: 💎

rust/dialog-prolly-tree/src/differential.rs
Gozala: Sure I'm happy to revisit this with new tree integration. I have explored various approaches but found all of them to be more complicated 🥹

PR #119, opened 2025-12-09
Transactional memory

Gozala:
Ports transactional memory from sync to main, based on #120

### DIFF
--- a/rust/dialog-storage/tests/s3_integration_tests/bucket.rs
+++ b/rust/dialog-storage/tests/s3_integration_tests/bucket.rs
@@ -1,7 +1,7 @@
 use dialog_storage::s3::{Address, Bucket, Credentials};
 
-/// Generate a globally unique test prefix using timestamp
-pub fn unique_prefix(base: &str) -> String {
+/// Adds timestamp to the given string to make it unique
+pub fn unique(base: &str) -> String {
     #[cfg(not(target_arch = "wasm32"))]
     let millis = std::time::SystemTime::now()
         .duration_since(std::time::UNIX_EPOCH)
@@ -45,5 +45,5 @@ pub fn open() -> Bucket<Vec<u8>, Vec<u8>> {
 }
 
 pub fn open_unque_at(base: &str) -> Bucket<Vec<u8>, Vec<u8>> {
-    open().at(unique_prefix(base))
+    open().at(unique(base))
 }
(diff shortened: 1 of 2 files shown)


rust/dialog-storage/src/storage/transactional_memory.rs
```diff
@@ -0,0 +1,810 @@
...
+        map.finish()
+    }
+}
+
+/// Policy controlling retry behavior for transactional operations.
+///
+/// Determines how many times to retry a failed Compare-And-Swap (CAS) operation
+/// after reloading fresh state from storage.
+#[derive(Debug, Clone)]
+pub enum UpdatePolicy {
```
cdata, line 115: Can you offer some speculative future policies to justify why this is an `enum`?

rust/dialog-storage/src/storage/backend/memory.rs
```diff
@@ -8,7 +8,15 @@ use tokio::sync::RwLock;
...
 use crate::{DialogStorageError, StorageSource};
 
-use super::StorageBackend;
+use super::{StorageBackend, TransactionalMemoryBackend};
+
+/// A 32-byte BLAKE3 content hash used as the edition for CAS operations.
+type ContentHash = [u8; 32];
+
+/// Compute BLAKE3 hash of content.
+fn content_hash(content: &[u8]) -> ContentHash {
```
cdata: Please prefer `dialog_common::Blake3Hash`

rust/dialog-storage/src/storage/backend/indexeddb.rs
```diff
@@ -116,6 +140,160 @@ where
...
+/// the current stored value matches the expected edition. This provides
+/// content-addressable semantics - identical values have identical editions.
+///
+/// # Cross-Tab Safety
+///
+/// IndexedDB transactions are atomic and serialized across tabs. The `replace`
+/// operation performs read and write within a single `ReadWrite` transaction,
+/// so concurrent writes from multiple tabs are safely serialized by IndexedDB.
+#[async_trait(?Send)]
+impl<Key, Value> TransactionalMemoryBackend for IndexedDbStorageBackend<Key, Value>
```
cdata, line 159: 🤔 I wonder if it would be possible to implement this generically for all `T: StorageBackend` if we extended `StorageBackend` to support locking paths...

cdata (approved):
Looks great, minor feedback

rust/dialog-storage/src/storage/transactional_memory.rs
Gozala: I would like to backport https://www.npmjs.com/package/pollicy at some point

rust/dialog-storage/src/storage/backend/indexeddb.rs
Gozala: I've considered but decided it was not worth complicating StorageBackend

PR #120, opened 2025-12-12
S3 backend use virtual host

Gozala:
Turns out S3 has migrated to [virtual-hosted](https://stackoverflow.com/questions/47755043/what-is-difference-between-aws-s3-virtual-hosted-style-url-v-s-path-style-url) style URLs from path styled one which made this implementation incompatible with AWS.

> ℹ️ @cdata In the call I made invalid claim that R2 and other services use path styled URLs. Turns out that was false and instead R2 supports both.

This PR migrates to virtual-hosted URLs to address this issue. There are also few simplifications introduce here

1. I have simplified `access.rs` module getting rid of `Authority` and `Session` constructs, because I wrongly assumed that it was possible to derive `Session` credentials that were time bound with the goal of deriving them on the server and passing them onto client. Since time bounds aren't actually possible those abstractions only add complexity so I got rid of the and now there's just `Credentials` left.
2. I renamed `S3` to `Bucket` to better reflect what it is.
3. Removal of `Session` and `Authority` caused some information to move into Request objects instead. 

> ℹ️ Changes in PR needs to be grafted over https://github.com/dialog-db/dialog-db/pull/122 once they're main, but I don't expect significant changes to actual delta


rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -1,15 +1,15 @@
 //! S3-compatible storage backend for AWS S3, Cloudflare R2, and other S3-compatible services.
 //!
-//! This module provides an [`S3`] storage backend that implements [`StorageBackend`],
-//! allowing you to use S3-compatible object storage as a key-value store.
+//! This module provides [`Bucket`] providing [`StorageBackend`] implementation
```
cdata: nit: "provides... providing" fluffy redundant use of language

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -166,108 +188,284 @@ impl From<reqwest::Error> for S3StorageError {
...
+    /// Use path-style URLs (bucket in path) vs virtual-hosted (bucket in subdomain)
+    /// Defaults to true for IP addresses and localhost, false otherwise.
+    path_style: bool,
+    /// Credentials for authorizing requests (None for public access)
+    credentials: Option<Credentials>,
     /// Hasher for computing checksums
     hasher: Hasher,
     /// HTTP client
-    client: reqwest::Client,
+    pub(crate) client: reqwest::Client,
```
cdata: nit: prefer an accessor in this case

rust/dialog-storage/src/storage/backend/s3/access.rs
```diff
@@ -350,15 +157,107 @@ impl Authority {
...
+    /// Sign data using this key.
+    fn sign(&self, data: &[u8]) -> Signature {
+        Signature(Self::hmac(&self.0, data))
     }
 }
 
+/// Authorize a public request.
+///
+/// Adds required headers (host, checksum) without signing.
+pub fn unauthorized<I: Invocation>(request: &I) -> Result<Authorization, AuthorizationError> {
```
cdata: Nit: the name here is kind of confusing. If I understand the documentation correctily, `unauthorized` is describing the state of the input parameter at call time. I would prefer a verb that is shorthand for what the function will actually do to the input parameter.

PR #127, opened 2025-12-19
feat: extend differential logic to detect tree novelty (new nodes)

Gozala:
Extends #117 logic so that `TreeDifference` can be used to enumerate all the nodes in the `target` tree that does not exists in the `base` tree.

This could be merged into #117 to reviewed together or reviewed separately to keep deltas small.

### DIFF
--- a/rust/dialog-prolly-tree/src/differential.rs
+++ b/rust/dialog-prolly-tree/src/differential.rs
@@ -223,15 +223,6 @@ where
 /// 3. **In-place pruning**: Shared nodes (same boundary + same hash) can be removed
 ///    from both trees simultaneously, leaving only differing nodes
 ///
-/// # Fields
-///
-/// - `nodes`: The current "frontier" of the sparse tree. Initially contains just
-///   the root. After expansion and pruning, contains only nodes that differ from
-///   the other tree being compared against.
-/// - `expanded`: Tracks branch nodes that were loaded and expanded during
-///   comparison. These are "novel" nodes that exist in this tree but not in the
-///   source tree (used by [`TreeDifference::novel_nodes`]).
-///
 /// # Algorithm Overview
 ///
 /// When comparing trees A (source) and B (target):
@@ -565,7 +556,8 @@ where
 /// TreeDifference contains sparse representations of both trees and can be used
 /// to produce either:
 ///
-/// - Entry-level changes via [`changes()`](Self::changes) - transforms source into target
+/// - Entry-level changes via [`changes()`](Self::changes) - can be applied to
+/// transform source into target
 /// - Node-level novelty via [`novel_nodes()`](Self::novel_nodes) - nodes in target abscent in source
 ///
 /// # Usage
@@ -1571,9 +1563,7 @@ mod tests {
         }
     }
 
-    // ========================================================================
     // Edge case tests
-    // ========================================================================
 
     #[dialog_common::test]
     async fn test_differentiate_both_empty() {


rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -123,8 +139,28 @@ where
...
+/// Instead, we maintain a "sparse" representation where nodes can be either:
+///
+/// - **Loaded (`Node`)**: The full node data is in memory, either because it was
+///   the root (already available) or because we needed to inspect its children.
+/// - **Unloaded (`Ref`)**: We only have the hash and upper bound (from a parent's
+///   child reference list). The actual node data remains in storage until needed.
+///
+/// This lazy loading is crucial for efficiency: when comparing two trees that share
+/// large subtrees, we can detect equality by comparing hashes without ever loading
+/// the shared nodes. Only nodes that differ need to be loaded and expanded.
```
cdata, line 154: Just want to observe: the `Tree` _already_ lazy loads, so it might be useful to call out why this lazy loading is not the same as the lazy loading in `Tree` (or else why `Tree`'s implementation is otherwise insufficient).

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -175,7 +211,52 @@ where
...
+/// sorted by their upper bounds. This structure enables efficient comparison
+/// between two trees by:
+///
+/// 1. **Lazy loading**: Nodes start as references and are only loaded when needed
+/// 2. **Level-by-level expansion**: Branch nodes can be "expanded" to reveal their
+///    children, replacing one entry with multiple child entries
+/// 3. **In-place pruning**: Shared nodes (same boundary + same hash) can be removed
+///    from both trees simultaneously, leaving only differing nodes
+///
+/// # Fields
```
cdata: Put field documentation on the fields

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -470,94 +560,73 @@ where
...
-/// target tree.
+/// TreeDifference contains sparse representations of both trees and can be used
+/// to produce either:
 ///
-/// The tuple structure is `(source_sparse_tree, target_sparse_tree)`.
-pub(crate) struct Delta<'a, Key, Value, Hash, Storage>(
-    SparseTree<'a, Key, Value, Hash, Storage>,
-    SparseTree<'a, Key, Value, Hash, Storage>,
-)
+/// - Entry-level changes via [`changes()`](Self::changes) - transforms source into target
```
cdata: Is this more accurately described as "entries needed to transform source into target"...?

rust/dialog-prolly-tree/src/differential.rs
Gozala, line 154:
```suggestion
/// This lazy loading is crucial for efficiency: when comparing two trees that share
/// large subtrees which are not replicated, as we can detect equality by comparing
/// hashes without fetching shared nodes over the network. Only nodes that differ
/// need to be loaded and expanded.
```

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -175,7 +211,52 @@ where
...
+///
+/// # Fields
+///
+/// - `nodes`: The current "frontier" of the sparse tree. Initially contains just
+///   the root. After expansion and pruning, contains only nodes that differ from
+///   the other tree being compared against.
+/// - `expanded`: Tracks branch nodes that were loaded and expanded during
+///   comparison. These are "novel" nodes that exist in this tree but not in the
+///   source tree (used by [`TreeDifference::novel_nodes`]).
+///
```
Gozala:
```suggestion
```

rust/dialog-prolly-tree/src/differential.rs
Gozala: Not exactly because 
```suggestion
/// - Entry-level changes via [`changes()`](Self::changes) - can be applied to
/// transform source into target
```

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -1406,10 +1571,11 @@ mod tests {
         }
     }
 
+    // ========================================================================
```
Gozala:
```suggestion
```

rust/dialog-prolly-tree/src/differential.rs
```diff
@@ -1406,10 +1571,11 @@ mod tests {
...
     // Edge case tests
+    // ========================================================================
```
Gozala:
```suggestion
```

PR #129, opened 2025-12-30
feat: Revised Search Tree

cdata:
This change represents a major revision to the search tree implementation in the existing crate `dialog-prolly-tree`. It is implemented as a new crate - `dialog-search-tree` - so that we can get it in and perform the replacement of the old tree as a follow-on exercise.

The main goal of this revised implementation is speed improvements. The previous tree was fast enough for certain demos, but it had some serious bottlenecks that hindered practical usage. In one notable case, performing many writes to the tree in one go was unacceptably slow (the time to import ~10,000 facts was measured in seconds).

The new tree applies the following optimizations:

- Tree nodes are (de)serialized using "total" zero-copy deserialization (via [`rkyv`](https://rkyv.org/))
  - There is [nearly zero deserialization cost for reads](https://rkyv.org/zero-copy-deserialization.html#total-zero-copy)
  - [Deserialization for writes is 3 ~ 4 x faster](https://github.com/djkoloski/rust_serialization_benchmark) than `serde_cbor` (used by `dialog-prolly-tree`)
- Node buffers are retained by the `Tree` in a cache that is shared by all branches from that `Tree`
- Node buffer hashes are computed at once most for a cached buffer
- `Storage` is _only_ used to populate cache when a node's block is otherwise not available in memory
  - The `Tree` _never_ writes to `Storage`
  - Instead, an internal buffer of new blocks needed to read the tree at the current state (relative to its base) is maintained and must be manually flushed to storage by the `Tree` user
    - In other words: the `Tree` user has full control over when (probably slow) writes happen, without compromising on the ability to read from a modified `Tree` 

Additionally, the new `Tree`:

- ...no longer supports arbitrary hashes
- ...no longer supports arbitrary branch factors
- ...depends on a simplified, in-crate `ContentAddressedStorage` wrapper over some `impl StorageBackend`
  - We should reconcile this trait with `dialog-storage` in a future change
- ...no longer owns a `Storage`
  - The user must bring their own `Storage` for each API interaction

### Abandoned Attempts

This change is the culmination of a few months of experimenting with a zero-copy tree that used the specialized encoding we landed in https://github.com/dialog-db/dialog-db/pull/95. All-told, I made five attempts to implement a satisfying tree that made use of `dialog-encoding` while also supporting fast, zero-copy reads. These attempts all ran aground against two constraints:

- Deserialization was always much slower than intended, dragged down by manually reassembling read-only structures from fragmented data
- Despite the deserialization overhead, it was not possible to modify the buffers in-place for efficient writes

My conclusion is that we need to revisit `dialog-encoding` with this experience in mind before we attempt another [Fressian](https://github.com/Datomic/fressian)-style (de)serialization scheme.

For now, `rkyv` is very fast for (de)serialization, and we can pair it with other compression schemes to get most of the benefits that we sought from `dialog-encoding`.

### Migration Outlook

In principle there should not be a reason why this new `Tree` is difficult to migrate to. The API embodies all the same verbs and semantics as the current one. However, there are some nuanced details related to `Storage` that need to be considered. The user of the `Tree` is now responsible for passing `Storage` to the `Tree` when reading and writing, and then also must handle persisting changes over time.

## Benchmark Results & Charts

As part of this change, I introduced some benchmarks to cover core API operations for `dialog-search-tree` (including backported analogous benchmarks for `dialog-prolly-tree`). I'm excited to report that  `dialog-search-tree` outperforms `dialog-prolly-tree` in all test scenarios. Here are the approximate speed improvements:

| Operation            | Change           |
|----------------------|------------------|
| Read (Single)        | 7 ~ 45 x faster  |
| Read (Range)         | 1.5 ~ 7 x faster  |
| Insert (Sequential)  | 2 ~ 4 x faster  |
| Insert (Random)      | 2 ~ 4 x faster  |
| Delete               | 2 ~ 5 x faster |

<details>
  <summary><strong><em>Read (Single)</em></strong></summary>

_N = size of tree_

| Size  | 🔵 Prolly Tree | 🟢 Search Tree | Speedup |
|-------|----------------|----------------|---------|
| 10    | 1.27 us        | 0.18 us        | 7.0x    |
| 100   | 3.61 us        | 0.31 us        | 11.7x   |
| 1000  | 36.91 us       | 0.81 us        | 45.7x   |
| 10000 | 54.00 us       | 1.29 us        | 41.8x   |

```mermaid
xychart-beta
    title "Read Performance (Single Key Lookup)"
    x-axis [10, 100, 1000, 10000]
    y-axis "Time (us, mean)" 0 --> 55
    line [1.27, 3.61, 36.91, 54.00]
    line [0.18, 0.31, 0.81, 1.29]
```
</details>

<details>
  <summary><strong><em>Read (Range)</em></strong></summary>

_N = number of entries streamed from 10100-sized tree_

_Note: I ran the benchmark a few times; I'm not sure why there is a spike in read time (in both crates) for N=100._

| Size  | 🔵 Prolly Tree | 🟢 Search Tree | Speedup |
|-------|----------------|----------------|---------|
| 10    | 92.34 us       | 25.28 us       | 3.7x    |
| 100   | 792.69 us      | 111.99 us      | 7.1x    |
| 1000  | 36.47 us       | 20.89 us       | 1.7x    |
| 10000 | 170.12 us      | 30.74 us       | 5.5x    |

```mermaid
xychart-beta
    title "Read Performance (Stream Key Range)"
    x-axis [10, 100, 1000, 10000]
    y-axis "Time (us, mean)" 0 --> 800
    line [92.34, 792.69, 36.47, 170.12]
    line [25.28, 111.99, 20.89, 30.74]
```
</details>

<details>
  <summary><strong><em>Insert (Sequential)</em></strong></summary>

_N = number of entries inserted to empty tree_

| Size  | 🔵 Prolly Tree | 🟢 Search Tree | Speedup |
|-------|----------------|----------------|---------|
| 10    | 0.04 ms        | 0.02 ms        | 2.3x    |
| 100   | 2.55 ms        | 0.73 ms        | 3.5x    |
| 1000  | 77.92 ms       | 19.01 ms       | 4.1x    |
| 10000 | 1143.91 ms     | 280.95 ms      | 4.1x    |

```mermaid
xychart-beta
    title "Insert Performance (Sequential Keys)"
    x-axis [10, 100, 1000, 10000]
    y-axis "Time (ms, mean)" 0 --> 1200
    line [0.04, 2.55, 77.92, 1143.91]
    line [0.02, 0.73, 19.01, 280.95]
```
</details>

<details>
  <summary><strong><em>Insert (Random)</em></strong></summary>

_N = number of entries inserted to empty tree_

| Size  | 🔵 Prolly Tree | 🟢 Search Tree | Speedup |
|-------|----------------|----------------|---------|
| 10    | 0.04 ms        | 0.02 ms        | 2.3x    |
| 100   | 2.34 ms        | 0.65 ms        | 3.6x    |
| 1000  | 88.95 ms       | 20.48 ms       | 4.3x    |
| 10000 | 1513.65 ms     | 335.78 ms      | 4.5x    |

```mermaid
xychart-beta
    title "Insert Performance (Random Keys)"
    x-axis [10, 100, 1000, 10000]
    y-axis "Time (ms, mean)" 0 --> 1600
    line [0.04, 2.34, 88.95, 1513.65]
    line [0.02, 0.65, 20.48, 335.78]
```
</details>

<details>
  <summary><strong><em>Delete</em></strong></summary>

_N = size of tree_

| Size  | 🔵 Prolly Tree | 🟢 Search Tree | Speedup |
|-------|----------------|----------------|---------|
| 10    | 0.07 ms        | 0.02 ms        | 2.9x    |
| 100   | 3.74 ms        | 0.88 ms        | 4.2x    |
| 1000  | 127.19 ms      | 25.09 ms       | 5.1x    |
| 10000 | 2191.91 ms     | 405.95 ms      | 5.4x    |

```mermaid
xychart-beta
    title "Delete Performance (Half the Keys)"
    x-axis [10, 100, 1000, 10000]
    y-axis "Time (ms, mean)" 0 --> 2200
    line [0.07, 3.74, 127.19, 2191.91]
    line [0.02, 0.88, 25.09, 405.95]
```
</details>

Data collected on a Framework 16 laptop:

- Ryzen 9 7940HS + Radeon RX 7700S, 64GB DDR5-5600 RAM
- rustc 1.91.1, x86_64-unknown-linux-gnu
- Linux kernel v6.12.59

### DIFF
--- a/rust/dialog-common/src/hash.rs
+++ b/rust/dialog-common/src/hash.rs
@@ -1,5 +1,7 @@
-use serde::{Deserialize, Serialize};
+use base58::ToBase58;
+use rkyv::Archive;
 use thiserror::Error;
+use zerocopy_derive::{FromBytes, Immutable, KnownLayout};
 
 /// The size of a BLAKE3 hash in bytes.
 ///
@@ -20,9 +22,33 @@ pub const BLAKE3_HASH_SIZE: usize = 32;
 /// let data = b"hello world";
 /// let hash = Blake3Hash::hash(data);
 /// ```
-#[derive(Clone, Default, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(
+    Clone,
+    Debug,
+    Hash,
+    Default,
+    Archive,
+    rkyv::Serialize,
+    rkyv::Deserialize,
+    serde::Serialize,
+    serde::Deserialize,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    FromBytes,
+    Immutable,
+    KnownLayout,
+)]
 #[repr(transparent)]
 pub struct Blake3Hash([u8; 32]);
+
+impl ArchivedBlake3Hash {
+    /// Returns the hash bytes as a 32-byte array reference.
+    pub fn bytes(&self) -> &[u8; BLAKE3_HASH_SIZE] {
+        &self.0
+    }
+}
 
 impl Blake3Hash {
     /// Computes the BLAKE3 hash of the given bytes.
@@ -48,8 +74,25 @@ impl Blake3Hash {
     }
 
     /// Returns the hash as a byte slice.
-    pub fn as_bytes(&self) -> &[u8; 32] {
+    pub fn as_bytes(&self) -> &[u8; BLAKE3_HASH_SIZE] {
         &self.0
+    }
+
+    /// Computes the BLAKE3 hash of multiple byte slices.
+    pub fn hash_iter<'a, I>(bytes: I) -> Self
+    where
+        I: Iterator<Item = &'a [u8]>,
+    {
+        let mut hasher = blake3::Hasher::new();
+        for chunk in bytes {
+            hasher.update(chunk);
+        }
+        Self(hasher.finalize().into())
+    }
+
+    /// Checks if the hash of the given bytes matches this hash.
+    pub fn matches(&self, bytes: &[u8]) -> bool {
+        Self::hash(bytes) == *self
     }
 }
 
@@ -77,6 +120,18 @@ impl TryFrom<Vec<u8>> for Blake3Hash {
     }
 }
 
+impl std::fmt::Display for Blake3Hash {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "blake3#{}", self.0.to_base58())
+    }
+}
+
+impl<'a> From<&'a ArchivedBlake3Hash> for &'a Blake3Hash {
+    fn from(value: &'a ArchivedBlake3Hash) -> Self {
+        zerocopy::transmute_ref!(value.bytes())
+    }
+}
+
 /// Error returned when trying to convert byte arry to a Blake3Hash.
 #[derive(Error, Debug, Clone)]
 pub enum ConversionError {
@@ -84,3 +139,8 @@ pub enum ConversionError {
     #[error("Expected {BLAKE3_HASH_SIZE} bytes, got {0}")]
     InvalidSize(usize),
 }
+
+/// A null hash consisting of all zero bytes.
+///
+/// This is used as a sentinel value to represent an empty or uninitialized state.
+pub const NULL_BLAKE3_HASH: &Blake3Hash = &Blake3Hash([0u8; 32]);
--- a/rust/dialog-search-tree/src/body.rs
+++ b/rust/dialog-search-tree/src/body.rs
@@ -1,0 +1,171 @@
+use std::cmp::Ordering;
+
+use crate::{ArchivedEntry, DialogSearchTreeError, Entry, Key, Link, SymmetryWith, Value};
+use rkyv::{
+    Archive, Deserialize, Serialize,
+    bytecheck::CheckBytes,
+    de::Pool,
+    rancor::Strategy,
+    ser::{Serializer, allocator::ArenaHandle, sharing::Share},
+    util::AlignedVec,
+    validation::{Validator, archive::ArchiveValidator, shared::SharedValidator},
+};
+
+/// An index node containing links to child nodes.
+#[derive(Debug, Clone, Archive, Serialize, Deserialize)]
+pub struct Index<Key> {
+    /// The child node links stored in this index.
+    pub links: Vec<Link<Key>>,
+}
+
+impl<Key> Index<Key>
+where
+    Key: self::Key,
+    Key::Archived: PartialOrd<Key> + PartialEq<Key> + SymmetryWith<Key> + Ord,
+{
+    /// Creates a new [`Index`] containing a single link.
+    pub fn new(link: Link<Key>) -> Self {
+        Self { links: vec![link] }
+    }
+}
+
+impl<Key> ArchivedIndex<Key>
+where
+    Key: self::Key,
+    Key::Archived: PartialOrd<Key> + PartialEq<Key> + SymmetryWith<Key> + Ord,
+{
+    /// Returns the upper bound key of the last link in this index.
+    pub fn upper_bound(&self) -> Option<&Key::Archived> {
+        self.links.last().map(|link| &link.upper_bound)
+    }
+}
+
+/// A leaf segment containing key-value entries.
+#[derive(Debug, Clone, Archive, Serialize, Deserialize)]
+pub struct Segment<Key, Value> {
+    /// The key-value entries stored in this segment.
+    pub entries: Vec<Entry<Key, Value>>,
+}
+
+impl<Key, Value> Segment<Key, Value>
+where
+    Key: self::Key,
+    Key::Archived: PartialOrd<Key> + PartialEq<Key> + SymmetryWith<Key> + Ord,
+    Value: self::Value,
+{
+    /// Creates a new [`Segment`] containing a single entry.
+    pub fn new(entry: Entry<Key, Value>) -> Self {
+        Self {
+            entries: vec![entry],
+        }
+    }
+}
+
+impl<Key, Value> ArchivedSegment<Key, Value>
+where
+    Key: self::Key,
+    Key::Archived: PartialOrd<Key> + PartialEq<Key> + SymmetryWith<Key> + Ord,
+    Value: self::Value,
+{
+    /// Returns the key of the last entry in this segment.
+    pub fn upper_bound(&self) -> Option<&Key::Archived> {
+        self.entries.last().map(|entry| &entry.key)
+    }
+}
+
+/// The body of a tree node, either an index or a leaf segment.
+#[derive(Debug, Clone, Archive, Serialize, Deserialize)]
+pub enum NodeBody<Key, Value> {
+    /// An index node containing links to child nodes.
+    Index(Index<Key>),
+    /// A leaf segment containing key-value entries.
+    Segment(Segment<Key, Value>),
+}
+
+impl<Key, Value> NodeBody<Key, Value>
+where
+    Key: self::Key
+        + for<'a> Serialize<
+            Strategy<Serializer<AlignedVec, ArenaHandle<'a>, Share>, rkyv::rancor::Error>,
+        >,
+    Key::Archived: PartialOrd<Key> + PartialEq<Key> + SymmetryWith<Key> + Ord,
+    Value: self::Value
+        + for<'a> Serialize<
+            Strategy<Serializer<AlignedVec, ArenaHandle<'a>, Share>, rkyv::rancor::Error>,
+        >,
+{
+    /// Serializes this node body to bytes.
+    pub fn as_bytes(&self) -> Result<Vec<u8>, DialogSearchTreeError> {
+        rkyv::to_bytes(self)
+            .map_err(|error| DialogSearchTreeError::Encoding(format!("{error}")))
+            .map(|bytes| bytes.to_vec())
+    }
+}
+
+impl<Key, Value> TryFrom<Vec<Link<Key>>> for NodeBody<Key, Value> {
+    type Error = DialogSearchTreeError;
+
+    fn try_from(links: Vec<Link<Key>>) -> Result<Self, Self::Error> {
+        if links.is_empty() {
+            return Err(DialogSearchTreeError::Node(
+                "Attempted to create an index from zero links".into(),
+            ));
+        }
+        Ok(NodeBody::Index(Index { links }))
+    }
+}
+
+impl<Key, Value> TryFrom<Vec<Entry<Key, Value>>> for NodeBody<Key, Value> {
+    type Error = DialogSearchTreeError;
+
+    fn try_from(entries: Vec<Entry<Key, Value>>) -> Result<Self, Self::Error> {
+        if entries.is_empty() {
+            return Err(DialogSearchTreeError::Node(
+                "Attempted to create an index from zero links".into(),
+            ));
+        }
+        Ok(NodeBody::Segment(Segment { entries }))
+    }
+}
+
+impl<Key, Value> ArchivedNodeBody<Key, Value>
+where
+    Key: self::Key,
+    Key::Archived: for<'a> CheckBytes<
+            Strategy<Validator<ArchiveValidator<'a>, SharedValidator>, rkyv::rancor::Error>,
+        > + Deserialize<Key, Strategy<Pool, rkyv::rancor::Error>>
+        + PartialOrd<Key>
+        + PartialEq<Key>
+        + SymmetryWith<Key>
+        + Ord,
+    Value: self::Value,
+{
+    /// Returns the upper bound key of this node body.
+    pub fn upper_bound(&self) -> Result<&Key::Archived, DialogSearchTreeError> {
+        match self {
+            ArchivedNodeBody::Index(index) => index.upper_bound(),
+            ArchivedNodeBody::Segment(segment) => segment.upper_bound(),
+        }
+        .ok_or_else(|| DialogSearchTreeError::Node("Node was unexpectedly empty".into()))
+    }
+
+    /// Searches for an entry with the given key in this segment.
+    ///
+    /// Returns `Ok(None)` if this is an index node or if the key is not found.
+    pub fn find_entry(
+        &self,
+        key: &Key,
+    ) -> Result<Option<&ArchivedEntry<Key, Value>>, DialogSearchTreeError> {
+        match self {
+            ArchivedNodeBody::Index(_) => Err(DialogSearchTreeError::Access(
+                "Attempted to find an entry in an index node".into(),
+            )),
+            ArchivedNodeBody::Segment(segment) => Ok(segment
+                .entries
+                .binary_search_by(|entry| entry.key.partial_cmp(key).unwrap_or(Ordering::Less))
+                .map(|index| segment.entries.get(index))
+                .ok()
+                .and_then(|entry| entry)),
+        }
+    }
+}
--- a/rust/dialog-search-tree/src/buffer.rs
+++ b/rust/dialog-search-tree/src/buffer.rs
@@ -1,0 +1,30 @@
+use std::sync::{Arc, OnceLock};
+
+use dialog_common::Blake3Hash;
+
+/// A reference-counted buffer with lazy hash computation.
+///
+/// Buffers store serialized node data with efficient cloning
+/// and on-demand hash calculation.
+#[derive(Clone, Debug)]
+pub struct Buffer(Arc<(Vec<u8>, OnceLock<Blake3Hash>)>);
+
+impl Buffer {
+    /// Returns the [`Blake3Hash`] of this buffer's contents, computing it if
+    /// necessary.
+    pub fn blake3_hash(&self) -> &Blake3Hash {
+        self.0.1.get_or_init(|| Blake3Hash::hash(&self.0.0))
+    }
+}
+
+impl AsRef<[u8]> for Buffer {
+    fn as_ref(&self) -> &[u8] {
+        self.0.0.as_ref()
+    }
+}
+
+impl From<Vec<u8>> for Buffer {
+    fn from(value: Vec<u8>) -> Self {
+        Self(Arc::new((value, OnceLock::new())))
+    }
+}
--- a/rust/dialog-search-tree/src/cache.rs
+++ b/rust/dialog-search-tree/src/cache.rs
@@ -1,0 +1,111 @@
+use dialog_common::{ConditionalSend, ConditionalSync};
+
+#[cfg(not(target_arch = "wasm32"))]
+use sieve_cache::ShardedSieveCache as SieveCache;
+#[cfg(target_arch = "wasm32")]
+use sieve_cache::SieveCache;
+
+use std::hash::Hash;
+#[cfg(target_arch = "wasm32")]
+use std::{cell::RefCell, rc::Rc};
+
+const CACHE_CAPACITY: usize = 2048;
+
+/// A thread-safe cache for storing frequently accessed values.
+#[derive(Clone)]
+pub struct Cache<K, V>
+where
+    K: Eq + Hash + Clone + ConditionalSend + ConditionalSync,
+    V: Clone + ConditionalSend + ConditionalSync,
+{
+    #[cfg(not(target_arch = "wasm32"))]
+    cache: SieveCache<K, V>,
+
+    #[cfg(target_arch = "wasm32")]
+    cache: Rc<RefCell<SieveCache<K, V>>>,
+}
+
+impl<K, V> std::fmt::Debug for Cache<K, V>
+where
+    K: Eq + Hash + Clone + ConditionalSend + ConditionalSync,
+    V: Clone + ConditionalSend + ConditionalSync,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut debug_struct = f.debug_struct("Cache");
+        #[cfg(not(target_arch = "wasm32"))]
+        let debug_struct = debug_struct.field("cache", &self.cache.len());
+
+        #[cfg(target_arch = "wasm32")]
+        let debug_struct = debug_struct.field("cache", &self.cache.borrow().len());
+
+        debug_struct.finish()
+    }
+}
+
+impl<K, V> Default for Cache<K, V>
+where
+    K: Eq + Hash + Clone + ConditionalSend + ConditionalSync,
+    V: Clone + ConditionalSend + ConditionalSync,
+{
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<K, V> Cache<K, V>
+where
+    K: Eq + Hash + Clone + ConditionalSend + ConditionalSync,
+    V: Clone + ConditionalSend + ConditionalSync,
+{
+    /// Creates a new cache with a fixed capacity.
+    pub fn new() -> Self {
+        // SAFETY: `SieveCache` only returns an error if the cache capacity is 0.
+        let cache = SieveCache::new(CACHE_CAPACITY).unwrap();
+
+        Self {
+            #[cfg(not(target_arch = "wasm32"))]
+            cache,
+            #[cfg(target_arch = "wasm32")]
+            cache: Rc::new(RefCell::new(cache)),
+        }
+    }
+
+    /// Retrieves a value from the cache, or fetches it using the provided
+    /// function.
+    pub async fn get_or_fetch<F, E>(&self, key: &K, fetcher: F) -> Result<Option<V>, E>
+    where
+        F: AsyncFnOnce(&K) -> Result<Option<V>, E>,
+    {
+        #[cfg(not(target_arch = "wasm32"))]
+        if let Some(value) = self.cache.get(key) {
+            return Ok(Some(value));
+        }
+
+        #[cfg(target_arch = "wasm32")]
+        if let Some(value) = self.cache.borrow_mut().get(key) {
+            let value = value.clone();
+            return Ok(Some(value));
+        }
+
+        Ok(if let Some(value) = fetcher(key).await? {
+            #[cfg(not(target_arch = "wasm32"))]
+            self.cache.insert(key.clone(), value.clone());
+            #[cfg(target_arch = "wasm32")]
+            self.cache.borrow_mut().insert(key.clone(), value.clone());
+
+            Some(value)
+        } else {
+            None
+        })
+    }
+
+    /// Inserts a key-value pair into the cache.
+    pub fn insert(&self, key: K, value: V) -> bool {
+        #[cfg(not(target_arch = "wasm32"))]
+        let cache = &self.cache;
+        #[cfg(target_arch = "wasm32")]
+        let mut cache = self.cache.borrow_mut();
+
+        cache.insert(key, value)
+    }
+}
--- a/rust/dialog-search-tree/src/delta.rs
+++ b/rust/dialog-search-tree/src/delta.rs
@@ -1,0 +1,403 @@
+use parking_lot::Mutex;
+use std::sync::Arc;
+
+use hashbrown::HashMap;
+
+/// A thread-safe accumulator of pending changes to tree nodes.
+#[derive(Clone, Debug)]
+pub struct Delta<K, V> {
+    contents: Arc<Mutex<HashMap<K, V>>>,
+}
+
+impl<K, V> Delta<K, V>
+where
+    K: Clone + std::hash::Hash + PartialEq + Eq + std::fmt::Display,
+    V: Clone,
+{
+    /// Creates an empty delta with no pending changes.
+    pub fn zero() -> Self {
+        Self {
+            contents: Default::default(),
+        }
+    }
+
+    /// Creates a new delta that contains a copy of this delta's contents.
+    pub fn branch(&self) -> Self {
+        Self {
+            contents: Arc::new(Mutex::new(self.contents.lock().clone())),
+        }
+    }
+
+    /// Adds a key-value pair to this delta.
+    pub fn add(&mut self, key: K, value: V) {
+        let mut contents = self.contents.lock();
+        contents.insert(key.clone(), value);
+    }
+
+    /// Adds multiple key-value pairs to this delta.
+    pub fn add_all<Entries>(&mut self, entries: Entries)
+    where
+        Entries: Iterator<Item = (K, V)>,
+    {
+        let mut contents = self.contents.lock();
+        for (key, value) in entries {
+            contents.insert(key.clone(), value);
+        }
+    }
+
+    /// Removes a key from this delta.
+    pub fn subtract(&mut self, key: &K) {
+        let mut contents = self.contents.lock();
+        contents.remove(key);
+    }
+
+    /// Removes multiple keys from this delta.
+    pub fn subtract_all<'a, Keys>(&'a mut self, keys: Keys)
+    where
+        Keys: Iterator<Item = &'a K>,
+    {
+        let mut contents = self.contents.lock();
+        for key in keys {
+            contents.remove(key);
+        }
+    }
+
+    /// Retrieves the value associated with a key from this delta.
+    pub fn get(&self, key: &K) -> Option<V> {
+        let contents = self.contents.lock();
+        contents.get(key).cloned()
+    }
+
+    /// Drains all entries from this delta and returns them as an iterator.
+    pub fn flush(&mut self) -> impl Iterator<Item = (K, V)> {
+        std::mem::take(&mut *self.contents.lock()).into_iter()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #![allow(unexpected_cfgs)]
+
+    use super::*;
+    use anyhow::Result;
+
+    #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+    wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_dedicated_worker);
+
+    #[dialog_common::test]
+    async fn it_creates_an_empty_delta() -> Result<()> {
+        let delta = Delta::<u32, String>::zero();
+        assert_eq!(delta.get(&1), None);
+        assert_eq!(delta.get(&100), None);
+
+        Ok(())
+    }
+
+    #[dialog_common::test]
+    async fn it_adds_a_single_entry() -> Result<()> {
+        let mut delta = Delta::<u32, String>::zero();
+
+        delta.add(1, "value1".to_string());
+        assert_eq!(delta.get(&1), Some("value1".to_string()));
+
+        Ok(())
+    }
+
+    #[dialog_common::test]
+    async fn it_adds_multiple_entries() -> Result<()> {
+        let mut delta = Delta::<u32, String>::zero();
+
+        delta.add(1, "value1".to_string());
+        delta.add(2, "value2".to_string());
+        delta.add(3, "value3".to_string());
+
+        assert_eq!(delta.get(&1), Some("value1".to_string()));
+        assert_eq!(delta.get(&2), Some("value2".to_string()));
+        assert_eq!(delta.get(&3), Some("value3".to_string()));
+        assert_eq!(delta.get(&4), None);
+
+        Ok(())
+    }
+
+    #[dialog_common::test]
+    async fn it_overwrites_existing_entries() -> Result<()> {
+        let mut delta = Delta::<u32, String>::zero();
+
+        delta.add(1, "value1".to_string());
+        assert_eq!(delta.get(&1), Some("value1".to_string()));
+
+        delta.add(1, "updated".to_string());
+        assert_eq!(delta.get(&1), Some("updated".to_string()));
+
+        Ok(())
+    }
+
+    #[dialog_common::test]
+    async fn it_adds_all_entries_from_an_iterator() -> Result<()> {
+        let mut delta = Delta::<u32, String>::zero();
+
+        let entries = vec![
+            (1, "value1".to_string()),
+            (2, "value2".to_string()),
+            (3, "value3".to_string()),
+        ];
+
+        delta.add_all(entries.into_iter());
+
+        assert_eq!(delta.get(&1), Some("value1".to_string()));
+        assert_eq!(delta.get(&2), Some("value2".to_string()));
+        assert_eq!(delta.get(&3), Some("value3".to_string()));
+
+        Ok(())
+    }
+
+    #[dialog_common::test]
+    async fn it_subtracts_an_entry() -> Result<()> {
+        let mut delta = Delta::<u32, String>::zero();
+
+        delta.add(1, "value1".to_string());
+        delta.add(2, "value2".to_string());
+
+        assert_eq!(delta.get(&1), Some("value1".to_string()));
+        assert_eq!(delta.get(&2), Some("value2".to_string()));
+
+        delta.subtract(&1);
+
+        assert_eq!(delta.get(&1), None);
+        assert_eq!(delta.get(&2), Some("value2".to_string()));
+
+        Ok(())
+    }
+
+    #[dialog_common::test]
+    async fn it_subtracts_nonexistent_entry_without_error() -> Result<()> {
+        let mut delta = Delta::<u32, String>::zero();
+
+        delta.add(1, "value1".to_string());
+
+        // Subtracting non-existent key should not panic
+        delta.subtract(&999);
+
+        assert_eq!(delta.get(&1), Some("value1".to_string()));
+
+        Ok(())
+    }
+
+    #[dialog_common::test]
+    async fn it_subtracts_all_entries_from_an_iterator() -> Result<()> {
+        let mut delta = Delta::<u32, String>::zero();
+
+        delta.add(1, "value1".to_string());
+        delta.add(2, "value2".to_string());
+        delta.add(3, "value3".to_string());
+        delta.add(4, "value4".to_string());
+
+        let keys_to_remove = [1, 3];
+        delta.subtract_all(keys_to_remove.iter());
+
+        assert_eq!(delta.get(&1), None);
+        assert_eq!(delta.get(&2), Some("value2".to_string()));
+        assert_eq!(delta.get(&3), None);
+        assert_eq!(delta.get(&4), Some("value4".to_string()));
+
+        Ok(())
+    }
+
+    #[dialog_common::test]
+    async fn it_branches_to_create_independent_copy() -> Result<()> {
+        let mut delta1 = Delta::<u32, String>::zero();
+
+        delta1.add(1, "value1".to_string());
+        delta1.add(2, "value2".to_string());
+
+        let mut delta2 = delta1.branch();
+
+        // Verify delta2 has the same contents initially
+        assert_eq!(delta2.get(&1), Some("value1".to_string()));
+        assert_eq!(delta2.get(&2), Some("value2".to_string()));
+
+        // Modify delta2
+        delta2.add(3, "value3".to_string());
+        delta2.subtract(&1);
+
+        // Verify delta1 is unchanged
+        assert_eq!(delta1.get(&1), Some("value1".to_string()));
+        assert_eq!(delta1.get(&2), Some("value2".to_string()));
+        assert_eq!(delta1.get(&3), None);
+
+        // Verify delta2 has the changes
+        assert_eq!(delta2.get(&1), None);
+        assert_eq!(delta2.get(&2), Some("value2".to_string()));
+        assert_eq!(delta2.get(&3), Some("value3".to_string()));
+
+        Ok(())
+    }
+
+    #[dialog_common::test]
+    async fn it_flushes_all_entries() -> Result<()> {
+        let mut delta = Delta::<u32, String>::zero();
+
+        delta.add(1, "value1".to_string());
+        delta.add(2, "value2".to_string());
+        delta.add(3, "value3".to_string());
+
+        let entries: Vec<(u32, String)> = delta.flush().collect();
+
+        // Verify all entries were flushed
+        assert_eq!(entries.len(), 3);
+        assert!(entries.contains(&(1, "value1".to_string())));
+        assert!(entries.contains(&(2, "value2".to_string())));
+        assert!(entries.contains(&(3, "value3".to_string())));
+
+        // Verify delta is now empty
+        assert_eq!(delta.get(&1), None);
+        assert_eq!(delta.get(&2), None);
+        assert_eq!(delta.get(&3), None);
+
+        Ok(())
+    }
+
+    #[dialog_common::test]
+    async fn it_flushes_an_empty_delta() -> Result<()> {
+        let mut delta = Delta::<u32, String>::zero();
+
+        let entries: Vec<(u32, String)> = delta.flush().collect();
+
+        assert_eq!(entries.len(), 0);
+
+        Ok(())
+    }
+
+    #[dialog_common::test]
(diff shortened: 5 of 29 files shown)


rust/dialog-common/src/hash.rs
```diff
@@ -17,10 +23,32 @@ pub const BLAKE3_HASH_SIZE: usize = 32;
...
+    PartialOrd,
+    Ord,
+    FromBytes,
+    Immutable,
+    KnownLayout,
+)]
 #[repr(transparent)]
 pub struct Blake3Hash([u8; 32]);
 
+impl ArchivedBlake3Hash {
```
Gozala, line 46: Is this a struct generated by `derive(Archive)` or something ?

rust/dialog-common/src/sync.rs
```diff
@@ -8,28 +8,36 @@
...
 
 #[cfg(not(target_arch = "wasm32"))]
 impl<S> ConditionalSend for S where S: Send {}
 
-#[allow(missing_docs)]
+/// A marker trait that is `Send + Sync` on native targets and no-op on Wasm.
+///
+/// Use this to conditionally require `Send + Sync` bounds in cross-platform code.
 #[cfg(not(target_arch = "wasm32"))]
 pub trait ConditionalSync: Send + Sync {}
```
Gozala, line 20: Unrelated to changes here but I always wondered why this has `Send +` in it.

Gozala: In other words I would have dropped `Send +` or called this `ConditionalSendSync`

rust/dialog-search-tree/src/tree.rs
```diff
@@ -0,0 +1,1161 @@
...
+    util::AlignedVec,
+    validation::{Validator, archive::ArchiveValidator, shared::SharedValidator},
+};
+
+use crate::{
+    Buffer, Cache, ContentAddressedStorage, Delta, DialogSearchTreeError, Entry, Key, Node,
+    SearchResult, SymmetryWith, TreeShaper, TreeWalker, Value, into_owned,
+};
+
+/// A key-value store backed by a ranked prolly tree with content-addressed
```
Gozala, line 21:
```suggestion
/// A key-value store backed by a Geometric Search Tree with content-addressed
```

I think this is more accurate

rust/dialog-search-tree/src/tree.rs
```diff
@@ -0,0 +1,1161 @@
...
+    Key::Archived: PartialOrd<Key> + PartialEq<Key> + SymmetryWith<Key> + Ord,
+    Value: self::Value,
+{
+    key: PhantomData<Key>,
+    value: PhantomData<Value>,
+
+    root: Blake3Hash,
+    node_cache: Cache<Blake3Hash, Buffer>,
+
+    delta: Delta<Blake3Hash, Buffer>,
```
Gozala, line 48: I think comment explaining what this holds would good here

rust/dialog-search-tree/src/tree.rs
```diff
@@ -0,0 +1,1161 @@
...
+    Key: self::Key,
+    Key: PartialOrd<Key::Archived> + PartialEq<Key::Archived>,
+    Key::Archived: PartialOrd<Key> + PartialEq<Key> + SymmetryWith<Key> + Ord,
+    Key::Archived: for<'a> CheckBytes<
+        Strategy<Validator<ArchiveValidator<'a>, SharedValidator>, rkyv::rancor::Error>,
+    >,
+    Key::Archived: Deserialize<Key, Strategy<Pool, rkyv::rancor::Error>>,
+    Key: for<'a> Serialize<
+        Strategy<Serializer<AlignedVec, ArenaHandle<'a>, Share>, rkyv::rancor::Error>,
+    >,
```
Gozala, line 62: 😱 Would it be possible to have something like `trait Key: ....`  that captures all this  maybe even with some explanation so we could write something like `impl Key<Key: crate::Key, value: crate::Value>`, maybe what I'm suggesting is not an idiomatic approach ? I just find giving invariants some name to refer by helps a lot.

rust/dialog-search-tree/src/tree.rs
```diff
@@ -0,0 +1,1161 @@
...
+    /// Inserts a `key`/`value` pair into the tree, returning a new tree
+    /// version.
+    ///
+    /// If the key already exists, its value is updated. If the key is new, it
+    /// is inserted in sorted order within the appropriate leaf segment.
+    pub async fn insert<Backend>(
+        &self,
+        key: Key,
+        value: Value,
+        storage: &ContentAddressedStorage<Backend>,
```
Gozala, line 145: Should not this be `&mut ContentAddressedStorage<Backend>` or is this an idiomatic way to signal that interior mutability takes place here ?

rust/dialog-search-tree/src/tree.rs
```diff
@@ -0,0 +1,1161 @@
...
+
+    /// Retrieves the value associated with `key` from the tree.
+    ///
+    /// This method performs a binary search through the tree hierarchy to
+    /// locate the leaf segment containing the key, then searches within that
+    /// segment for the specific entry.
+    ///
+    /// Returns `Ok(Some(value))` if the key exists, `Ok(None)` if the key is
+    /// not found, or an error if the tree structure is invalid or storage
+    /// access fails.
```
Gozala, line 116: I have not looked for answer yet, but I find myself wondering if this tries to load things from `node_cache` falling back to `storage` on cache miss and whether things loaded from store get cached into `node_cache` afterwards. Perhaps doc comment here should be speaking to that.

rust/dialog-search-tree/src/tree.rs
```diff
@@ -0,0 +1,1161 @@
...
+    Key: self::Key,
+    Key: PartialOrd<Key::Archived> + PartialEq<Key::Archived>,
+    Key::Archived: PartialOrd<Key> + PartialEq<Key> + SymmetryWith<Key> + Ord,
+    Value: self::Value,
+{
+    key: PhantomData<Key>,
+    value: PhantomData<Value>,
+
+    root: Blake3Hash,
+    node_cache: Cache<Blake3Hash, Buffer>,
```
Gozala, line 46: I find this bit controversial (without exploring much of code yet) as in why does it cache nodes as opposed making it concern of the store itself. Furthermore I'm wondering what is considered "frequently accessed".

rust/dialog-search-tree/src/tree.rs
```diff
@@ -0,0 +1,1161 @@
...
+
+    /// Returns an async stream over all entries in the tree.
+    ///
+    /// Entries are yielded in sorted order by key. This method traverses the
+    /// tree from the leftmost leaf to the rightmost, streaming entries as they
+    /// are encountered without loading the entire tree into memory.
+    ///
+    /// Internally, this calls [`stream_range`](Self::stream_range) with an
+    /// unbounded range covering all possible keys.
+    pub fn stream<'a, Backend>(
```
Gozala, line 187: I think it should be called entries to describe what it returns as opposed what representation it returns it.

rust/dialog-search-tree/src/tree.rs
```diff
@@ -0,0 +1,1161 @@
...
+            storage,
+        )
+    }
+
+    /// Returns an async stream over entries with keys within the provided
+    /// range.
+    ///
+    /// The range can be bounded or unbounded on either end, following Rust's
+    /// standard [`RangeBounds`] trait. Entries are yielded in sorted order.
+    pub fn stream_range<'a, R, Backend>(
```
Gozala, line 205: I would call it `select` or `entries_in_range`. Separately maybe just one method that takes optional range is good enough ?

rust/dialog-search-tree/src/tree.rs
```diff
@@ -0,0 +1,1161 @@
...
+            self.get_node(hash, storage).await
+        })
+        .stream(range)
+    }
+
+    /// Flushes all pending changes, returning an iterator of nodes to be
+    /// written to storage.
+    ///
+    /// This method drains the accumulated delta and returns hash-buffer pairs
+    /// representing the new or modified nodes that need to be persisted to
```
Gozala, line 224: Does modified nodes here imply nodes that are no longer part of the tree ? If so how does one tell if node is part of this tree or not ?

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+
+/// A layer in the tree traversal path, containing a node and its sibling links.
+pub struct TreeLayer<Key, Value>
+where
+    Key: self::Key,
+    Key::Archived: PartialOrd<Key> + PartialEq<Key> + SymmetryWith<Key> + Ord,
+{
+    // pub host: Blake3Hash,
+    /// The node at this layer of the tree.
+    pub host: Node<Key, Value>,
```
Gozala, line 219: Is this a target node into which we're traversing, meaning it's middle between left and right siblings ?

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+                                break;
+                            }
+                        }
+                    },
+                }
+            }
+        }
+    }
+
+    /// Searches for the leaf segment that would contain the given key.
```
Gozala, line 133:
```suggestion
    /// Searches for the segment node containing entry for the given key.
```

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+                }
+                ArchivedNodeBody::Segment(_) => {
+                    return Ok(Some(SearchResult { leaf: node, path }));
+                }
+            }
+        }
+    }
+}
+
+/// A layer in the tree traversal path, containing a node and its sibling links.
```
Gozala, line 211:
```suggestion
/// A horizontal slice of the tree at a specific depth, capturing the node
/// along the search path and its immediate siblings at that level.
```

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+pub struct TreeLayer<Key, Value>
+where
+    Key: self::Key,
+    Key::Archived: PartialOrd<Key> + PartialEq<Key> + SymmetryWith<Key> + Ord,
+{
+    // pub host: Blake3Hash,
+    /// The node at this layer of the tree.
```
Gozala, line 218:
```suggestion
    /// The node on the search path at this level.
```

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+    pub host: Node<Key, Value>,
+    /// Links to sibling nodes to the left of the current path.
```
Gozala, line 220:
```suggestion
    /// Links to sibling nodes to the left of the search path at this level.
```

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+    pub left_siblings: Option<NonEmpty<Link<Key>>>,
+    /// Links to sibling nodes to the right of the current path.
```
Gozala, line 222:
```suggestion
    /// Links to sibling nodes to the right of the search path at this level.
```

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+    pub right_siblings: Option<NonEmpty<Link<Key>>>,
```
Gozala, line 223: I'm bit confused is this representation of the split node as in as in host, left and right siblings are all children of the node or left and right siblings are spans across multiple parent nodes ? If later I think name is misleading because nodes that have different parent aren't really siblings, if former `TreeLayer` is confusing as because it's more of cursor / view over node node.

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+                return Err(DialogSearchTreeError::Operation(format!(
+                    "Tree depth exceded the soft maximum ({MAXIMUM_TREE_DEPTH})"
+                )));
+            }
+
+            let node = (self.get_node)(&next_node).await?;
+
+            match node.body()? {
+                ArchivedNodeBody::Index(index) => {
+                    let mut left = vec![];
```
Gozala, line 160:
```suggestion
                ArchivedNodeBody::Index(index) => {
                    // Partition sibling links as we scan for the descent path.
                    //
                    // We iterate through child links, each tagged with an upper_bound.
                    // The first link whose upper_bound >= key is our descent target.
                    // Links before it go to `left`, links after go to `right`.
                    let mut left = vec![];
```

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+                    let mut right = vec![];
+                    let mut next_descendant = None;
+
+                    for link in index.links.iter() {
+                        if next_descendant.is_some() {
+                            right.push(link);
```
Gozala, line 166:
```suggestion
                           // Already found our path, remaining links are on the right
                            right.push(link);
```

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+                        } else if key <= &link.upper_bound {
+                            next_descendant = Some(&link.node);
```
Gozala, line 168:
```suggestion
                            // Found the child that covers our key
                            next_descendant = Some(&link.node);
```

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+                        } else {
+                            left.push(link);
```
Gozala, line 170:
```suggestion
                            // This child's range is entirely less than our key
                            left.push(link);
```

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+                        }
+                    }
+
+                    if next_descendant.is_none() {
```
Gozala, line 174:
```suggestion
                    // Key is higher than right most upper bound (e.g. during insertions that
                    // extend tree's range) in which case we fall back to the rightmost child
                    // (last one we put in `left`).
                    if next_descendant.is_none() {
```

rust/dialog-search-tree/src/tree.rs
```diff
@@ -0,0 +1,1161 @@
...
+    Key: for<'a> Serialize<
+        Strategy<Serializer<AlignedVec, ArenaHandle<'a>, Share>, rkyv::rancor::Error>,
+    >,
+    Value: self::Value,
+    Value::Archived: for<'a> CheckBytes<
+            Strategy<Validator<ArchiveValidator<'a>, SharedValidator>, rkyv::rancor::Error>,
+        > + Deserialize<Value, Strategy<Pool, rkyv::rancor::Error>>,
+    Value: for<'a> Serialize<
+        Strategy<Serializer<AlignedVec, ArenaHandle<'a>, Share>, rkyv::rancor::Error>,
+    >,
```
cdata, line 69: I will collapse some of the redundant `where` clauses in a follow-up PR.

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+pub struct TreeWalker<Key, Value, GetNode>
+where
+    Key: self::Key,
+    Key::Archived: PartialOrd<Key> + PartialEq<Key> + SymmetryWith<Key> + Ord,
+    Key: PartialOrd<Key::Archived> + PartialEq<Key::Archived>,
+    Value: self::Value,
+    Value::Archived: for<'a> CheckBytes<
+        Strategy<Validator<ArchiveValidator<'a>, SharedValidator>, rkyv::rancor::Error>,
+    >,
+    GetNode: AsyncFn(&Blake3Hash) -> Result<Node<Key, Value>, DialogSearchTreeError>,
```
cdata, line 33: This API will be changed in a follow-up PR; instead of receiving a closure the walker depends on a dedicated `Accessor` that curries the lookup without exposing the details.

(and yeah, I can imagine how this may be more elegantly expressed via some kind of effect system)

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+    rancor::Strategy,
+    validation::{Validator, archive::ArchiveValidator, shared::SharedValidator},
+};
+
+use crate::{
+    ArchivedNodeBody, DialogSearchTreeError, Entry, Key, Link, Node, SymmetryWith, Value,
+    into_owned,
+};
+
+/// A traversal mechanism for walking through a tree structure.
```
Gozala, line 23:
```suggestion
/// A traversal mechanism for walking through a tree structure.
///
/// The walker provides two main operations:
/// - `search`: Descend to the leaf that would contain a given key, recording
///   the path taken as a zipper structure for later modification
/// - `stream`: Iterate over entries within a key range
```

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+    pub fn new(root: Blake3Hash, get_node: GetNode) -> Self {
+        Self {
+            root,
+            get_node,
+            key: PhantomData,
+            value: PhantomData,
+        }
+    }
+
+    /// Returns a stream of entries within the specified key range.
```
Gozala, line 68:
```suggestion
    /// Returns a stream of entries within the specified key range.
    ///
    /// The stream performs an in-order traversal starting from the first key
    /// in the range, yielding entries until the range is exhausted or the
    /// end of the tree is reached.
```

I would also rename method to `select` because `stream` does not really tells what it does but rather what format of produces.

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+            let Some(search_result) = self.search(&start_key).await? else {
+                return;
+            };
+            let mut search_path = search_result.into_indexed()?;
+            let mut entered_range = false;
+
+            while let Some((node, maybe_index)) = search_path.pop() {
+                match node.body()? {
+                    ArchivedNodeBody::Index(index) => {
+                        let child_index = if let Some(index) = maybe_index {
```
Gozala, line 98:
```suggestion
                        // Determine next child to visit, which is either the one after our
                        // last visited child, or the first child if we just started descend
                        let child_index = if let Some(index) = maybe_index {
```

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+                            index + 1
+                        } else {
+                            0
+                        };
+
+                        match index.links.get(child_index) {
+                            Some(link) => {
+                                let next_node = (self.get_node)(<&Blake3Hash>::from(&link.node)).await?;
```
Gozala, line 106:
```suggestion
                                // Descend into this child, marking our position
                                let next_node = (self.get_node)(<&Blake3Hash>::from(&link.node)).await?;
```

rust/dialog-search-tree/src/walker.rs
Gozala, line 211:
```suggestion
/// A node visited during search, capturing the siblings on either side
/// of the child we descended into.
///
/// This structure enables efficient tree modifications: after changing a
/// descendant, we rebuild the index by combining the modified child link
/// with its recorded siblings, then continue propagating upward.
```

I would also rename this into `SearchNode` or `SearchFrame` or more direct `NodeZipper` referencing terms in the zipper algorithm.

rust/dialog-search-tree/src/walker.rs
Gozala, line 220:
```suggestion
    /// Child links to the left of the descent path (lower keys).
```

rust/dialog-search-tree/src/walker.rs
Gozala, line 218:
```suggestion
    /// The index node we descended through.
```

rust/dialog-search-tree/src/walker.rs
Gozala, line 222:
```suggestion
    /// Child links to the right of the descent path (higher keys).
```

rust/dialog-search-tree/src/walker.rs
```diff
@@ -0,0 +1,293 @@
...
+    // pub host: Blake3Hash,
+    /// The node at this layer of the tree.
+    pub host: Node<Key, Value>,
+    /// Links to sibling nodes to the left of the current path.
+    pub left_siblings: Option<NonEmpty<Link<Key>>>,
+    /// Links to sibling nodes to the right of the current path.
+    pub right_siblings: Option<NonEmpty<Link<Key>>>,
+}
+
+/// The path taken from the root to a leaf during a tree search.
```
Gozala, line 226:
```suggestion
/// The search path leading from the root to a leaf where search key falls into.
```

rust/dialog-search-tree/src/error.rs
```diff
@@ -0,0 +1,40 @@
+use dialog_storage::DialogStorageError;
+use thiserror::Error;
+
+/// Errors that can occur when working with search trees.
+#[derive(Error, Debug)]
+pub enum DialogSearchTreeError {
+    /// An error that occurs when working with a tree entry.
+    #[error("Problem with tree entry: {0}")]
```
Gozala, line 8: This may come clear as I review more, but the moment I'm not sure what issue could be with an entry as in is it corrupt entry ? In which case capturing key might be a good idea. If not malformed I'm not even sure what else could be

rust/dialog-search-tree/src/error.rs
```diff
@@ -0,0 +1,40 @@
...
+    Entry(String),
+
+    /// An error that occurs when accessing a node.
+    #[error("Problem accessing node: {0}")]
```
Gozala, line 12: Not found or corrupt node ? Should we capture node hash maybe here maybe even key that lead us here ?

rust/dialog-search-tree/src/error.rs
```diff
@@ -0,0 +1,40 @@
...
+    Node(String),
+
+    /// An error from the storage backend.
+    #[error("{0}")]
+    Storage(DialogStorageError),
+
+    /// An error that occurs during a tree operation.
+    #[error("Failed to operate the tree: {0}")]
+    Operation(String),
```
Gozala, line 21: Seems like all of the errors would be occurring on some tree operation, so maybe operation is a field on all of the variants ?

rust/dialog-search-tree/src/error.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    /// An error that occurs when retrieving an item from the cache.
+    #[error("Failed to retrieve item in cache: {0}")]
+    Cache(String),
+
+    /// An error that occurs when accessing part of the tree.
+    #[error("Failed to access part of the tree: {0}")]
```
Gozala, line 28: Node variant also talks about access error there is some overlap or redundancy between two ?

rust/dialog-search-tree/src/link.rs
```diff
@@ -0,0 +1,14 @@
+use dialog_common::Blake3Hash;
+use rkyv::{Archive, Deserialize, Serialize};
+
+/// A reference to a child node in an index node.
+///
+/// Links connect index nodes to their children, storing the child's upper bound
+/// key and its content hash.
+#[derive(Clone, Debug, Archive, Serialize, Deserialize)]
+pub struct Link<Key> {
```
Gozala, line 9: I've being thinking perhaps calling this `NodeRoute`, `TreeRoute` or simply `Route` might be better. Because it contains more than just hash link to the node and conceptually it is a route to subtree covering specific key range

rust/dialog-search-tree/src/node.rs
```diff
@@ -0,0 +1,126 @@
...
+/// (index).
+///
+/// Nodes are content-addressed by their [`Blake3Hash`] and store serialized
+/// data in a [`Buffer`].
+#[derive(Clone, Debug)]
+pub struct Node<Key, Value> {
+    key: PhantomData<Key>,
+    value: PhantomData<Value>,
+
+    buffer: Buffer,
```
Gozala, line 26: Nit: I would call this `content` as in node content.

rust/dialog-search-tree/src/node.rs
```diff
@@ -0,0 +1,126 @@
...
+        rkyv::access::<_, rkyv::rancor::Error>(self.buffer.as_ref())
+            .map_err(|error| DialogSearchTreeError::Access(format!("{error}")))
+    }
+
+    /// Interprets this node as an index node, returning an error if it's a
+    /// segment.
+    pub fn as_index(&self) -> Result<&ArchivedIndex<Key>, DialogSearchTreeError> {
+        self.body().and_then(|body| match body {
+            ArchivedNodeBody::Index(index) => Ok(index),
+            ArchivedNodeBody::Segment(_) => Err(DialogSearchTreeError::Access(
```
Gozala, line 93: Maybe `DialogSearchTreeError::TypeError` would have being obvious.

rust/dialog-search-tree/src/buffer.rs
```diff
@@ -0,0 +1,30 @@
+use std::sync::{Arc, OnceLock};
+
+use dialog_common::Blake3Hash;
+
+/// A reference-counted buffer with lazy hash computation.
+///
+/// Buffers store serialized node data with efficient cloning
+/// and on-demand hash calculation.
+#[derive(Clone, Debug)]
+pub struct Buffer(Arc<(Vec<u8>, OnceLock<Blake3Hash>)>);
```
Gozala, line 10: Nit: Should we call this a Blob since we've being referring to things as blob stores etc...

rust/dialog-search-tree/src/body.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+}
+
+impl<Key> ArchivedIndex<Key>
+where
+    Key: self::Key,
+    Key::Archived: PartialOrd<Key> + PartialEq<Key> + SymmetryWith<Key> + Ord,
+{
+    /// Returns the upper bound key of the last link in this index.
+    pub fn upper_bound(&self) -> Option<&Key::Archived> {
```
Gozala, line 38: Q: why is this on ArchivedIndex as opposed to say Index ? I'm not sure about two interrelate

rust/dialog-search-tree/src/cache.rs
```diff
@@ -0,0 +1,111 @@
...
+pub struct Cache<K, V>
+where
+    K: Eq + Hash + Clone + ConditionalSend + ConditionalSync,
+    V: Clone + ConditionalSend + ConditionalSync,
+{
+    #[cfg(not(target_arch = "wasm32"))]
+    cache: SieveCache<K, V>,
+
+    #[cfg(target_arch = "wasm32")]
+    cache: Rc<RefCell<SieveCache<K, V>>>,
```
Gozala, line 25: Mind adding comment why we need these wrappers on wasm it's not obvious to me

rust/dialog-search-tree/src/cache.rs
```diff
@@ -0,0 +1,111 @@
...
+use std::{cell::RefCell, rc::Rc};
+
+const CACHE_CAPACITY: usize = 2048;
+
+/// A thread-safe cache for storing frequently accessed values.
+#[derive(Clone)]
+pub struct Cache<K, V>
+where
+    K: Eq + Hash + Clone + ConditionalSend + ConditionalSync,
+    V: Clone + ConditionalSend + ConditionalSync,
```
Gozala, line 19: 😂 Looks like I'm not the only one confused by the fact that `ConditionalSync` is `Send + Sync`

rust/dialog-search-tree/src/delta.rs
```diff
@@ -0,0 +1,403 @@
...
+{
+    /// Creates an empty delta with no pending changes.
+    pub fn zero() -> Self {
+        Self {
+            contents: Default::default(),
+        }
+    }
+
+    /// Creates a new delta that contains a copy of this delta's contents.
+    pub fn branch(&self) -> Self {
```
Gozala, line 25: Nit: I'd find `fork` to be more intuitive here especially because we trees have branches, db replica has branches.

rust/dialog-search-tree/src/delta.rs
```diff
@@ -0,0 +1,403 @@
...
+        Entries: Iterator<Item = (K, V)>,
+    {
+        let mut contents = self.contents.lock();
+        for (key, value) in entries {
+            contents.insert(key.clone(), value);
+        }
+    }
+
+    /// Removes a key from this delta.
+    pub fn subtract(&mut self, key: &K) {
```
Gozala, line 49: Nit: I would find add / remove to be more natural

rust/dialog-search-tree/src/distribution.rs
```diff
@@ -0,0 +1,53 @@
...
+    use dialog_common::Blake3Hash;
+
+    use super::Rank;
+
+    /// The branch factor of the [`Tree`]s that constitute [`Artifact`] indexes
+    pub const BRANCH_FACTOR: u32 = 254;
+
+    /// Computes the rank of a node from its hash using a geometric distribution.
+    pub fn rank(hash: &Blake3Hash) -> Rank {
+        compute_geometric_rank(hash, BRANCH_FACTOR)
```
Gozala, line 15: I love the simplicity, but I found it extremely valuable to be able to adjust branching in the tests & I can't even imagine doing proper diff testing without this https://github.com/dialog-db/dialog-db/pull/117/files#diff-6d2f25a3a31ad4907f677748ba07640047427eeda0a68f1542fa6d9b11defb45R34

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -0,0 +1,517 @@
...
+            // Still have entries; create new segment and merge up the path
+            let new_segment_node = Node::<Key, Value>::new(Buffer::from(
+                NodeBody::try_from(segment.entries)?.as_bytes()?,
+            ));
+
+            let mut delta = self.delta.branch();
+            delta.subtract(search_result.leaf.hash());
+
+            Self::merge_with_path(
+                NonEmpty::singleton((new_segment_node, 1)),
```
Gozala, line 214: I'm not sure I fully understand the `merge_with_path` yet, but I'm under impression that it may be doing zip incorrectly. Specifically if removed entry was the last one boundary will shift causing new segment and one after it to join which can propagate upwards. I could be wrong though and maybe `merge_with_path` handles this, but fact that it takes node as opposed to entries is surprising to me.

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -0,0 +1,517 @@
...
+                // Empty tree; just a single entry
+                (NonEmpty::singleton(new_entry), None)
+            }
+        };
+
+        // Rank the entries and distribute
+        let next_entries = entries
+            .into_iter()
+            .map(|entry| {
+                let rank = distribution::geometric::rank(&Blake3Hash::hash(entry.key.as_ref()));
```
Gozala, line 155: Unless I'm missing something I'm under impression that you do a lot more hashing + ranking than strictly necessary. Insertion would not affect existing entries so all but last would have a base rank and we'd need to rank new entry which may be higher that base if it's a boundary.

So I think you could avoid whole lot of hashing and ranking here by just mapping all but last and new entry to base and computing those two.

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -0,0 +1,517 @@
...
+        let mut segment = into_owned::<Segment<Key, Value>>(segment)?;
+        segment.entries.remove(removal_index);
+
+        if !segment.entries.is_empty() {
+            // Still have entries; create new segment and merge up the path
+            let new_segment_node = Node::<Key, Value>::new(Buffer::from(
+                NodeBody::try_from(segment.entries)?.as_bytes()?,
+            ));
+
+            let mut delta = self.delta.branch();
```
Gozala, line 210: I might understand this later, but right now I'm confused about why do we need to create copy of the delta here. Comment explaining it might be very helpful.

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -0,0 +1,517 @@
...
+            links.extend(right_siblings);
+        }
+
+        // If no siblings remain, propagate the removal upward
+        if links.is_empty() {
+            return if path.is_empty() {
+                Ok((NULL_BLAKE3_HASH.clone(), delta))
+            } else {
+                // Continue removing up the path with the same delta
+                Self::remove_from_path_with_delta(path, delta)
```
Gozala, line 408: Is recursion here not a problem ?

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -0,0 +1,517 @@
...
+            };
+        }
+
+        // We have siblings; rebuild with them
+        let ranked_links = links
+            .into_iter()
+            .map(|link| {
```
Gozala, line 415: I think you can just hash first and last link, all the links between them will have same rank as first one or our tree was corrupt. If any link had higher rank they would have had a different parent.

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -0,0 +1,517 @@
...
+    /// 1. Accumulate children in a pending list
+    /// 2. When a child's rank > minimum_rank, create a node from pending
+    ///    children
+    /// 3. Continue until all children are processed
+    /// 4. Any remaining children form a final node with rank = minimum_rank
+    ///
+    /// This rank-based partitioning is what gives the prolly tree its
+    /// probabilistic splitting behavior, which in turn enables efficient
+    /// structural sharing and diff computation.
+    fn collect<Child>(
```
Gozala, line 452: I find collect to be very misleading name here, maybe `partition` or `group` or something along those lines would make more sense ?

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -0,0 +1,517 @@
...
+}
+
+/// Converts a collection of links into ranked links by computing each link's
+/// rank from its node hash.
+///
+/// This helper function is used when merging sibling references during tree
+/// reconstruction. The rank is computed using a geometric distribution over the
+/// node hash, ensuring consistent rank assignment for the same content.
+fn into_ranked_links<Key>(links: NonEmpty<Link<Key>>) -> NonEmpty<(Link<Key>, Rank)> {
+    links.map(|link| {
```
Gozala, line 496: I think Link (which I'd rename into `Head` as you already have `Body`)  should have `CellOnce` that lazily computes rank just ones, I think something else did that for hash.  Seems like that would reduce number of times it may end up recomputed.

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -0,0 +1,517 @@
...
+    ///
+    /// This is the shared path-reconstruction logic used by both insert (after
+    /// distributing entries) and delete (after modifying a segment).
+    pub fn merge_with_path(
+        mut nodes: NonEmpty<(Node<Key, Value>, Rank)>,
+        mut search_path: Vec<TreeLayer<Key, Value>>,
+        mut delta: Delta<Blake3Hash, Buffer>,
+    ) -> Result<(Blake3Hash, Delta<Blake3Hash, Buffer>), DialogSearchTreeError> {
+        const MINIMUM_RANK: u32 = 2;
+        let mut minimum_rank = MINIMUM_RANK;
```
Gozala, line 284: Why is it 2 ? Maybe add a comment. Doc comment made me think you could do merge from any depth within a tree, yet rank here assumes segments ?

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -0,0 +1,517 @@
...
+
+                match search_path.pop() {
+                    Some(layer) => {
+                        delta.subtract(layer.host.hash());
+                        // TBD if we must recompute rank for siblings references
+                        // when building up the tree. Attempt to try setting
+                        // rank to `0` for references outside of the modified
+                        // path.
+                        let ranked_left_siblings = layer.left_siblings.map(into_ranked_links);
+                        let ranked_right_siblings = layer.right_siblings.map(into_ranked_links);
```
Gozala, line 311: I think it is safe to assume here that all nodes (left and right) but last one on the right will have same rank (otherwise they'd have different parents) so it should be possible to just calculate two ranks and deduce the rest.

rust/dialog-search-tree/src/shaper.rs
Gozala: After more time looking at `merge_with_path` I think I'm convinced it does zipping correctly, however it still seems that we should not be creating a segment here as removed node may cause entries from this segment to overflow into next one which I don't believe is accounted for.

It also appears to me that there is no obvious reason for `merge_with_path` to take nodes as opposed to entries directly as in case of overflow we'd be creating segment that is both invalid and should not be stored.

rust/dialog-search-tree/src/storage.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub fn new(backend: Backend) -> Self {
+        Self { backend }
+    }
+
+    /// Stores bytes under their content hash, verifying the hash matches.
+    pub async fn store(
+        &mut self,
+        bytes: Vec<u8>,
+        expected_identity: &Blake3Hash,
+    ) -> Result<(), DialogStorageError> {
```
Gozala, line 29: Guided by discussions around effects and more broadly application / platform separation I'm inclined to suggest that all effectful functions/methods should be taking base capabilities in one argument kind of like

```rs
async fn greet <Capability: Stdout + BlobStore>(&mut env) -> Result<(), Err> {
   let name = BlobStore::get(env, b"user_name").await?;
   Stdout::print(env, format!("Hello {}", String::from_utf8(name))).await?
   Ok(())
}
```

That is to say `ContentAddressedStorage` should not be a thing passed into effectful method tree methods, instead `ContentAddressedStorage` should have `store` / `retrieve` methods that take `Backend` so more like

 ```suggestion
    pub async fn store <Env: StorageBackend<Key = Blake3Hash, Value = Vec<u8>, Error = DialogStorageError>> (
        bytes: Vec<u8>,
        expected_identity: &Blake3Hash,
        env: &mut Env,
    ) -> Result<(), DialogStorageError> {
```

That way tree methods would also require same Env and do something like `ContentAddressedStorage::store(bytes, hash, env)`.

With macros it would be basically the same except without having add bounds manually or threading `env`. 

```suggestion
    #[effectful(StorageBackend<Key = Blake3Hash, Value = Vec<u8>, Error = DialogStorageError>)]
    pub fn store(
        bytes: Vec<u8>,
        expected_identity: &Blake3Hash,
    ) -> Result<(), DialogStorageError> {
      if !expected_identity.matches(&bytes) {
            return Err(DialogStorageError::Verification(
                "Cannot store the provided bytes".to_string(),
            ));
        }

        perform!(Backend.set(expected_identity.clone(), bytes))?;
        
        Ok(())
    }
```

P.S.: Not implying we do this, it's just you asked during the call how it would look like.

rust/dialog-search-tree/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+//! # tokio_test::block_on(async {
+//! use dialog_search_tree::{Tree, ContentAddressedStorage};
+//! use dialog_storage::MemoryStorageBackend;
+//!
+//! let mut storage = ContentAddressedStorage::new(MemoryStorageBackend::default());
+//! let tree_v1 = Tree::<[u8; 4], Vec<u8>>::empty();
+//!
+//! // Create version 1 with some data
+//! let tree_v1 = tree_v1.insert([0, 0, 0, 1], vec![1], &storage).await.unwrap();
+//! let tree_v1 = tree_v1.insert([0, 0, 0, 2], vec![2], &storage).await.unwrap();
```
Gozala, line 101:
```suggestion
//! // Create version 1 with some data
//! let tree_v1 = Tree::<[u8; 4], Vec<u8>>::empty()
//!     .insert([0, 0, 0, 1], vec![1], &storage).await.unwrap()
//!     .insert([0, 0, 0, 2], vec![2], &storage).await.unwrap();
```

This seem bit less confusing to me

rust/dialog-search-tree/src/tree.rs
cdata: It might be, but I spent so long wrestling this into shape that I am hesitant to try to refine it further. This could definitely be more clear.

rust/dialog-search-tree/src/tree.rs
cdata: There is no mutable access of storage in this method. Nor is there any mutation of the `Tree` (instead, it returns a new `Tree` branched from this one).

rust/dialog-search-tree/src/tree.rs
cdata: Internally it uses an LRU-like scheme (https://cachemon.github.io/SIEVE-website/) to organize by frequency, but this is somewhat arbitrary at this time.

As for controversy:

- This cache is more specialized than the LRU layer in `dialog-storage`, because it works with `Buffer` (which internally caches computed hashes and wraps bytes in an `Arc` for cheap cloning)
- The `Tree` is the primary beneficiary of a cache like this, because it cares more about content-addressed storage than other concepts in this domain
  - Bearing in mind: you yourself expressed uncertainty / consternation that the cache was bleeding into other domains and creating problems
  - So: we definitely want caching in the `Tree`; maybe we don't want it anywhere else

rust/dialog-search-tree/src/tree.rs
cdata: Down with the name change, but `Option<R>` signature to me suggests implicit behavior (and I would prefer that the caller act explicitly to stream the whole tree).

rust/dialog-search-tree/src/shaper.rs
cdata: I would like to suggest this be a follow-up change, as it's an optimization and actually it complicates our use of `rkyv` significantly.

rust/dialog-search-tree/src/shaper.rs
cdata: I'm not sure, actually. This is a vestige of the original implementation. I'll look into it.

rust/dialog-search-tree/src/error.rs
cdata: In cases like this, it can be instructive to look where the error is used. Often the concrete problem is downstream of fallible library code that is used to work on the subject in question. `rkyv` access / deserialize is fallible (as you say, the entry may be corrupt) and so the reported error will be `Problem with tree entry: <some rkyv deserialization error>`).

rust/dialog-search-tree/src/error.rs
cdata: I don't necessarily disagree that the errors could be formatted differently, but I'm probably not going to take action on them in this PR. I'll file an issue for us to revisit them.

rust/dialog-search-tree/src/link.rs
cdata: `Link` is more legible to me, as in a content address pointing to another block a la IPLD.

rust/dialog-search-tree/src/node.rs
cdata: the content is either a list of `Link` or `Entry`. A `Buffer` is raw, serialized bytes. I can't do anything with a `Buffer` until I "access" or deserialize it.

rust/dialog-search-tree/src/buffer.rs
cdata: We should have some discussion about jargon. To me a blob store is something I use to persist files (which I would call blobs). A buffer is an allocated chunk of memory where I store some bytes.

rust/dialog-search-tree/src/body.rs
cdata: `Archived*` is code-generated by `rkyv`. Every `Archived*` struct can be realized in memory without actually deserializing bytes.

rust/dialog-search-tree/src/cache.rs
cdata: This is actually downstream of me fixing compiler errors, having to get stricter on the bounds, and forgetting to remove the other.

rust/dialog-search-tree/src/delta.rs
cdata: I'm not opposed to the name, but the rule you are applying seems somewhat arbitrary to me. Why can't `Delta` have branches?

rust/dialog-search-tree/src/shaper.rs
cdata: Yes, I fixed this in the WIP integration of the tree into `dialog-artifacts`.

rust/dialog-search-tree/src/distribution.rs
cdata: I'm not sure how you would prefer to manage this, but I'm open to suggestions. The best answer I have is to make the tree generic over its distribution, or else perhaps pipe through the branch factor as a const generic like we did before.

From my PoV it's far, far simpler to assume that all search trees have a well-understood shape. After all, we won't be working with search trees of arbitrary branching factor in the wild, so ideally our tests would reflect that.

rust/dialog-search-tree/src/shaper.rs
cdata: As I said above, fixed in follow-up work.

rust/dialog-search-tree/src/shaper.rs
cdata: I'll expand on the delta branching semantics, it's kind of scattered in the public documentation on various methods. But, TL;DR every mutation of the tree "branches"/"forks" (jargon TBD per other comment) the `Delta` so that different versions of the tree can flush independently.

rust/dialog-search-tree/src/shaper.rs
cdata: Fixed in follow-up change.

rust/dialog-common/src/sync.rs
cdata: `Send + Sync` is not the same as `Sync`, so we should probably change the name. I'm not sure why I didn't use that naming convention, I first created these a long time ago.

rust/dialog-search-tree/src/tree.rs
Gozala: I just meant to say we can probably alias all those bounds into one named thing

rust/dialog-search-tree/src/delta.rs
Gozala: I'm fine with `branch` and I can't really explain why it feels less intuitive.

rust/dialog-search-tree/src/distribution.rs
Gozala: I think you know better what's the right way to go about it. Either of options you suggested would work I suppose. I just think we need a way to simulate trees of certain shape to have a way of spec-ing what traversals take place when comparing two trees. Trying to test diffing without it seemed intractable.

Gozala: We can probably pun on this until we're trying to integrate #127

rust/dialog-search-tree/src/error.rs
Gozala: Yeah but we could capture key of the entry no ? That way it provides at least some context for investigation.

rust/dialog-search-tree/src/link.rs
Gozala: Except it's not what this is Block3Hash is the Link where's this has routing information as in entires in these key range can be found under (hash) linked node.

rust/dialog-search-tree/src/tree.rs
Gozala: I think some of this as a code comment would be very valuable

rust/dialog-search-tree/src/tree.rs
Gozala: Might be worth adding note in the function description about the fact that storage is only used to read nodes on the insertion path and all nodes are added to delta that can be flushed.

PR #135, opened 2026-01-13
feat(s3): `Authorizer` trait with UCAN support

jackddouglas:
## Summary
- introduces async `Authorizer` trait for pluggable S3 authorization strategies
- adds `UcanAuthorizer` for UCAN-based authorization via external access services
- refactors S3 backend to use new trait, enabling multi-subject backends

## Description
1. Async `Authorizer` trait (`access.rs`): interface for authorizing S3 requests
    - `authorize()`: async authorization returning presigned URLs
    - `build_url()`: URL construction for bucket/key paths
    - `region()` / `path_style()`: configuration accessors
2. Updated `Credentials` and `Public` types
    - now implement `Authorizer`
    - encapsulate `Address` (endpoint, region, bucket), removing need to pass separately
3. New `UcanAuthorizer` (`ucan.rs`)
    - supports multiple subject DIDs with separate delegation chains
    - builds UCAN invocations signed by user-provided operator ID
    - delegates to external access service returning pre-signed URLs via 307 redirects

## Breaking Changes
1. `Credentials` and `Public` constructors now require `Address` and return `Result`
2. `Credentials` fields now private, use `access_key_id()` accessor instead
3. `authorize()` method renamed to `sign()`, external code should use `Authorizer::authorize()` trait method
4. S3 backend now generic over `Authorizer`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -134,7 +134,10 @@
             nativeBuildInputs = common-build-inputs "stable";
             cargoLock = {
               lockFile = ./Cargo.lock;
-              outputHashes = { };
+              outputHashes = {
+                "ucan-0.5.0" = "sha256-5KQ7wIXv7PHgd6y1pq0+aUU/VFW7BLxECmVUNk1JfGw=";
+                "varsig-0.1.0" = "sha256-5KQ7wIXv7PHgd6y1pq0+aUU/VFW7BLxECmVUNk1JfGw=";
+              };
             };
           };
 


.vscode/settings.json
```diff
@@ -5,10 +5,11 @@
             "async_trait"
         ]
     },
+    "rust-analyzer.diagnostics.disabled": ["proc-macro-disabled"],
```
Gozala, line 8: I don't use vscode, so it's up to you and @cdata to agree on

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -212,19 +225,10 @@ where
...
-    bucket: String,
     /// Optional prefix/directory within the bucket
     path: Option<String>,
-    /// Use path-style URLs (bucket in path) vs virtual-hosted (bucket in subdomain)
-    /// Defaults to true for IP addresses and localhost, false otherwise.
-    path_style: bool,
-    /// Credentials for authorizing requests (None for public access)
-    credentials: Option<Credentials>,
+    /// Authorizer for signing requests and building URLs
+    authorizer: Arc<dyn Authorizer>,
```
Gozala, line 231: I believe dyn(amic) dispatch does come with tradeoffs that we historically tried to avoid. My level in Rust makes it hard for me to evaluate tradeoffs here, but I wonder if turning Credentials into an enum with fixed number of variants would be a more pragmatic approach without any real tradeoffs. UCAN based authorizer could simply be feature geated.

e.g we could have something like:

```rs
enum Credentials {
   S3(BucketCredentials),
   #[cfg(feature = "ucan")]
   UCAN(UCANCredentials)
}
```

I assume here that `s3::Credentials` is also an enum like

```rs
enum BucketCredentials {
   Public { ... },
   Private { ... }
}
```

Gozala: Maybe @cdata can offer more nuanced view in regards to tradeoffs.

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -343,58 +313,23 @@ where
...
 
     /// Get the region for signing.
     pub fn region(&self) -> &str {
-        &self.region
-    }
-
-    /// Get the bucket name.
-    pub fn bucket(&self) -> &str {
-        &self.bucket
+        self.authorizer.region()
```
Gozala, line 332: I wonder if bucket even need to expos `region` , e.g. with ucan-access maybe region is not a thing or not a fixed thing anyway.

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -408,52 +343,25 @@ where
...
         let encoded_key = encode_s3_key(key);
-        let full_key = match &self.path {
+        let full_path = match &self.path {
             Some(prefix) => format!("{}/{}", prefix, encoded_key),
             None => encoded_key,
         };
 
-        self.build_url(&full_key)
+        self.authorizer
+            .build_url(&full_path)
```
Gozala, line 355: Nit: I have not reviewed this enough to have full picture, but conceptually it feels to me that authorizer should just get a request description and authorize it. That perhaps we do not need resolve method we simply construct Put / Get etc.. with relative paths.

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -877,76 +791,14 @@ mod tests {
     }
 
     #[dialog_common::test]
-    fn it_detects_path_style_default() {
```
Gozala, line 880: Can we have an equivalent of this, as authorizer needs to be able to know to use path style URLs for specific types of addresses

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -877,76 +791,14 @@ mod tests {
...
-            &ipv4_other
-        ));
-
-        // Remote domains should use virtual-hosted
-        let remote = Url::parse("https://s3.amazonaws.com").unwrap();
-        assert!(!Bucket::<Vec<u8>, Vec<u8>>::is_path_style_default(&remote));
-    }
-
-    #[dialog_common::test]
-    fn it_generates_signed_urls() {
```
Gozala, line 902: Can we have equivalent of this test, it just checks that pre-signed URLs do have expected query parameters

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -966,9 +818,8 @@ mod tests {
     async fn it_sets_and_gets_values(env: PublicS3Address) -> anyhow::Result<()> {
         // Using public access for simplicity. Signed sessions are tested separately.
         let address = Address::new(&env.endpoint, "us-east-1", &env.bucket);
-        let mut backend = Bucket::<Vec<u8>, Vec<u8>>::open(address, None)?
-            .with_path_style(true)
-            .at("test");
+        let authorizer = Public::new(address)?.with_path_style(true);
```
Gozala, line 865: Nit: I find it `Public::new` returning Result to be controversial, as in I expect `::new` to return `Self`, all the validation / parsing should occur prior. Or there should e something like `Public::create(address)` that could fail.

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -1388,226 +1247,19 @@ mod tests {
...
-        // but the bytes get normalized through UTF-8 replacement characters.
-        let test_keys: Vec<Vec<u8>> = vec![
+        // Test that encode and decode are inverse operations
+        let test_cases = vec![
             b"simple-key".to_vec(),
             b"path/to/key".to_vec(),
             b"key with spaces".to_vec(),
-            b"key@with!special#chars".to_vec(),
-            b"safe/unsafe@mixed/safe2".to_vec(),
+            b"key!with@special#chars".to_vec(),
```
Gozala: What is the reason for changing one fixture and removing the other ?

rust/dialog-storage/src/storage/backend/s3.rs
```diff
@@ -1388,226 +1247,19 @@ mod tests {
...
         }
-
-        assert_eq!(items.len(), 2);
-
-        // Verify the items (order may vary)
-        let keys: Vec<&[u8]> = items.iter().map(|(k, _)| k.as_slice()).collect();
-        assert!(keys.contains(&b"a".as_slice()));
-        assert!(keys.contains(&b"b".as_slice()));
-
-        Ok(())
```
Gozala, line 1611: These are whole lot of tests being deleted which I am guessing is not intentional, if so can we please restore them otherwise can you please provide rational for removing them.

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: Alternatively we could make `Bucket` generic so something like

```rs
pub struct Bucket<Key, Value, Credentials: Authorizer> {
   authorizer: Credentials
   // ...
}
```

 Which would make it more open ended than enum, although carrying around type parameter has it's downsides

rust/dialog-storage/examples/ucan_e2e_test.rs
Gozala, line 1: Can we turn this into actual test ? You can create ucan authorizer service in helper similar to how we create local s3 server we do something like what we do in other tests, specifically  you'll need to

1. Define an address for the service your integration test will need

https://github.com/dialog-db/dialog-db/blob/a4995749d5eddc53e1884128053be102b11ba28e/rust/dialog-storage/src/storage/backend/s3/helpers.rs#L12-L22

2. Create test that receives it as an argument

https://github.com/dialog-db/dialog-db/blob/a4995749d5eddc53e1884128053be102b11ba28e/rust/dialog-storage/src/storage/backend/s3.rs#L1442-L1463

3. Define service provider that will start the service and give you back address 

https://github.com/dialog-db/dialog-db/blob/a4995749d5eddc53e1884128053be102b11ba28e/rust/dialog-storage/src/storage/backend/s3/helpers/server.rs#L361-L382

4. Provider function will need to take something that implements default

https://github.com/dialog-db/dialog-db/blob/a4995749d5eddc53e1884128053be102b11ba28e/rust/dialog-storage/src/storage/backend/s3/helpers/server.rs#L333-L352

You can learn about more details here https://github.com/dialog-db/dialog-db/blob/a4995749d5eddc53e1884128053be102b11ba28e/rust/dialog-common/src/lib.rs#L45-L100


In this case I expect you can have an Address that contains details for ucan-access service and it's provider starts both local s3 (through linked function) and ucan-access service.

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -0,0 +1,582 @@
...
+    pub proofs: Vec<Vec<u8>>,
+
+    /// CIDs of the proofs (for inclusion in invocations).
+    ///
+    /// These content identifiers reference the proofs and are included
+    /// in the UCAN invocation to link it to the delegation chain.
+    pub proof_cids: Vec<Cid>,
+}
+
+impl DelegationChain {
```
Gozala, line 64: It seems that methods like subject and audience might be relevant to knowing what is the subject of the authorization and who is authorized.

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -0,0 +1,582 @@
...
+
+/// Operator identity for signing UCAN invocations.
+///
+/// This wraps an Ed25519 signing key and provides methods to create
+/// UCAN invocations. The operator is the entity making requests on
+/// behalf of a subject (who has delegated authority to them).
+#[derive(Clone)]
+pub struct OperatorIdentity {
+    signer: Ed25519Signer,
+}
```
Gozala, line 87: I'm bit lost on what this provides it seems to just wrap signer and allow unwrapping it, is there reason why it's not `Ed25519Signer` directly ?

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -0,0 +1,582 @@
...
+        self.signer.did()
+    }
+
+    /// Returns reference to the inner signer for creating invocations.
+    pub(crate) fn signer(&self) -> &Ed25519Signer {
+        &self.signer
+    }
+}
+
+/// UCAN-based authorizer that delegates to an external access service.
```
Gozala, line 125: This is confusing maybe something, like

```suggestion
/// Authorizer that can be used to authorize S3 bucket access through UCAN capabilities.
```

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -0,0 +1,582 @@
...
+///
+/// This authorizer implements the [`Authorizer`] trait by:
+///
+/// 1. Extracting the subject DID from the request URL path (first path segment)
+/// 2. Looking up the delegation chain for that subject
+/// 3. Building a UCAN invocation signed by the operator
+/// 4. Sending the invocation to the configured access service
+/// 5. Returning the pre-signed URL from the service's 307 redirect response
```
Gozala, line 133: Seeing all the pieces together I kind of regret suggesting redirect based API. I kind of assumed that we would be able to take advantage of transparent redirects, but in practice it seems like we are creating more surface for error specifically

1. URL not matching subject / key / path is an error.
2. Extra checksum header hat needs to align
3. Method that needs to align

I think it may be a lot cleaner if we clearly define set of capabilities that backend can submit and let the service provide URL, set of headers and HTTP method that can be used to complete the task.

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -0,0 +1,582 @@
...
+///
+/// # Example
+///
+/// ```ignore
+/// let authorizer = UcanAuthorizer::builder()
+///     .service_url("https://access.example.com")
+///     .operator(operator)
+///     .delegation("did:key:z6MkSubject1...", chain1)
+///     .delegation("did:key:z6MkSubject2...", chain2)
+///     .build()?;
```
Gozala, line 149: I think you may have misunderstood what I was suggesting I meant authorizer could have a bag of delegation chains keyed by audience (not subject), that is because given an operator in use authorizer would be able to pick relevant delegation chain. However that assumed a future where dialog passes operator in use which is not yet the case. For this reason I think it may make sense to keep things simpler as usage would be something more like

```rs
repository.branch('main').set_upstream(Upstream::Remote {
   repository: subject,
   credentials: UCANAuthorizer::builder()
        .service_url(&ucan_service)
        .issuer(signer)
        .delegation(chain_from_repo_to_signer_for_subject)
})
```

This way when we push or pull from the remote we'll have a way to authorize block read / writes.

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -0,0 +1,582 @@
...
+/// ```
+#[derive(Debug, Clone)]
+pub struct UcanAuthorizer {
+    service_url: String,
+    operator: OperatorIdentity,
```
Gozala, line 154: 🤔 I wonder if we should make on `OperatorIdentity` into a trait or wrapper around Signer that can not leak the private, instead it could only expose method for signing payloads. This way key could not be leaked.

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -0,0 +1,582 @@
...
+/// - `operator`: The operator identity for signing invocations
+///
+/// # Optional Fields
+///
+/// - `delegation`: One or more subject DID to delegation chain mappings
+#[derive(Default)]
+pub struct UcanAuthorizerBuilder {
+    service_url: Option<String>,
+    operator: Option<OperatorIdentity>,
+    delegations: HashMap<String, DelegationChain>,
```
Gozala, line 202: I think you should call this `authorizations` and rename `DelegationChain` to `Authorization`, or alternatively you could keep the struct name and rename field to `delegation_chains`. This is because I found myself multiple times as I was reading code and seeing `.delegation` assuming it would be a `UcanDelegation` as opposed to chain.

.vscode/settings.json
jackddouglas: i don't either, this is just a standard for sending flags to rust-analyzer

rust/dialog-storage/src/storage/backend/s3.rs
jackddouglas: for network I/O `dyn` seems negligible to me but fair enough if you generally want to avoid.

my gripe with the enum approach is that users couldn't extend `Authorizer`, and feature gating while having to match for every variant would be nasty.

generic `Bucket` makes sense to me, we'd lose the `dyn` overhead and type parameter would probably be inferred most of the time?

rust/dialog-storage/src/storage/backend/s3.rs
jackddouglas: yep could make it `pub(crate)` 👍

rust/dialog-storage/src/storage/backend/s3.rs
jackddouglas: you're basically asking if we should separate URL building from key/path logic?

we could have `Put` store a path instead of URL, then `perform()` would build URL from `authorizer`.

we could also have `Authorizer` handle everything but this would couple it to `reqwest` and move HTTP execution. i think this is wrong because the trait should handle URL construction (since it's different between UCAN and public/private local signing) as well as singing (it's authorization after all). More than that feels like scope creep to me

rust/dialog-storage/src/storage/backend/s3.rs
jackddouglas: we'd need to play around with where the URL validation will happen. could be pushed up to `Address`, but same thing where you'd want `Address::new` to be infallible and return Self.

could also accept `Url` in `Address` so the parsing would happen even earlier.

either way, would be nice to know the `Address` is valid at creation time so we don't have to handle repeated errors if invalid

rust/dialog-s3-credentials/src/ucan.rs
jackddouglas: mostly it provides an api boundary, Debug just displays the DID, and `from_secret` is helpful for creating an identity without `SigningKey::from_bytes` intermediary step

rust/dialog-s3-credentials/src/ucan.rs
jackddouglas: capabilities could be good. we could also move requests to the body then and maybe avoid this issue you brought up with caps on header sizes. we could also get more descriptive ucan commands out of this, something like `/blob/read` etc

rust/dialog-s3-credentials/src/ucan.rs
jackddouglas: `signer()` is already `pub(crate)` for this reason, maybe i'm missing something but why would a trait be better here?

rust/dialog-storage/src/storage/backend/s3.rs
jackddouglas: was playing around with tests a little so this and the above modifications shouldn't have made it in, sorry!

rust/dialog-storage/src/storage/backend/s3.rs
jackddouglas: actually generic `Bucket` won't work, because in `dialog-artifacts/src/replica.rs` we have `pub type RemoteBackend = ErrorMappingBackend<S3Bucket<Vec<u8>, Vec<u8>>>;` which would need to be extended to include an Authorizer type.

so that leaves us with going back to dynamic dispatch, or we can prevent users from implementing their own Authorizers and using an enum approach

rust/dialog-storage/src/storage/backend/s3.rs
jackddouglas: this is tested in `dialog-s3-credentials`

rust/dialog-storage/src/storage/backend/s3.rs
jackddouglas: also tested elsewhere but i'll bring it back here too

rust/dialog-storage/src/storage/backend/s3.rs
Gozala: No I meant you give authorizer command you want to perform and it returns abstract description of the HTTP request that is url + headers than it’s up to the bucket to turn that into reqwest or whatever else call to do the job. In case of Set authorizer doesn’t even need content being written

rust/dialog-s3-credentials/src/ucan.rs
jackddouglas: so you're saying just store one delegation, and we don't need to extract the subject did because it would already be known from the remote config?

none of the above methods exist as far as i can tell, so not really sure what you're suggesting here

PR #145, opened 2026-01-22
feat: support Web Crypto API

jackddouglas:
Enables dialog to work with async signing like Web Crypto API, supporting non-extractable keys.

## Changes
- updated `rs-ucan` dependency to `tonk` branch which includes `WebCryptoEd25519Signer` and async signing traits
- refined WASM target detection from `target_arch = "wasm32"` to `all(target_arch = "wasm32", target_os = "unknown")` to distinguish from WASI
- added integration test feature flags (`integration-tests`, `web-integration-tests`) to `dialog-s3-credentials`

## Testing
- `it_generates_webcrypto_signer` - verifies key generation produces valid DIDs
- `it_produces_valid_webcrypto_signature` - verifies signatures can be verified
- `it_rejects_wrong_message` - verifies signature verification fails for tampered messages
- `it_produces_different_signatures_from_different_signers` - verifies key uniqueness
- `it_builds_invocation_with_webcrypto_signer` - verifies UCAN invocations work with async signing
- `it_serializes_webcrypto_invocation` - verifies DAG-CBOR serialization roundtrip

`wasm-pack test --headless --chrome rust/dialog-s3-credentials`

### DIFF
--- a/rust/dialog-artifacts/src/replica.rs
+++ b/rust/dialog-artifacts/src/replica.rs
@@ -3015,7 +3015,8 @@ mod tests {
             &operator_did,
             operator_signer.did(),
             &["archive", "memory"],
-        )?;
+        )
+        .await?;
         let delegation_chain = DelegationChain::new(delegation);
 
         // Create UCAN credentials
@@ -3104,7 +3105,8 @@ mod tests {
             &operator_did,
             operator_signer.did(),
             &[], // Empty command = root capability (/)
-        )?;
+        )
+        .await?;
         let delegation_chain = DelegationChain::new(delegation);
 
         // Step 3: Create UCAN credentials
@@ -3215,7 +3217,8 @@ mod tests {
             &second_operator_did,
             operator_signer.did(),
             &[], // Empty command = root capability (/)
-        )?;
+        )
+        .await?;
         let second_delegation_chain = DelegationChain::new(second_delegation);
         let second_ucan_credentials =
             ucan::Credentials::new(env.access_service_url.clone(), second_delegation_chain);
--- a/rust/dialog-s3-credentials/Cargo.toml
+++ b/rust/dialog-s3-credentials/Cargo.toml
@@ -56,8 +56,6 @@ getrandom = { workspace = true, features = ["js"] }
 dialog-common = { workspace = true, features = ["helpers"] }
 serde_json = { workspace = true }
 tokio = { workspace = true, features = ["rt", "macros"] }
-
-[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dev-dependencies]
 wasm-bindgen-test = { workspace = true }
 
 [lints.rust]
--- a/rust/dialog-s3-credentials/src/ucan.rs
+++ b/rust/dialog-s3-credentials/src/ucan.rs
@@ -31,5 +31,5 @@ pub use provider::UcanAuthorizer;
 /// Only available with the `helpers` feature.
 #[cfg(feature = "helpers")]
 pub mod test_helpers {
-    pub use super::delegation::tests::{create_delegation, generate_signer};
+    pub use super::delegation::helpers::{create_delegation, generate_signer};
 }
--- a/rust/dialog-s3-credentials/src/ucan/credentials.rs
+++ b/rust/dialog-s3-credentials/src/ucan/credentials.rs
@@ -175,7 +175,7 @@ where
 pub mod tests {
     use super::*;
     use crate::capability::archive;
-    use crate::ucan::delegation::tests::create_delegation;
+    use crate::ucan::delegation::helpers::create_delegation;
     use anyhow;
     use dialog_capability::{Authority, Authorization, Did, Principal, Subject};
     use dialog_common::Blake3Hash;
@@ -184,13 +184,14 @@ pub mod tests {
     use ucan::promise::Promised;
 
     /// Helper to create a test delegation chain from subject to operator.
-    pub fn test_delegation_chain(
+    pub async fn test_delegation_chain(
         subject_signer: &ucan::did::Ed25519Signer,
         operator_did: &Ed25519Did,
         ability: &[&str],
     ) -> DelegationChain {
         let subject_did = subject_signer.did();
         let delegation = create_delegation(subject_signer, operator_did, subject_did, ability)
+            .await
             .expect("Failed to create test delegation");
         DelegationChain::new(delegation)
     }
@@ -254,7 +255,7 @@ pub mod tests {
 
         let credentials = Credentials::new(
             "https://access.ucan.com".into(),
-            test_delegation_chain(&operator, operator.did(), &["archive"]),
+            test_delegation_chain(&operator, operator.did(), &["archive"]).await,
         );
 
         let mut session = Session::new(credentials, &[0u8; 32]);
@@ -270,7 +271,7 @@ pub mod tests {
             .acquire(&mut session)
             .await?;
 
-        let authorization = read.authorization().invoke(&session)?;
+        let authorization = read.authorization().invoke(&session).await?;
 
         let ucan = match authorization {
             UcanAuthorization::Invocation { chain, .. } => chain,
@@ -307,7 +308,7 @@ pub mod tests {
         // Create credentials with a delegation (won't be used for self-auth)
         let credentials = Credentials::new(
             "https://access.ucan.com".into(),
-            test_delegation_chain(&operator, operator.did(), &["archive"]),
+            test_delegation_chain(&operator, operator.did(), &["archive"]).await,
         );
 
         let mut session = Session::new(credentials, &[0u8; 32]);
@@ -344,7 +345,7 @@ pub mod tests {
 
         let credentials = Credentials::new(
             "https://access.ucan.com".into(),
-            test_delegation_chain(&operator, operator.did(), &["archive"]),
+            test_delegation_chain(&operator, operator.did(), &["archive"]).await,
         );
 
         let mut session = Session::new(credentials, &[0u8; 32]);
@@ -362,7 +363,7 @@ pub mod tests {
             .await?;
 
         // Invoke the authorization - should create an Invocation with empty proofs
-        let authorization = authorized.authorization().invoke(&session)?;
+        let authorization = authorized.authorization().invoke(&session).await?;
 
         let ucan = match authorization {
             UcanAuthorization::Invocation { chain, .. } => chain,
@@ -402,6 +403,7 @@ pub mod tests {
     #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
     mod webcrypto_tests {
         use signature::Verifier;
+        use ucan::did::Did;
         use ucan::{AsyncDidSigner, WebCryptoEd25519Signer};
         use wasm_bindgen_test::wasm_bindgen_test_configure;
 
@@ -428,10 +430,7 @@ pub mod tests {
                 .expect("Failed to generate signer");
             let msg = b"test message for WebCrypto signing";
 
-            let signature = signer
-                .sign_async(msg)
-                .await
-                .expect("Failed to sign message");
+            let signature = signer.sign(msg).await.expect("Failed to sign message");
 
             let verifier = signer.did().verifier();
             verifier
@@ -447,10 +446,7 @@ pub mod tests {
             let msg = b"original message";
             let wrong_msg = b"wrong message";
 
-            let signature = signer
-                .sign_async(msg)
-                .await
-                .expect("Failed to sign message");
+            let signature = signer.sign(msg).await.expect("Failed to sign message");
 
             let verifier = signer.did().verifier();
             assert!(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,9 +43,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.20"
+version = "0.6.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3ae563653d1938f79b1ab1b5e668c87c76a9930414574a6583a7b7e11a8e6192"
+checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
 dependencies = [
  "anstyle",
  "anstyle-parse",
@@ -58,9 +58,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.11"
+version = "1.0.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "862ed96ca487e809f1c8e5a8447f6ee2cf102f846893800b20cebdf541fc6bbd"
+checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
 
 [[package]]
 name = "anstyle-parse"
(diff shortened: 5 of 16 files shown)


rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -586,4 +902,107 @@ mod tests {
         assert!(authorizer.has_delegation("did:key:z6MkOne"));
         assert!(authorizer.has_delegation("did:key:z6MkTwo"));
     }
+
+    // Tests for async types
+    #[test]
```
cdata: We should use the `dialog_common::test` macro for all tests unless there is a good reason not to. At a minimum, we aren't actually running tests in a web browser unless we are using `wasm_bindgen_test` (which `dialog_common::test` will use) . A change that adds support for `WebCrypto` should make sure that these tests run in a browser.

rust/dialog-artifacts/src/replica.rs
```diff
@@ -162,21 +174,61 @@ impl Issuer {
...
+/// This trait enables signing with keys that don't expose their private bytes,
+/// such as Web Crypto non-extractable keys or hardware security modules.
+///
+/// # Platform Differences
+///
+/// On native targets, implementors must be `Send + Sync` to allow use across threads.
+/// On WASM, these bounds are relaxed since WASM is single-threaded.
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait Issuer: Clone + dialog_common::ConditionalSync + 'static {
```
cdata: nit: `use dialog_common::ConditionalSync`, Rust bounds are noisy enough as it is.

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -78,25 +85,50 @@ impl DelegationChain {
...
+pub struct OperatorIdentity<S> {
+    signer: S,
 }
 
-impl std::fmt::Debug for OperatorIdentity {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("OperatorIdentity")
-            .field("did", &self.did().to_string())
-            .finish()
+impl<S: AsyncDidSigner> OperatorIdentity<S>
```
cdata: We can address this elsewhere (since the trait is from another crate), but my preference would be that `DidSigner` is canonically "async".

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -443,19 +473,157 @@ fn method_to_command(method: &str) -> Result<Vec<String>, AuthorizationError> {
     }
 }
 
+/// Prepared request data for authorization.
+///
+/// Contains all the validated and parsed data needed to build a UCAN invocation.
+struct PreparedRequest<'a> {
```
cdata: Seems fine in isolation but it seems likely that you'll want to revisit this with the capability PR in mind.

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -443,19 +473,157 @@ fn method_to_command(method: &str) -> Result<Vec<String>, AuthorizationError> {
...
+            command,
+            access_url,
+        })
+    }
+}
+
+/// Send the invocation to the access service and handle the response.
+///
+/// This is the shared HTTP handling logic for authorization.
+async fn send_to_access_service(
```
cdata: Ditto: https://github.com/dialog-db/dialog-db/pull/145/changes#r2733790478

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -443,19 +473,157 @@ fn method_to_command(method: &str) -> Result<Vec<String>, AuthorizationError> {
...
+        .send()
+        .await
+        .map_err(|e| AuthorizationError::AccessService(e.to_string()))?;
+
+    // Handle response - expect 307 redirect
+    if response.status().as_u16() != 307 {
+        let status = response.status();
+        let body = response.text().await.unwrap_or_default();
+        return Err(AuthorizationError::AccessService(format!(
+            "Access service returned {}: {}",
```
cdata: May be good to qualify this as "unexpectedly returned," since even an ostensibly "good" status code (e.g., 200) will provoke the error condition, and "The error was: 200" kind of reads like "The error was: success."

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -443,19 +473,157 @@ fn method_to_command(method: &str) -> Result<Vec<String>, AuthorizationError> {
...
+    client: &reqwest::Client,
+    request: &RequestInfo,
+    prepared: &PreparedRequest<'_>,
+    invocation_bytes: Vec<u8>,
+) -> Result<Authorization, AuthorizationError> {
+    // Build HTTP request
+    let mut req = match request.method {
+        "GET" => client.get(&prepared.access_url),
+        "PUT" => client.put(&prepared.access_url),
+        "DELETE" => client.delete(&prepared.access_url),
```
cdata: May be able to avoid some noise here by using [`client.request`](https://docs.rs/reqwest/latest/reqwest/struct.Client.html#method.request)

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -466,62 +634,84 @@ mod tests {
...
     fn test_operator_identity_deterministic() {
         let secret = [42u8; 32];
         let identity1 = OperatorIdentity::from_secret(&secret);
         let identity2 = OperatorIdentity::from_secret(&secret);
-        assert_eq!(identity1.did().to_string(), identity2.did().to_string());
+        assert_eq!(identity1.did(), identity2.did());
+    }
+
+    #[dialog_common::test]
+    fn test_operator_identity_from_ed25519_signer() {
```
cdata: nit: prefer BDD-style test name convention to align with all the other tests.

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -586,4 +776,204 @@ mod tests {
...
+
+    /// WebCrypto-specific tests (wasm32 only).
+    ///
+    /// These tests verify that the UCAN authorization flow works correctly
+    /// with WebCrypto-backed signers in browser environments. They exercise
+    /// non-extractable key generation, signing, and signature verification
+    /// using the Web Crypto API.
+    ///
+    /// Run with: `wasm-pack test --headless --chrome rust/dialog-s3-credentials --features ucan`
+    #[cfg(target_arch = "wasm32")]
```
cdata: It's good to feature gate web-specific Wasm where appropriate: `#[cfg(target_arch = "wasm32", target_os = "unknown")]`

Otherwise, this will be ungated for e.g., `wasm32-wasip3` (which does not have Web Crypto support).

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -586,4 +776,204 @@ mod tests {
...
+    mod webcrypto_tests {
+        use super::*;
+        use signature::Verifier;
+        use ucan::WebCryptoEd25519Signer;
+        use ucan::did::{Did, Ed25519Did};
+        use ucan::invocation::builder::InvocationBuilder;
+        use wasm_bindgen_test::*;
+
+        wasm_bindgen_test_configure!(run_in_browser);
```
cdata: You may want to be more specific and `run_in_service_worker` since that's where this stuff will be running in practice: https://docs.rs/wasm-bindgen-test/latest/wasm_bindgen_test/macro.wasm_bindgen_test_configure.html

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -586,4 +776,204 @@ mod tests {
...
+
+        #[wasm_bindgen_test]
```
cdata: You should be able to use `dialog_common::test` here (it will spit out `wasm_bindgen_test` for you).

rust/dialog-s3-credentials/src/ucan.rs
```diff
@@ -266,130 +343,86 @@ impl UcanAuthorizerBuilder {
...
-        }
+#[cfg(target_arch = "wasm32")]
+#[async_trait(?Send)]
+impl<S> Authorizer for UcanAuthorizer<S>
+where
+    S: AsyncDidSigner<Did = Ed25519Did, Signature = ed25519_dalek::Signature>
+        + DidSigner<Did = Ed25519Did>
+        + Clone,
+{
+    async fn authorize(&self, request: &RequestInfo) -> Result<Authorization, AuthorizationError> {
```
cdata: Is this implementation tested (directly or indirectly) somewhere?

rust/dialog-s3-credentials/Cargo.toml
```diff
@@ -14,13 +14,17 @@ ucan = [
...
     "dep:serde_ipld_dagcbor",
     "dep:serde_json",
+    "dep:signature",
     "dep:ucan",
 ]
+integration-tests = ["dialog-common/integration-tests"]
+web-integration-tests = ["dialog-common/web-integration-tests"]
 
 [dependencies]
 async-trait = "0.1"
```
cdata: Can you uplift all these deps to the workspace `Cargo.toml` in this change so that everyone can inherit the same version ranges?

rust/dialog-s3-credentials/Cargo.toml
```diff
@@ -33,11 +37,20 @@ ipld-core = { workspace = true, optional = true }
...
 getrandom = { workspace = true, features = ["js"] }
 
 [dev-dependencies]
+dialog-common = { workspace = true, features = ["helpers"] }
 serde_json = { workspace = true }
+
+[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dev-dependencies]
+wasm-bindgen-test = { workspace = true }
+
+[lints.rust]
```
cdata, line 61: 💎

rust/dialog-s3-credentials/src/ucan.rs
jackddouglas: the challenge is that making `DidSigner` async would mean significant low level breaking changes in `rs-ucan`, so i didn't want to address that as a drive by change. do you or others on rs-ucan have preferences for how that would be handled?

rust/dialog-artifacts/src/replica/remote/connection.rs
```diff
@@ -23,9 +23,9 @@ pub use s3::S3Connection;
 pub enum Connection {
     /// S3-compatible storage connection (includes both direct S3 and UCAN-based access).
     #[cfg(feature = "s3")]
-    S3(S3Connection),
+    S3(Box<S3Connection>),
     /// In-memory storage connection (useful for testing).
-    Memory(MemoryConnection),
+    Memory(Box<MemoryConnection>),
```
Gozala, line 28: Nit: can you offer some insight into why box them ?

rust/dialog-s3-credentials/src/ucan/credentials.rs
```diff
@@ -115,10 +118,22 @@ impl Access for Credentials {
         &self,
         claim: Claim<C>,
     ) -> Result<Self::Authorization, Self::Error> {
-        // Verify the claim's subject matches our delegation's subject
+        let mut parameters = Parameters::new();
+        claim.capability().parametrize(&mut parameters);
```
Gozala, line 133: Nit: I would fin it easier to follow if there were declared after the if blocks kind of where it was before because it isn't used under those conditionals and takes up mental context.

rust/dialog-s3-credentials/src/ucan/credentials.rs
```diff
@@ -279,4 +297,259 @@ pub mod tests {
...
+
+        #[dialog_common::test]
+        async fn it_produces_valid_webcrypto_signature() {
+            let signer = WebCryptoEd25519Signer::generate()
+                .await
+                .expect("Failed to generate signer");
+            let msg = b"test message for WebCrypto signing";
+
+            let signature = signer
+                .sign_async(msg)
```
Gozala: 🤔 Didn't we made sign async by default why do we need `.sign_async` ?

rust/dialog-s3-credentials/src/ucan/credentials.rs
```diff
@@ -279,4 +297,259 @@ pub mod tests {
...
+                .await
+                .expect("Failed to sign message");
+
+            let verifier = signer.did().verifier();
+            verifier
+                .verify(msg, &signature)
```
Gozala, line 437: 🤔 I think verification should also be async

rust/dialog-s3-credentials/src/ucan/credentials.rs
```diff
@@ -105,8 +105,11 @@ impl Credentials {
 ///
 /// This allows Credentials to find authorization proofs for capability claims
 /// by looking up delegation chains for the subject.
-#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
-#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(
+    not(all(target_arch = "wasm32", target_os = "unknown")),
+    async_trait::async_trait
+)]
+#[cfg_attr(all(target_arch = "wasm32", target_os = "unknown"), async_trait::async_trait(?Send))]
```
Gozala, line 112: It isn't obvious to me if excluding `wasm32-wasip3` is intentional here and if so why, I don't think anything here introduces dependency on web API does it ?

rust/dialog-s3-credentials/src/ucan/credentials.rs
```diff
@@ -141,8 +153,8 @@ impl Access for Credentials {
     }
 }
 
-#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
-#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(all(target_arch = "wasm32", target_os = "unknown")), async_trait)]
+#[cfg_attr(all(target_arch = "wasm32", target_os = "unknown"), async_trait(?Send))]
```
Gozala, line 157: Same as above

> It isn't obvious to me if excluding `wasm32-wasip3` is intentional here and if so why, I don't think anything here introduces dependency on web API does it ?

rust/dialog-s3-credentials/src/ucan/credentials.rs
```diff
@@ -200,8 +212,11 @@ pub mod tests {
         }
     }
 
-    #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
-    #[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+    #[cfg_attr(
+        not(all(target_arch = "wasm32", target_os = "unknown")),
+        async_trait::async_trait
+    )]
+    #[cfg_attr(all(target_arch = "wasm32", target_os = "unknown"), async_trait::async_trait(?Send))]
```
Gozala, line 220: Same as above

> It isn't obvious to me if excluding `wasm32-wasip3` is intentional here and if so why, I don't think anything here introduces dependency on web API does it ?

rust/dialog-s3-credentials/src/ucan/credentials.rs
```diff
@@ -218,8 +233,11 @@ pub mod tests {
             &self.did
         }
     }
-    #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
-    #[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+    #[cfg_attr(
+        not(all(target_arch = "wasm32", target_os = "unknown")),
+        async_trait::async_trait
+    )]
+    #[cfg_attr(all(target_arch = "wasm32", target_os = "unknown"), async_trait::async_trait(?Send))]
```
Gozala, line 241: Same as above

> It isn't obvious to me if excluding `wasm32-wasip3` is intentional here and if so why, I don't think anything here introduces dependency on web API does it ?

rust/dialog-s3-credentials/src/ucan/credentials.rs
```diff
@@ -279,4 +297,259 @@ pub mod tests {
...
+
+            let verifier = signer.did().verifier();
+            assert!(
+                verifier.verify(wrong_msg, &signature).is_err(),
+                "Verification should fail for wrong message"
+            );
+        }
+
+        #[dialog_common::test]
+        async fn it_produces_different_signatures_from_different_signers() {
```
Gozala: Nit: I'm not sure there's much value in this test

rust/dialog-s3-credentials/src/ucan/credentials.rs
```diff
@@ -279,4 +297,259 @@ pub mod tests {
...
+
+            // Cross-verification should fail
+            assert!(
+                signer1.did().verifier().verify(msg, &sig2).is_err(),
+                "Signature 2 should not verify with signer 1's key"
+            );
+        }
+
+        #[dialog_common::test]
+        async fn it_builds_invocation_with_webcrypto_signer() {
```
Gozala: Nit: Good test although it does feel out of place here as ti test ucan library and `WebCryptoEd25519Signer` compatibility nothing about credentials.

rust/dialog-s3-credentials/src/ucan/credentials.rs
```diff
@@ -279,4 +297,259 @@ pub mod tests {
...
+            assert_eq!(invocation.audience(), &subject_did);
+            assert_eq!(invocation.subject(), &subject_did);
+
+            invocation
+                .verify_signature()
+                .expect("Signature verification failed");
+        }
+
+        #[dialog_common::test]
+        async fn it_serializes_webcrypto_invocation() {
```
Gozala: Same as above

> Nit: Good test although it does feel out of place here as ti test ucan library and `WebCryptoEd25519Signer` compatibility nothing about credentials.

rust/dialog-s3-credentials/src/ucan/credentials.rs
jackddouglas: it's not about the web API, it's because the browser is single threaded, so wasm32-unknown-unknown needs `?Send`. someone can correct if i'm wrong though

PR #146, opened 2026-01-23
feat: UCAN-based capability system for S3 authorization

Gozala:
##  Summary

This PR adds capability-based system with pluggable authorization and adds UCAN based authorization system (gated by `ucan` feature) for S3 storage operations authorized through remote access service. The changes restructure how credentials and authorization in the S3 backend and how operattor keypair is being threaded through the layers.

### New Capability System (dialog-common)

- Added a hierarchical capability system with Subject, Policy,
  Attenuation, and Effect traits
- Capabilities form chains from a root subject (DID) through policy constraints to executable effects
- Struct Constrained, Claim, Authorization, Invocation, Delegation that offer convenient interface
- Added bytes.rs for opting into serialization / deserialization as bytes

### S3 Credentials Restructure (dialog-s3-credentials)

  - Removed: Old monolithic access.rs module
  - Added: Modular capability structure:
    - capability/archive.rs - Archive storage capabilities (Get, Put)
    - capability/memory.rs - Transactional memory capabilities (Resolve,
  Publish, Retract)
    - capability/storage.rs - Basic  key value storage capabilities (Get, Set, Delete)
  - Added: Separate s3/ and ucan/ modules for different authorization flows:
    - s3/credentials.rs - Direct S3 presigned URL generation
    - s3/provider.rs - S3 capability provider implementation
    - ucan/credentials.rs - UCAN-based credential management
    - ucan/provider.rs - UCAN authorization provider with delegation support
    - ucan/invocation.rs - UCAN invocation container handling
    - ucan/delegation.rs - Delegation chain verification

Note: ucan/provier is designed to be used by access service

### Storage Backend Updates (dialog-storage)

  - Updated: S3 backend to use new credential/authorizer system
  - Added: Capability types mirroring dialog-s3-credentials structure
  - Added: helpers/ucan_server.rs - Test UCAN assess service
  - Removed: Old request.rs module, ucan_e2e_test.rs example

### Artifacts & Replica Updates (dialog-artifacts)

- Refactored: replica/remote.rs - Redesigned and refactored remote site/branch management
- Refactored: replica/operator.rs - Signing identity management
- Added: replica/repository.rs - Repository trait for remotes
- Updated: platform.rs with PrefixedBackend wrapper for key namespacing to avoid leaking key names spacing details into remotes backends

## Left out

My hope is we can migrate to capability based system which would allow removing whole bunch of code, but I hope we can do it in followup workstreams


.github/workflows/test.yml
```diff
@@ -86,24 +86,24 @@ jobs:
...
+    needs: ["lints"]
+    runs-on: "ubuntu-latest"
     strategy:
       matrix:
         include:
-          - flags: '--features s3,s3-list,integration-tests'
-          - flags: '--features s3,s3-list,integration-tests --release'
-          - flags: '--features s3,s3-list --target wasm32-unknown-unknown'
-          - flags: '--features s3,s3-list --target wasm32-unknown-unknown --release'
+          - flags: "--features s3,s3-list,ucan,integration-tests"
```
cdata, line 95: Note: now that the feature list has reached significant size, we need to start testing variants with the features disabled (otherwise we won't know if the code works as intended when the features are _not_ enabled).

rust/dialog-capability/src/lib.rs
```diff
@@ -0,0 +1,102 @@
+//! Capability-based authorization system.
```
cdata, line 1: This needs to be its own crate e.g., `dialog-capability`. The spirit of `dialog-common` is small footprint, low-dependency utilities that get used everywhere. The `capability` mod has eclipsed the crate several times over and probably isn't relevant in a lot of places.

rust/dialog-common/src/capability.rs
```diff
@@ -0,0 +1,102 @@
...
+//! associations.
+//!
+//! ## Policy, Attenuation, and Effect
+//!
+//! These three traits form a hierarchy where each builds on the previous:
+//!
+//! - **[`Policy`]**: The base constraint type. Contributes only to **parameters**.
+//!   Policies scope and filter access without changing the command identity. They
+//!   define *where* or *on what* the operation applies (e.g., `Catalog { catalog: "index" }`
+//!   restricts to a specific catalog). A policy does NOT add to the command path.
```
cdata: You never define what a command or command path is, which makes this definition list difficult to grok.

rust/dialog-capability/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+//!   attenuations contribute to the **command path**. They define the operation
+//!   namespace and appear in the path (e.g., `Archive` → `/archive`). The command
+//!   path identifies *what* operation is being performed.
+//!
+//! - **[`Effect`]**: Every Effect is also an Attenuation. Effects add to the command
+//!   path like attenuations, but unlike attenuations, effects are **invocable** - they
+//!   can be executed by a [`Provider`] and produce an output (defined by
+//!   `Effect::Output`).
+//!
+//! # Key Types
```
cdata, line 129: An example or two would be very welcome in this explainer. It's difficult to get a sense of which "key types" are actually relevant when you have no idea how the code is used in practice.

rust/dialog-common/src/bytes.rs
```diff
@@ -0,0 +1,197 @@
...
+/// use dialog_common::Bytes;
+///
+/// let bytes = Bytes::from(vec![1, 2, 3]);
+/// assert_eq!(bytes.as_slice(), &[1, 2, 3]);
+///
+/// // Convert back to Vec<u8>
+/// let vec: Vec<u8> = bytes.into();
+/// ```
+#[derive(Debug, Clone, Default, PartialEq, Eq, Hash)]
+pub struct Bytes(Vec<u8>);
```
cdata, line 31: I gave you this feedback already, so I'm curious how you netted out here: why not create a generic wrapper e.g.,

```rs
pub struct Bytes<T>(T) where T: AsRef<[u8]> + Debug + Clone;
```

Then you can use `Bytes` with anything that may embody an array bytes e.g., `bytes::Bytes`, `[u8; N]` and many, many other cases besides.

It could spare you from having to create manual `Serialize`/`Deserialize` implementations for `Blake3Hash`.

Although honestly I'm surprised this is even a problem. Surely the least efficient option is not the default when serializing byte arrays to CBOR... and if it is, maybe we should be ditching CBOR.

cdata: After reading about this some more, it may be best for us to make use of the [`#[serde(with = "serde_bytes")]`](https://docs.rs/serde_bytes/latest/serde_bytes/index.html) approach.

Yes, it's somewhat noisy, but it beats this approach in that it does not intrude on the type system and can be applied only where it is needed (e.g., only on paths where we care about serializing with `serde`, which are arguably few and far between).

rust/dialog-common/src/hash.rs
```diff
@@ -43,4 +49,87 @@ impl Blake3Hash {
...
+    type Error = Vec<u8>;
+
+    fn try_from(bytes: Vec<u8>) -> Result<Self, Self::Error> {
+        let arr: [u8; 32] = bytes.try_into()?;
+        Ok(Self(arr))
+    }
+}
+
+/// We want to serialize hashes as byte
+impl Serialize for Blake3Hash {
```
cdata, line 75: See comment on `bytes.rs`, but TL;DR it would be better if we didn't have to manually implement `Serialize`/`Deserialize`.

rust/dialog-common/src/lib.rs
```diff
@@ -10,6 +10,14 @@ extern crate self as dialog_common;
 mod sync;
 pub use sync::*;
 
+pub mod bytes;
+/// Capability-based authorization system with hierarchical policies.
+pub mod capability;
+pub use bytes::Bytes;
```
cdata, line 14: nit: prefer publishing every public thing in the crate e.g., `pub use bytes::*`. That way the public API is immediately obvious when looking at the module.

rust/dialog-common/src/capability.rs
```diff
@@ -0,0 +1,102 @@
...
+mod interface;
+mod invocation;
+mod policy;
+pub mod provider;
+mod selector;
+mod settings;
+mod subject;
+
+// Re-exports
+pub use ability::Ability;
```
cdata: Same comment as https://github.com/dialog-db/dialog-db/pull/146/changes#r2726035467

rust/dialog-s3-credentials/src/ucan/authorization.rs
```diff
@@ -0,0 +1,367 @@
...
+        .collect()
+}
+
+/// UCAN-based authorization proof for a capability.
+///
+/// This enum represents authorization in two forms:
+/// - `Owned`: The subject is the same as the audience (self-authorization)
+/// - `Delegated`: Authority is proven through a UCAN delegation chain
+#[derive(Debug, Clone)]
+pub enum UcanAuthorization {
```
jackddouglas, line 55: nit: add `#[allow(clippy::large_enum_variant)]` here too or we could `Box` `InvocationChain`

rust/dialog-common/src/capability/authority.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// Get this principal's DID.
+    fn did(&self) -> &Did;
+}
+
+/// An authority that can sign data.
+///
+/// Extends `Principal` with the ability to sign payloads.
+pub trait Authority: Principal {
+    /// Sign the given payload.
+    fn sign(&mut self, payload: &[u8]) -> Vec<u8>;
```
jackddouglas: noting we'll need an async version of this for web crypto with extractable keys

rust/dialog-s3-credentials/src/ucan/authorization.rs
```diff
@@ -0,0 +1,367 @@
...
+                .split('/')
+                .map(|s| s.to_string())
+                .collect();
+
+            let args = parameters_to_args(self.parameters().clone());
+
+            let key = SigningKey::from_bytes(&authority.secret_key_bytes().ok_or(
+                AuthorizationError::Serialization("Authority key can not be used".into()),
+            )?);
+            let issuer = Ed25519Signer::from(key);
```
jackddouglas, line 256: and support for async ucan authorization

jackddouglas: i wonder if a `Signer` abstraction over sync and async would be useful

jackddouglas: maybe something like:

```rust
pub trait Signer {
    fn sign(&self, message: &[u8]) -> impl Future<Output = Result<Signature, SignError>> + Send;
}
```

where synchronous `Signer::sign` returns `Ready` immediately, and async signers return real future.

then this could be passed to `Authorization::invoke`

Gozala: We discussed this on the call today and I believe agreed we can make async-ness in a followup change. I mostly avoided because ucan-rs was not setup for it and I wanted to reduce to make in already huge pr.

rust/dialog-artifacts/src/replica/operator.rs
```diff
@@ -0,0 +1,92 @@
+pub use super::Replica;
+use super::principal::Principal;
+use super::{
+    Formatter, PlatformBackend, ReplicaError, SECRET_KEY_LENGTH, Signature, SignerMut, SigningKey,
+};
+use dialog_common::Authority;
+pub use dialog_common::capability::Did;
+use dialog_common::capability::Principal as PrincipalTrait;
+
+/// Represents a principal operating a replica.
```
cdata: A little more detail to explain how this is an abstraction over a keypair, plus how this fits into the overall authorization scheme, would be welcome. If you wrote about it elsewhere, link to it.

rust/dialog-artifacts/src/replica/operator.rs
```diff
@@ -0,0 +1,92 @@
...
+        Ok(Self::new(SigningKey::generate(&mut rand::thread_rng())))
+    }
+
+    /// Signs a payload with this issuer's signing key.
+    pub fn sign(&mut self, payload: &[u8]) -> Signature {
+        self.key.sign(payload)
+    }
+
+    /// Returns the DID (Decentralized Identifier) for this issuer.
+    pub fn did(&self) -> &str {
```
cdata: It's a little confusing that the internal field is `id` and the public accessor is `did`. We go from `Principal::did` to `id` only to expose it as `Operator::did`. Let's just call it a `did`?

cdata: Also: in Noosphere we realized late that it is useful to indicate with a newtype when a `String` is actually a `Did`. It would be nice to avoid stringly types where possible.

rust/dialog-artifacts/src/replica/operator.rs
```diff
@@ -0,0 +1,92 @@
...
+        &self.principal
+    }
+
+    /// Returns the raw secret key bytes.
+    pub fn secret_key_bytes(&self) -> [u8; SECRET_KEY_LENGTH] {
+        self.key.to_bytes()
+    }
+
+    /// Opens a replica with this operator as the issuer.
+    pub fn open<Backend: PlatformBackend + 'static>(
```
cdata, line 72: It's confusing to have this inversion of control in here. It may be convenient, but it begs the question: which is the correct way to open the Replica?

1 way to do it is better than 2 ways that are effectively the same IMO.

rust/dialog-artifacts/src/replica/principal.rs
```diff
@@ -0,0 +1,42 @@
...
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct Principal(pub(super) [u8; 32]);
+impl Principal {
+    /// Formats principal as did:key using proper multicodec encoding.
+    ///
+    /// The did:key format for Ed25519 public keys is:
+    /// 1. Prepend multicodec prefix `[0xed, 0x01]` to the 32-byte public key
+    /// 2. Base58btc encode the 34-byte result
+    /// 3. Prepend "z" (base58btc identifier) and wrap in "did:key:"
+    pub fn did(&self) -> String {
```
cdata: Might be good to add a TODO to memoize this in the future.

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+/// A named remote site identifier.
+pub type Site = String;
+
+/// Represents a configured remote site with its credentials.
+///
+/// This is the persisted state for a remote, storing the site name
+/// and the credentials needed to connect to it.
+pub struct RemoteSite<Backend: PlatformBackend> {
+    /// The site name.
+    pub name: Site,
```
cdata: nit: either all fields public or none; this form is confusing because it suggests you should be able to destructure or initialize the struct directly but you cannot; use an accessor instead.

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+    /// Memory cell storing the remote state.
+    memory: TypedStoreResource<RemoteState, Backend>,
+    /// Storage for persistence (cloned, cheap).
+    storage: PlatformStorage<Backend>,
+    /// Issuer for signing requests.
+    issuer: Operator,
+    /// Subject DID for this replica.
+    #[allow(dead_code)]
```
cdata: nit: Please add a `reason = "..."` clause here.

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+use serde::{Deserialize, Serialize};
+
+use super::{Operator, PlatformBackend, RemoteBackend, RemoteState, Revision};
+use crate::replica::ReplicaError;
+use crate::{ErrorMappingBackend, PlatformStorage, TypedStoreResource};
+
+/// A named remote site identifier.
+pub type Site = String;
+
+/// Represents a configured remote site with its credentials.
```
cdata: What is a remote site? Can you go into a little more detail for clarity or link to an expanded explanation?

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+    /// Issuer for signing requests.
+    issuer: Operator,
+    /// Subject DID for this replica.
+    #[allow(dead_code)]
+    subject: Did,
+}
+
+impl<Backend: PlatformBackend + 'static> RemoteSite<Backend> {
+    /// Add a new remote site, persisting its state to storage.
+    pub async fn add(
```
cdata: nit: Something more appropriate to convey the effects of this method would be better e.g., `retrieve_or_add`

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+        Ok(Self {
+            name: state.site,
+            memory,
+            storage,
+            issuer,
+            subject,
+        })
+    }
+
+    /// Load an existing remote site from storage.
```
cdata: It's unclear from the docs why one may prefer to `load` or `add`, and reading the code makes `load` almost seem redundant.

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+        }
+    }
+
+    /// Get the remote state.
+    pub fn state(&self) -> Option<RemoteState> {
+        self.memory.read()
+    }
+
+    /// Connect to the remote S3 storage.
+    #[cfg(feature = "s3")]
```
cdata: It's weird that this public method doesn't exist unless `s3` is enabled. It smells like a leaky abstraction.

rust/dialog-common/src/capability/subject.rs
```diff
@@ -0,0 +1,43 @@
...
+//!
+//! The `Subject` represents the resource owner (identified by a DID) and serves
+//! as the starting point for building capability chains.
+
+use super::constrained::Constrained;
+use super::interface::Capability;
+use super::policy::Policy;
+
+/// A DID (Decentralized Identifier).
+pub type Did = String;
```
cdata: Strongly consider making this a newtype. See also:

- https://github.com/dialog-db/dialog-db/pull/146/changes#r2733152193
- https://github.com/subconsciousnetwork/noosphere/blob/94b0c5e8a91437e1486e02e29bc1aba1b7b225a8/rust/noosphere-core/src/data/strings.rs#L121-L131

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+    }
+}
+
+/// A reference to a repository on a remote site.
+///
+/// This is a builder step for accessing remote branches.
+#[derive(Clone)]
+pub struct RemoteRepository<Backend: PlatformBackend> {
+    /// The subject DID identifying the repository owner.
+    pub subject: Did,
```
cdata: Ditto https://github.com/dialog-db/dialog-db/pull/146/changes#r2733208957

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+            site_name: self.name.clone(),
+            subject: subject.into(),
+            storage: self.storage.clone(),
+            issuer: self.issuer.clone(),
+            state: self.memory.read(),
+        }
+    }
+}
+
+/// A reference to a repository on a remote site.
```
cdata: Probably need a centralized doc with a detailed explainer of the domain that you can link to from these comments. In isolation, they read like, "A blorp that ryzzles the flim."

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
+//! Remote repository credentials and configuration.
```
cdata, line 1: This module probably should be broken out into smaller ones, one for each struct.

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+        subject: Did,
+        /// Storage for persistence (cloned, cheap).
+        storage: PlatformStorage<Backend>,
+        /// Issuer for signing requests.
+        issuer: Operator,
+        /// The remote state (credentials).
+        state: Option<RemoteState>,
+    },
+    /// An open connection to a remote branch.
+    #[cfg(feature = "s3")]
```
cdata: Similar to https://github.com/dialog-db/dialog-db/pull/146/changes#r2733236667 it seems like a leaky abstraction to me that the state can only exist in code if `s3` is enabled. To the extent that this represents the state of some local thing, the detail that it is possible to reach that state or not should probably be enforced elsewhere. It's also just weird to have an enum with only one variant.

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+        match self {
+            Self::Reference { .. } => None,
+            #[cfg(feature = "s3")]
+            Self::Open { down, .. } => down.read(),
+        }
+    }
+
+    /// Get the connection to the remote memory storage (only available when open).
+    /// This is used for storing branch revision state.
+    #[cfg(feature = "s3")]
```
cdata: Ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733236667

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+    pub fn connection(&self) -> Option<PlatformStorage<RemoteBackend>> {
+        match self {
+            Self::Reference { .. } => None,
+            Self::Open { connection, .. } => Some(connection.clone()),
+        }
+    }
+
+    /// Get the connection to the remote archive/index storage (only available when open).
+    /// This is used by the Archive for storing and retrieving tree blocks.
+    #[cfg(feature = "s3")]
```
cdata: Ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733236667

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+                CborEncoder,
+            )),
+        }
+    }
+
+    /// Open a connection to the remote branch and return a mutable reference for chaining.
+    ///
+    /// If already open, this is a no-op. If in Reference state, opens the connection
+    /// and transitions to Open state. On error, self remains unchanged.
+    #[cfg(feature = "s3")]
```
cdata: Ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733236667

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+        }
+
+        // Clone and try to open - if successful, replace self
+        // This avoids unsafe code and keeps self valid on error
+        let opened = self.clone().into_open().await?;
+        *self = opened;
+        Ok(self)
+    }
+
+    /// Internal: consume self and return Open variant or error.
```
cdata: nit: Internal seems redundant here (API is not public)

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+    /// and transitions to Open state. On error, self remains unchanged.
+    #[cfg(feature = "s3")]
+    pub async fn open(&mut self) -> Result<&mut Self, ReplicaError> {
+        // If already open, nothing to do
+        if matches!(self, Self::Open { .. }) {
+            return Ok(self);
+        }
+
+        // Clone and try to open - if successful, replace self
+        // This avoids unsafe code and keeps self valid on error
```
cdata: Thanks for the helpful comment

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+                // Connect to remote
+                let s3 = S3::new(remote_state.credentials.clone(), issuer.clone());
+                let memory_bucket = Bucket::new(s3.clone(), &subject, "memory");
+                let mut connection =
+                    PlatformStorage::new(ErrorMappingBackend::new(memory_bucket), CborEncoder);
+                let index = Bucket::new(s3, &subject, "archive/index");
+
+                // Open remote revision storage
+                let up = connection
+                    .open::<Revision>(&format!("local/{}", &name).into_bytes())
```
cdata: It's a bit worrisome that this is effectively a conventional path (as it is treated here) but probably will cause problems if someone chooses a different convention. It would probably be better to abstract the construction of these paths so that we ensure we are doing it the same way everywhere.

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+    }
+
+    /// Mount the transactional memory for a remote branch from local storage.
+    async fn mount_local(
+        site: &Site,
+        subject: &Did,
+        branch: &str,
+        storage: &mut PlatformStorage<Backend>,
+    ) -> Result<TypedStoreResource<Revision, Backend>, ReplicaError> {
+        let address = format!("remote/{}/{}/{}", site, subject, branch);
```
cdata: Ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733295700

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+        storage
+            .open::<Revision>(&address.into_bytes())
+            .await
+            .map_err(|e| ReplicaError::StorageError(format!("{:?}", e)))
+    }
+
+    /// Resolves remote revision for this branch. If remote revision is different
+    /// from local revision, updates local one to match the remote. Returns
+    /// revision of this branch.
+    #[cfg(feature = "s3")]
```
cdata: Ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733236667

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+    pub async fn resolve(&mut self) -> Result<Option<Revision>, ReplicaError> {
```
cdata: `update` or `sync` seems like it would be more appropriate.

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+                Ok(down.read())
+            }
+            Self::Reference { .. } => Err(ReplicaError::InvalidState {
+                message: "Branch should be open after successful open() call".to_string(),
+            }),
+        }
+    }
+
+    /// Publishes new canonical revision. Returns error if publishing fails.
+    #[cfg(feature = "s3")]
```
cdata: Ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733236667

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+        match self {
+            Self::Open {
+                storage,
+                connection,
+                down,
+                up,
+                ..
+            } => {
+                // Force reload from storage to ensure we get fresh data
+                let _ = up.reload(connection).await;
```
cdata: It seems bad that we are just swallowing the failure case here.

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+            }),
+        }
+    }
+
+    /// Uploads novel nodes from a stream into remote storage.
+    ///
+    /// This method takes a stream of tree nodes (typically from `TreeDifference::novel_nodes()`)
+    /// and pushes them concurrently to the remote storage. Use this before publishing a new
+    /// revision to ensure all tree blocks are available on the remote.
+    #[cfg(feature = "s3")]
```
cdata: Ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733236667

rust/dialog-artifacts/src/replica/remote.rs
cdata: After going through and reading a lot of the usage of this, one basic way to improve this is to move the feature gate inside of the method and have the method return an `Err` automatically in the case that remote connectivity is not possible.

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+                Ok(())
+            }
+            Self::Reference { .. } => Err(ReplicaError::InvalidState {
+                message: "Branch should be open after successful open() call".to_string(),
+            }),
+        }
+    }
+
+    /// Create a new RemoteBranch from site, branch and storage (for Upstream::open).
+    #[cfg(feature = "s3")]
```
cdata: Ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733236667

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+    {
+        use futures_util::pin_mut;
+
+        // Ensure we're open
+        self.open().await?;
+
+        match self {
+            Self::Open { index, .. } => {
+                let mut queue = TaskQueue::default();
+                pin_mut!(nodes);
```
cdata: nit: I mildly prefer `tokio::pin` and I believe it covers the same use cases (although I'd be interested to hear if it doesn't).

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+                // Wait for all uploads to complete
+                queue
+                    .join()
+                    .await
+                    .map_err(|e| ReplicaError::StorageError(format!("Upload failed: {:?}", e)))?;
+
+                Ok(())
+            }
+            Self::Reference { .. } => Err(ReplicaError::InvalidState {
+                message: "Branch should be open after successful open() call".to_string(),
```
cdata: How confident are you in your claim?

https://doc.rust-lang.org/std/macro.unreachable.html

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+    /// Direct S3 access with optional signing credentials.
+    S3(s3::Credentials),
+    /// UCAN-based access via an authorization service.
+    #[cfg(feature = "ucan")]
+    Ucan(ucan::Credentials),
+}
+
+impl RemoteCredentials {
+    /// Create S3 credentials for public access.
+    pub fn s3_public(
```
cdata: It's weird to have these implementation-detail-specific methods on public API. I think I would prefer a struct that wraps `RemoteCredentials` and brings these along.

rust/dialog-artifacts/src/replica/remote.rs
```diff
@@ -0,0 +1,604 @@
...
+    }
+}
+
+/// Credentials for connecting to a remote repository.
+///
+/// This enum stores the credentials configuration that can be persisted.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum RemoteCredentials {
+    /// Direct S3 access with optional signing credentials.
+    S3(s3::Credentials),
```
cdata: Why not gate this on `feature = "s3"`?

rust/dialog-artifacts/src/replica/repository.rs
```diff
@@ -0,0 +1,43 @@
+pub use super::Replica;
+use super::remote::Site;
+use super::{PlatformBackend, RemoteSite, RemoteState, ReplicaError};
+
+/// Manages remote repositories for synchronization.
```
cdata: Who should implement this trait?

rust/dialog-artifacts/src/platform.rs
```diff
@@ -146,6 +146,53 @@ where
...
+    pub fn new(prefix: impl Into<Vec<u8>>, backend: Backend) -> Self {
+        Self {
+            prefix: prefix.into(),
+            backend,
+        }
+    }
+
+    /// Prepends the prefix to a key.
+    fn prefixed_key(&self, key: &[u8]) -> Vec<u8> {
+        let mut prefixed = self.prefix.clone();
```
cdata, line 168: It sucks to have to do this allocation repeatedly for every storage operation. Assuming keys are fixed length (which I think is the case?) you could pre-allocate a buffer of appropriate length and copy `key` into it instead.

rust/dialog-artifacts/src/platform.rs
```diff
@@ -146,6 +146,53 @@ where
     }
 }
 
+/// A storage backend wrapper that prepends a prefix to all keys.
+/// This allows namespacing storage without modifying the underlying backend.
+#[derive(Clone, Debug)]
+pub struct PrefixedBackend<Backend> {
```
cdata, line 152: Is this tested anywhere?

rust/dialog-artifacts/src/replica.rs
```diff
@@ -105,86 +96,29 @@ impl From<NodeReference> for Blake3Hash {
...
 pub struct Replica<Backend: PlatformBackend> {
-    issuer: Issuer,
+    issuer: Operator,
+    subject: Did,
     #[allow(dead_code)]
     storage: PlatformStorage<Backend>,
-    /// Remote repositories for synchronization
-    pub remotes: Remotes<Backend>,
+    // /// Remote repositories for synchronization
+    // pub remotes: Remotes<Backend>,
```
cdata: Remove?

rust/dialog-artifacts/src/replica.rs
```diff
@@ -105,86 +96,29 @@ impl From<NodeReference> for Blake3Hash {
...
-    /// Returns the principal (public key bytes) for this issuer.
-    pub fn principal(&self) -> &Principal {
-        &self.principal
-    }
-}
-
 /// A replica represents a local instance of a distributed database.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Replica<Backend: PlatformBackend> {
```
cdata, line 98: I dunno when it will be appropriate to address this, but I want to keep fresh in our minds that we discussed calling this something else e.g., `Database`.

rust/dialog-artifacts/src/replica.rs
```diff
@@ -193,49 +127,114 @@ impl<Backend: PlatformBackend + 'static> Replica<Backend> {
...
 /// demand replication. Uses Arc to share remote state across clones.
-#[derive(Clone, Debug)]
-pub struct Archive<Backend: PlatformBackend> {
-    local: Arc<PlatformStorage<Backend>>,
+///
+/// Archive is storage-agnostic and uses whatever backend you give it.
+/// If you want prefixed keys, wrap your backend with PrefixedBackend before
+/// passing it to Archive.
+#[derive(Clone)]
+pub struct Archive<Backend>
```
cdata, line 198: I would prefer a name that is more literally descriptive of what this is. My read is that it is a sort of sparse mirror of two storage backends (one local, and one remote). If I exercise some imagination, archive says "cold storage" but doesn't say anything about the relationship between two storage backends.

.github/workflows/test.yml
Gozala: Are you suggesting to use to do separate set for all of these one with UCANs and other without ?

rust/dialog-artifacts/src/replica.rs
```diff
@@ -1748,28 +1349,32 @@ impl<Backend: PlatformBackend + 'static> Upstream<Backend> {
...
                     Ok(Upstream::Remote(remote_branch))
                 }
             }
         })
     }
 
-    /// Returns the branch id of this upstream
-    pub fn id(&self) -> &BranchId {
+    /// Returns the branch id of this upstream as a string
+    pub fn id(&self) -> String {
```
cdata, line 1343: Ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733152193

rust/dialog-artifacts/src/replica.rs
```diff
@@ -1783,7 +1388,7 @@ impl<Backend: PlatformBackend + 'static> Upstream<Backend> {
 
     /// Returns site of the branch. If local returns None otherwise
     /// returns site identifier
-    pub fn site(&self) -> Option<&Site> {
+    pub fn site(&self) -> Option<&str> {
```
cdata, line 1360: Ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733152193

rust/dialog-common/src/capability/ability.rs
```diff
@@ -0,0 +1,76 @@
+//! Ability trait for capability chains.
```
cdata: nit: Module documentation is actually the trait documentation?

rust/dialog-capability/src/ability.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// Subject of this capability in `did:key` format which is both a resource
+    /// and a root issuer of this capability.
+    fn subject(&self) -> &Did;
+
+    /// Command path representing a level of access this capability
+    /// has over the `subject` resource.
+    fn command(&self) -> String;
+
+    /// Collects parameters into given settings
+    #[cfg(feature = "ucan")]
```
cdata, line 20: It is a sign of leaky abstraction to change public APIs based on implementation details. Please reconsider how this is factored.

rust/dialog-common/src/capability/ability.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/// Subject represents unconstrained capability, hence
+/// the `/` for command.
+impl Ability for Subject {
+    fn subject(&self) -> &Did {
+        &self.0
+    }
+
+    fn command(&self) -> String {
```
cdata: This seems to apply https://github.com/dialog-db/dialog-db/pull/146/changes#r2733152193

rust/dialog-artifacts/src/replica/operator.rs
Gozala: Would making `Replica::open` crate level private address this ? Or do you want me to remove `Replica::open` entirely ?

rust/dialog-common/src/capability/access.rs
```diff
@@ -0,0 +1,33 @@
+//! Access trait for managing access to capability authorizations.
```
cdata: Ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733579754

Noting: if you re-export everything up to the parent module you do not need to add module-level documentation to every micro-module. In fact, I prefer if you do re-export the contents the module as high-up as possible.

rust/dialog-common/src/capability/attenuation.rs
```diff
@@ -0,0 +1,39 @@
+//! Attenuation trait for command path segments.
```
cdata: Ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733605568

rust/dialog-common/src/capability/attenuation.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use super::constraint::Constraint;
+use super::effect::Effect;
+use super::policy::Policy;
+use super::settings::Settings;
```
cdata: See? These `use` paths are pretty redundant. They could just be `use super::Thing` rather than `use super::thing::Thing`. Then, the module-level documentation at `super` can address all its re-exports together as a whole rather than be a redundant explanation of each small piece.

rust/dialog-common/src/capability/attenuation.rs
```diff
@@ -0,0 +1,39 @@
...
+
+/// Marker trait for policies that also constrain a command path.
+///
+/// Attenuation implies `Policy` via blanket impl. The `attenuation()` method
```
cdata: nit:
```
[`Policy`]
```

rust/dialog-common/src/capability/attenuation.rs
```diff
@@ -0,0 +1,39 @@
...
+/// provides the path segment for the command path.
+///
+/// Note: `Effect` types automatically implement `Attenuation` via blanket impl.
```
cdata: Same, link the thingies!

rust/dialog-capability/src/attenuation.rs
```diff
@@ -0,0 +1,39 @@
...
+/// Note: `Effect` types automatically implement `Attenuation` via blanket impl.
+pub trait Attenuation: Sized + Settings {
+    /// The capability this type constrains.
+    /// Must implement `Constraint` so the blanket `Policy` impl works.
+    type Of: Constraint;
+
+    /// Get the attenuation segment for this type.
+    /// Attenuation types contribute to the command path.
+    fn attenuation() -> &'static str {
+        let full = std::any::type_name::<Self>();
```
cdata, line 44: Can you comment in the documentation about the implications of the type domain on the actual act of attenuating? It seems like I have to care / be careful about how I name things.

rust/dialog-common/src/capability/authority.rs
cdata: Seconded; why not make this pre-emptively async?

rust/dialog-capability/src/authorization.rs
```diff
@@ -0,0 +1,68 @@
+//! Authorization trait for capability-based access control.
+//!
+//! The `Authorization` trait represents proof of authority over a capability.
+
+use super::{Authority, subject::Did};
+
+/// Errors that can occur during authorization.
+#[derive(Debug, thiserror::Error)]
+pub enum AuthorizationError {
```
cdata, line 9: nit: I prefer that errors live in a distinctly named `error.rs` module.

Separately: up until now we have namespaced errors due to the high likelihood that there will be name collisions with other crates e.g., `DialogCapabilityAuthorizationError`. I'm not married to the convention but I would like there to be a consistent convention.

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Well this only adds it does not retrieve or add. If remote with the conflicting name already exists it will error, except for the case when what is being added is exactly what's already persisted, in that case it's a noop to make API idempotent.

rust/dialog-common/src/capability/authorization.rs
```diff
@@ -0,0 +1,68 @@
...
+/// - Derived from a delegation chain
+pub trait Authorization: Sized {
+    /// The subject (resource owner) this authorization covers.
+    fn subject(&self) -> &Did;
+
+    /// The audience who has been granted authority.
+    fn audience(&self) -> &Did;
+
+    /// The command path (ability) this authorization permits.
+    fn can(&self) -> &str;
```
cdata: ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733152193

rust/dialog-common/src/capability/authorized.rs
```diff
@@ -0,0 +1,115 @@
...
+use super::constraint::Constraint;
+use super::interface::Capability;
+use super::provider::Provider;
+use super::{Authority, Authorization, AuthorizationError};
+use std::error::Error;
+
+/// A capability paired with its authorization proof.
+///
+/// `Authorized` bundles a capability with proof that the invoker has
+/// permission to execute it. This is the input to authorized `Provider`
```
cdata: nit: linkify "`Provider`"

rust/dialog-capability/src/authorized.rs
```diff
@@ -0,0 +1,115 @@
...
+use std::error::Error;
+
+/// A capability paired with its authorization proof.
+///
+/// `Authorized` bundles a capability with proof that the invoker has
+/// permission to execute it. This is the input to authorized `Provider`
+/// implementations.
+///
+/// - `C` is the constraint type (e.g., `storage::Get`)
+/// - `A` is the authorization type (e.g., `UcanAuthorization`)
```
cdata, line 11: Link these!

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Do you have a suggestion what to do in that case ? Because we don't have Remote's other than S3 at the moment ? Should I just create a connect that fails ? Or should I feature flag all of the remote functionality under s3 flag ?

rust/dialog-common/src/capability/constrained.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// Extract a policy or ability from this chain.
+    pub fn policy<T, Index>(&self) -> &T
+    where
+        Self: Selector<T, Index>,
+    {
+        self.select()
+    }
+
+    /// Get the subject DID from the capability chain.
+    pub fn subject(&self) -> &super::subject::Did {
```
cdata: nit: `use super::subject::Did` above.

rust/dialog-capability/src/constrained.rs
```diff
@@ -0,0 +1,134 @@
...
+            constraint: fx,
+            capability: self,
+        }
+    }
+
+    /// Collect all parameters from this capability chain into a new map.
+    ///
+    /// This walks the capability chain and collects parameters from each
+    /// constraint.
+    #[cfg(feature = "ucan")]
```
cdata, line 75: Ditto: https://github.com/dialog-db/dialog-db/pull/146/changes#r2733592176

rust/dialog-common/src/capability/effect.rs
```diff
@@ -0,0 +1,17 @@
...
+
+use super::constraint::Constraint;
+use super::settings::Settings;
+use crate::ConditionalSend;
+
+/// Trait for effect types that can be performed.
+///
+/// Effects are capabilities that can be invoked and therefor require their
+/// output type. Implementing `Effect` automatically makes the type an
+/// `Attenuation` (and thus a `Policy`) via blanket impls.
```
cdata: Nit: linkify `Attenuation` and `Policy`

rust/dialog-common/src/capability/interface.rs
```diff
@@ -0,0 +1,384 @@
+//! Capability wrapper type.
```
cdata: nit: if I was looking for this trait, it would be difficult to find burried in a module called `interface.rs`. It would make more sense in the host `capability.rs` or perhaps in a `capability/mod.rs`.

rust/dialog-common/src/capability/policy.rs
```diff
@@ -0,0 +1,41 @@
...
+use super::selector::Selector;
+use super::settings::Settings;
+
+/// Trait for policy types that restrict capabilities.
+///
+/// `Policy` is for types that represent restrictions on what can be done
+/// with a capability. Implement this for types that don't contribute to
+/// the command path.
+///
+/// For types that contribute to the command path, implement `Attenuation`
```
cdata: nit: linkify `Attenuation`

.github/workflows/test.yml
cdata: I'm just observing we have a growing blindspot in our test coverage here. An example of this coming up is how you didn't realize you were running tests locally without the `integration-tests` feature enabled.

rust/dialog-common/src/bytes.rs
Gozala: Just to be clear you're suggesting dropping `Bytes` all together and annotating fields that need to be serialized into bytes with  `#[serde(with = "serde_bytes")]` ? Double checking because I think that is what I had and you suggested what I thought I did here, although sounds like I did not exactly understood what you said.

.github/workflows/test.yml
Gozala: Not that it is important here, but I forgot to pass it once because I was running it as we were speaking on the call, and figured it out right after.

That said I'm genuinely not sure what the right approach here. Features do introduce whole another dimension to things and I'm not sure what is the best way to manage this added complexity.

rust/dialog-artifacts/src/platform.rs
Gozala: Yes it is used here https://github.com/dialog-db/dialog-db/pull/146/files#diff-fa9e30d02cf21ef593f959fe49be1a3b2856075c152045ed44a10e9703110bffR64 and all the code the tests that do queries touch it.

rust/dialog-artifacts/src/platform.rs
Gozala: If that's ok with you I'd like to postpone this and address as a part of https://github.com/dialog-db/dialog-db/issues/147 because there whole lot of abstractions were introduced in a rush to overcome various limitations (around key sizes) and I would like to remove them all instead of trying to patch up all of them which also span across here and tonk repo.

Gozala: I suggest we do this as a followup cleanup, because there are too many layers affected by it which I would like to replace with capability based solution anyway as per https://github.com/dialog-db/dialog-db/issues/147

rust/dialog-artifacts/src/replica.rs
Gozala: https://github.com/dialog-db/dialog-db/issues/149

rust/dialog-artifacts/src/replica.rs
Gozala: I suggest we tackle this as part of https://github.com/dialog-db/dialog-db/issues/147

rust/dialog-artifacts/src/replica.rs
```diff
@@ -1715,6 +1311,7 @@ impl BranchState {
 /// Upstream branch that is used to push & pull changes
 /// to / from. It can be local or remote.
 #[derive(Debug, Clone)]
+#[allow(clippy::large_enum_variant)]
```
Gozala, line 1283: @cdata I don't know hat is the right solution here, perhaps you can offer an advice. I believe problem is that indirectly `RemoteBranch` can contain UCAN Delegation which can be large.

rust/dialog-artifacts/src/replica/operator.rs
Gozala: For the context, I'd like API to be something along the lines 

```rs
let main = Operator::from_secret(secret)
   .open(repo_did).await?
   .branch()
   .open('main').await?;
```

I was trying to avoid refactoring all of the code to conform this API however, so I think we could either

1. Go and actually refactor
2. Make `Replica::open` pub (crate) until we refactor.
3. Refactor `Replica::open` code into `Operator`.


Let me know which one would you like me to do and I'll act change accordingly

rust/dialog-artifacts/src/replica/remote.rs
Gozala: I mean that is only thing open does it literary transitions form whatever state to Open or fails.  In different language I'd had something like `RemoteBranch { Reference, Open(OpenRemoteBranch) }` and have `open` return `OpenRemoteBranch` to get type system onboard, but I could not work out the way to do it yet retain ownership and all so I resorted to this

Gozala: I think I misunderstood your comment. I'm pretty confident `unreachable!` would be fine here, only reason I did not use it is because `upload` operation can fail anyhow so it was not really affecting consumer API and this seemed like safer choice over time where behavior of `open` could change.

rust/dialog-capability/src/ability.rs
Gozala: I'm not too happy about this either, but also found it very difficult to avoid this without introducing dependency on `Ipld` or defining whole new abstraction that is like `Ipld`.

Perhaps you can advice something better ? If introducing dependency on `Ipld` seems reasonable to you I can simply remove the feature gate.

rust/dialog-capability/src/authorized.rs
Gozala: These don't exist in this create so I don't think that would work would it ?

rust/dialog-common/src/capability/subject.rs
Gozala: Done already https://github.com/dialog-db/dialog-db/pull/148 will merge it back in here once I address everything else.

rust/dialog-common/src/capability/authority.rs
Gozala: Addressed in https://github.com/dialog-db/dialog-db/pull/148/changes/ad8746bb99879bdfa828b033637875679982bbe8

rust/dialog-s3-credentials/src/ucan/authorization.rs
Gozala: Addressed this in https://github.com/dialog-db/dialog-db/pull/148/changes/ad8746bb99879bdfa828b033637875679982bbe8

Gozala: Fixed with [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-capability/src/lib.rs
Gozala: Broke out this into #150 and will merge it here after I'm done with it.

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Refactored all of the remote stuff to avoid these kind of feature flags [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Refactored remote into small modules [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Made large refactor of all the remote.rs to get rid of most feature flags [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Done in [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Done in [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Fixed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/remote.rs
Gozala: addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Addressed by Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/remote.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica/repository.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-artifacts/src/replica.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-common/src/capability/ability.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-common/src/capability/access.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-common/src/capability/attenuation.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-common/src/capability/attenuation.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-common/src/capability/attenuation.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-common/src/capability/attenuation.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-capability/src/attenuation.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

Added documentation covering this and how to provide custom name instead

rust/dialog-common/src/capability/authorized.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-common/src/capability/constrained.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-common/src/capability/effect.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-common/src/capability/interface.rs
Gozala: Will be addressed by https://github.com/dialog-db/dialog-db/pull/150

rust/dialog-common/src/capability/policy.rs
Gozala: Addressed by [7e78cd3](https://github.com/dialog-db/dialog-db/pull/146/commits/7e78cd3eaf636481c763139cc547adac8dc80aeb)

rust/dialog-common/src/bytes.rs
Gozala: Created followup task for this https://github.com/dialog-db/dialog-db/issues/151

PR #153, opened 2026-01-29
feat(artifacts): make `Replica` generic over `Authority`

jackddouglas:
This allows using custom Authority implementations (like WebCryptoAuthority) instead of the concrete Operator type. Key changes:

- Replica<Backend, A> now generic with default A = Operator
- Branches, Branch, Upstream, RemoteBranch, RemoteSite, RemoteRepository also generic
- Added ConditionalSync bound for platform-specific Send/Sync requirements
- Added Principal::from_did() to reconstruct principal from DID string
- Remote S3 operations extract key bytes via Authority::secret_key_bytes()


rust/dialog-artifacts/src/replica.rs
```diff
@@ -94,18 +94,23 @@ impl From<NodeReference> for Blake3Hash {
 }
 
 /// A replica represents a local instance of a distributed database.
+///
+/// The type parameter `A` represents the authority used for signing operations.
+/// It defaults to `Operator` for backward compatibility.
 #[derive(Debug, Clone)]
-pub struct Replica<Backend: PlatformBackend> {
-    issuer: Operator,
+pub struct Replica<Backend: PlatformBackend, A: Authority + Clone + Debug = Operator> {
```
Gozala: Can you please define some trait that contains all the constraints in it e.g. something like 

```rs
trait OperatingAuthority: Authority + Clone + Debug {}
impl <A: Authority + Clone + Debug> OperatingAuthority for A {}
```

So it can be referred everywhere without having to repeat same set of bounds. Above trait def also creates opportunity to write a comment explaining why certain bounds are needed.

rust/dialog-artifacts/src/replica.rs
```diff
@@ -570,21 +588,21 @@ impl<Backend: PlatformBackend + 'static> Branch<Backend> {
...
     fn state(&self) -> BranchState {
         self.memory.read().unwrap_or_else(|| {
-            BranchState::new(
-                self.id.clone(),
-                Revision::new(self.issuer.principal().to_owned()),
-                None,
-            )
+            let principal = Principal::from_did(self.issuer.did())
+                .expect("Authority DID must be valid did:key format");
+            BranchState::new(self.id.clone(), Revision::new(principal), None)
```
Gozala: 🤔 Wonder if we should just use Did in `BranchState` instead.

rust/dialog-artifacts/src/replica.rs
```diff
@@ -745,9 +763,10 @@ impl<Backend: PlatformBackend + 'static> Branch<Backend> {
                         Ok(Some(revision))
                     } else {
                         // Create new revision with integrated changes
-                        #[allow(clippy::clone_on_copy)]
+                        let principal = Principal::from_did(self.issuer.did())
+                            .expect("Authority DID must be valid did:key format");
                         let new_revision = Revision {
-                            issuer: self.issuer.principal().clone(),
+                            issuer: principal,
```
Gozala: Nit: I think just `Did` in the `Revision` be enough, I don't believe it's used for anything other than capturing DID of the issuer anyway.

rust/dialog-artifacts/src/replica/principal.rs
```diff
@@ -24,6 +25,36 @@ impl Principal {
         }
     }
 
+    /// Creates a Principal from a DID string.
+    ///
+    /// Parses `did:key:z...` format to extract the Ed25519 public key bytes.
+    /// Returns `None` if the DID is not a valid Ed25519 did:key.
+    pub fn from_did(did: &Did) -> Option<Self> {
```
Gozala: Nit: Perhaps this should be `impl TryFrom<Did> for Principal` instead ?

Gozala: Please use `Result` instead of `Option` here, I think [roc makes pretty good case for this in general](https://www.roc-lang.org/faq#option-type) but it seems particularly relevant here, because in  reality there could be other DIDs keys like RSA for example not to mention other DID methods. So error saying it does not support did method or did algorithm seems a lot more appropriate.

Gozala (commented):
Providing some feedback. Broadly I'm not sure adding more generics is a direction I personally favor going into. I feel like we have too many already and if anything I'm trying to reduce that. One pattern I find useful personally is something along these lines

https://github.com/dialog-db/dialog-db/blob/e88a5dce82136459771965be80ed82218a5c3d54/rust/dialog-query/src/predicate/concept.rs#L16-L21

Where you can have closed system with an escape hatch to support different use case, here I imagine it could end up looking something like 

```rs
enum SigningAuthority {
   #[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
   NativeEd25519(ed25519_dalek::SigningKey),
   #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
   WebEd25519(WebCryptoEd25519Signer),
   // hatch to bring your own signer at the expense of dynamic dispatch
   Dynamic(dyn Signer)
}

#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
pub trait Signer {
    /// Sign the given payload.
    async fn sign(&mut self, payload: &[u8]) -> Result<Vec<u8>, SignError>;
}

impl Signer for SigningAuthority { ... }
```

This way you get all the best IMO, static that can be feature gated by algorithm / platform  and escape hash for whatever one might desire at expense of dynamic dispatch.

rust/dialog-artifacts/src/replica/remote/branch.rs
```diff
@@ -312,8 +320,19 @@ impl<Backend: PlatformBackend + 'static> RemoteBranch<Backend> {
                 .await
                 .map_err(|e| ReplicaError::StorageError(format!("{:?}", e)))?;
 
-            // Connect to remote using credentials
-            let connection = credentials.connect(desc.issuer.clone(), &desc.subject)?;
+            // Connect to remote using credentials.
+            // Remote S3 operations require a SigningAuthority with secret key access.
+            // Construct one from the Authority's secret key bytes if available.
+            let authority = match desc.issuer.secret_key_bytes() {
```
Gozala: Nit: we need to figure out a way to remove this requirement. I added it in order to overcome inflexibility of ucan-rs, but I suspect with support of web-crypto you've added we may not need that anymore.

rust/dialog-artifacts/src/replica/remote/site.rs
```diff
@@ -101,9 +102,22 @@ impl<Backend: PlatformBackend + 'static> RemoteSite<Backend> {
...
+            // Remote S3 operations require a SigningAuthority with secret key access.
+            let authority = match self.issuer.secret_key_bytes() {
+                Some(bytes) => SigningAuthority::from_secret(&bytes),
+                None => {
+                    return Err(ReplicaError::StorageError(
+                        "Remote operations require an authority with extractable key material"
+                            .to_string(),
+                    ));
+                }
+            };
```
Gozala: Can we do `impl <T: Authority> TryFrom<T> for  SigningAuthority` to encapsulate this in one place

rust/dialog-artifacts/src/replica/principal.rs
```diff
@@ -1,8 +1,34 @@
...
+    /// The multibase prefix is not 'z' (base58btc).
+    #[error("Invalid multibase encoding: expected 'z' (base58btc) prefix")]
+    InvalidMultibase,
+
+    /// The base58 encoding is malformed.
+    #[error("Invalid base58 encoding: {0}")]
+    InvalidBase58(String),
+
+    /// The decoded key has wrong length.
+    #[error("Invalid key length: expected 34 bytes (2 prefix + 32 key), found {0}")]
```
Gozala: Size varies by key type e.g. RSA keys are much longer perhaps this could have `{ expected_size: usize, actual_size: usize, key_type: String }` or something like that ?

rust/dialog-artifacts/src/replica/signing_authority.rs
```diff
@@ -0,0 +1,375 @@
...
+///     println!("DID: {}", authority.did());
+/// }
+/// ```
+#[derive(Clone)]
+pub enum SigningAuthority {
+    /// Native Ed25519 signing using ed25519_dalek (non-WASM platforms).
+    #[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
+    Native {
+        /// The DID derived from the public key.
+        did: Did,
```
Gozala: Give that principal has did() method that returns &Did this seems redundant.

rust/dialog-artifacts/src/replica/signing_authority.rs
```diff
@@ -0,0 +1,375 @@
...
+        /// The ucan WebCrypto signer (owns the non-extractable CryptoKey).
+        signer: WebCryptoEd25519Signer,
+    },
+
+    /// Userspace Ed25519 fallback (WASM, when WebCrypto Ed25519 is unavailable).
+    #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+    Fallback {
+        /// The DID derived from the public key.
+        did: Did,
```
Gozala: Give that principal has did() method that returns &Did this seems redundant.

rust/dialog-artifacts/src/replica/signing_authority.rs
```diff
@@ -0,0 +1,375 @@
...
+    WebCrypto {
+        /// The DID derived from the public key.
+        did: Did,
+        /// The ucan WebCrypto signer (owns the non-extractable CryptoKey).
+        signer: WebCryptoEd25519Signer,
+    },
+
+    /// Userspace Ed25519 fallback (WASM, when WebCrypto Ed25519 is unavailable).
+    #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+    Fallback {
```
Gozala: This variant seems identical to `Native` can we drop one of them ?

rust/dialog-artifacts/src/replica/signing_authority.rs
```diff
@@ -0,0 +1,375 @@
...
+    }
+}
+
+impl PartialEq for SigningAuthority {
+    fn eq(&self, other: &Self) -> bool {
+        self.did() == other.did()
+    }
+}
+
+impl Eq for SigningAuthority {}
```
Gozala, line 107: I think you should be able to derive this

rust/dialog-artifacts/src/replica/signing_authority.rs
```diff
@@ -0,0 +1,375 @@
...
+    /// 2. If WebCrypto fails, fall back to the `Fallback` variant with extractable keys
+    ///
+    /// On non-WASM platforms, this uses the native Ed25519 implementation.
+    /// On WASM platforms without the `webcrypto` feature, this uses the fallback.
+    #[cfg(all(target_arch = "wasm32", target_os = "unknown", feature = "webcrypto"))]
+    pub async fn generate() -> Result<Self, ReplicaError> {
+        match WebCryptoEd25519Signer::generate().await {
+            Ok(signer) => {
+                // Convert Ed25519Did to dialog_capability::Did via string
+                let did = Did::from(signer.did().to_string());
```
Gozala: Nit: Maybe it's worth creating a Principal for consistency with all the other variants ?

rust/dialog-artifacts/src/replica/signing_authority.rs
```diff
@@ -0,0 +1,375 @@
...
+        Self::from_secret(blake3::hash(bytes).as_bytes())
+    }
+
+    /// Creates a new signing authority from a secret key.
+    pub fn from_secret(secret: &[u8; SECRET_KEY_LENGTH]) -> Self {
+        Self::from_signing_key(SigningKey::from_bytes(secret))
+    }
+
+    /// Creates a new signing authority from a signing key.
+    pub fn from_signing_key(key: SigningKey) -> Self {
```
Gozala, line 134: It is possible to turn [signing keys into WebCrypto keys](https://github.com/storacha/ucanto/blob/main/packages/principal/src/rsa.js#L129-L147) _(it is awkward and each key has different format)_. This is not something I suggest to tackle in this pull request but perhaps ew can file an issue and a todo comment here with a pointer to it.

Gozala, line 134: Nit: I'm not yet sure what the `SigningKey` is but it feels like there is some redundancy across `from_signing_key` and `from_secret`.

rust/dialog-artifacts/src/replica/signing_authority.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    /// Creates a signing authority from a WebCrypto signer.
+    ///
+    /// This is used to reconstruct a signing authority from a stored `CryptoKey`.
+    /// The signer should be created via `WebCryptoEd25519Signer::from_key()`.
+    #[cfg(all(target_arch = "wasm32", target_os = "unknown", feature = "webcrypto"))]
+    pub fn from_webcrypto_signer(signer: WebCryptoEd25519Signer) -> Self {
+        let did = Did::from(signer.did().to_string());
+        Self::WebCrypto { did, signer }
+    }
```
Gozala, line 185: Seems like `From<..> for SigningAuthority` might be more idiomatic here

rust/dialog-artifacts/src/replica/signing_authority.rs
```diff
@@ -0,0 +1,375 @@
...
+    /// Returns the WebCrypto signer if this is a WebCrypto signing authority.
+    ///
+    /// This is useful for accessing the underlying `CryptoKey` for storage.
+    #[cfg(all(target_arch = "wasm32", target_os = "unknown", feature = "webcrypto"))]
+    pub fn webcrypto_signer(&self) -> Option<&WebCryptoEd25519Signer> {
+        match self {
+            Self::WebCrypto { signer, .. } => Some(signer),
+            _ => None,
+        }
+    }
```
Gozala, line 202: Nit: I think `TryFrom<SigningAuthority> for WebCryptoEd25519Signer` might be more idiomatic

rust/dialog-artifacts/src/replica/signing_authority.rs
```diff
@@ -0,0 +1,375 @@
...
+                    // For now, we leak a small amount of memory to provide a stable reference.
+                    // TODO: Consider caching the DID in the enum variant.
+                    let did = guard.did().clone();
+                    Box::leak(Box::new(did))
+                } else {
+                    // If we can't get the lock, create a placeholder DID
+                    // This should be extremely rare in practice
+                    static PLACEHOLDER: std::sync::OnceLock<Did> = std::sync::OnceLock::new();
+                    PLACEHOLDER.get_or_init(|| Did::from("did:key:locked"))
+                }
```
Gozala: I would much rather fail on construction than defer this to later use & if we can not support this type of use case we can just leave it out until we need it. I imagine @cdata might have more helpful insight into how this can done properly.

rust/dialog-artifacts/src/replica/signing_authority.rs
```diff
@@ -0,0 +1,375 @@
...
+
+            #[cfg(all(target_arch = "wasm32", target_os = "unknown", feature = "webcrypto"))]
+            Self::WebCrypto { .. } => None, // Principal not directly accessible
+
+            #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+            Self::Fallback { principal, .. } => Some(principal),
+
+            Self::Dynamic(_) => None,
+        }
+    }
```
Gozala: Principal is just a different representation for Did so we should be able to support it everywhere. I think we should either do that or drop this method.

rust/dialog-artifacts/src/replica/signing_authority.rs
```diff
@@ -0,0 +1,375 @@
...
+    /// Returns the raw secret key bytes, if available.
+    ///
+    /// Returns `None` for `WebCrypto` (non-extractable) and `Dynamic` variants.
+    pub fn secret_key_bytes(&self) -> Option<[u8; SECRET_KEY_LENGTH]> {
+        match self {
+            #[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
+            Self::Native { key, .. } => Some(key.to_bytes()),
+
+            #[cfg(all(target_arch = "wasm32", target_os = "unknown", feature = "webcrypto"))]
+            Self::WebCrypto { .. } => None, // Non-extractable!
```
Gozala: Nit: WebCrypto keys can be generates as extractable in which case they would be extractable.

rust/dialog-artifacts/src/replica/signing_authority.rs
```diff
@@ -0,0 +1,375 @@
...
+            Self::Fallback { did, .. } => did,
+
+            Self::Dynamic(signer) => {
+                // Note: This blocks on the lock, which is not ideal but necessary
+                // for the sync did() method. In practice, contention should be rare.
+                // We use try_read to avoid blocking if possible.
+                if let Ok(guard) = signer.try_read() {
+                    // SAFETY: We need to return a reference, but we can't hold the guard.
+                    // This is a limitation of the Dynamic variant's design.
+                    // For now, we leak a small amount of memory to provide a stable reference.
```
cdata: For something the size of a DID it would be better to give an owned copy than leak any memory, no matter how small.

The default allocator has optimizations for repeated allocations of buffers with a predictable size. But, it won't do anything at all to recover compounding leaked memory. You're cloning in this variant anyway, so...

rust/dialog-artifacts/src/replica/signing_authority.rs
cdata: If the locking mechanism is fallible, this accessor method should also be fallible (IMO).

But: not all `RwLock` implementations are fallible in this way. For example: [`parking_lot`](https://docs.rs/parking_lot/latest/parking_lot/type.RwLock.html#method.try_read) returns `Option<...>` (so the accessor could return `Option<Did>`).

That said, stepping back: I'm not sure I buy the rationale for `try_read`. If we take the explanation at face value that contention is rare, then blocking the thread is low-risk and `read` - which is not fallible and always gives you a guard - should suffice (removing the need for this condition altogether).

rust/dialog-artifacts/src/replica/signing_authority.rs
jackddouglas: i don't think so, at least with the way SigningAuthority is structured now. the dyn variant doesn't implement PartialEq and we compare DIDs semantically, not structurally

rust/dialog-artifacts/src/replica/signing_authority.rs
jackddouglas: `from_secret` just wraps `from_signing_key` so you don't have to do `from_signing_key(SigningKey::from_bytes(secret))` directly

rust/dialog-artifacts/src/replica/signing_authority.rs
jackddouglas: i'm not sure because this returns a reference, whereas IIUC `TryFrom` should return Result with owned. though i will probably replace Option here with Result anyway. @cdata do you have thoughts on this?

rust/dialog-artifacts/src/replica/signing_authority.rs
jackddouglas: i think regardless this issue is resolved by storing the `Principal` in the dyn variant and getting the DID from that

rust/dialog-artifacts/src/replica/signing_authority.rs
Gozala: I overlooked reference part, although I think there is the way to deal with refs with From / Into traits.

My general impression from prior feedback is that conditionally available methods is an anti pattern. Suggested `TryFrom` was one way to avoid this but alternative could be to define some trait like `WebSigner` or something that implemented and available under specific features.

However perhaps even better way to think of this would be what is this for, your code comment suggest it's for persisting it in storage perhaps better API would have something like `async fn save(&self, storage: Storage) → Result<...>` that way it can be persisted (assuming there is a way for us to define storage) and format in which it stores becomes variant specific.

rust/dialog-artifacts/src/replica/remote/branch.rs
```diff
@@ -312,8 +320,11 @@ impl<Backend: PlatformBackend + 'static> RemoteBranch<Backend> {
                 .await
                 .map_err(|e| ReplicaError::StorageError(format!("{:?}", e)))?;
 
-            // Connect to remote using credentials
-            let connection = credentials.connect(desc.issuer.clone(), &desc.subject)?;
+            // Connect to remote using credentials.
+            // TODO: Remove this requirement once ucan-rs is more flexible with WebCrypto support,
+            // allowing us to use non-extractable keys for remote operations.
+            let authority = SigningAuthority::try_from_authority(&desc.issuer)?;
```
Gozala, line 326: I'm bit lost in regards to what's going on here, seems like `desc.issuer` is already a `SigningAuthority` isn't it ?

Gozala: Oh I guess it is  `OperatingAuthority` that `SigningAuthority` implements. Can't help but wonder again why have that generic, if we just used   `SigningAuthority` no conversion here would be necessary.

rust/dialog-artifacts/src/replica/signing_authority.rs
Gozala: I was under impression if you implement `PartialEq` which you do above it should be possible to derive `Eq` but could be wrong.

PR #208, opened 2026-02-26
📝 feat: spec out formal JSON and abbrevated YAML notation

Gozala:
This specs out **formal** JSON notation for dialog primitives and companion **abbreviated** notation in YAML

### DIFF
--- a/notes/notation.md
+++ b/notes/notation.md
@@ -16,21 +16,23 @@ An attribute's identity is the tuple `(the, type, cardinality)`. A concept's ide
 
 The `the` component within an attribute is nominal: it carries meaning beyond structure. `diy.cook/quantity` and `diy.cook/price` may both be `(*, Integer, one)` structurally, but they are distinct attributes because `the` denotes the kind of relation they form, which is what makes it part of the identity in the first place.
 
+### Selector
+
+An attribute selector is the combined `domain/name` string. The total length must not exceed **64 bytes**, which is the storage-layer encoding budget.
+
 ### Domain
 
-A domain is a dot-separated reversed domain name, always normalized to lowercase. Domains group related attributes and colocate them in the index, which complements dialog's query-driven replication: all attributes sharing a domain prefix can be fetched together efficiently, making it practical to replicate a coherent slice of information without pulling unrelated data. The reversed domain notation avoids collisions between independently developed schemas.
+A domain groups related attributes. Domains may use dot-separated segments for hierarchical organization, following a reversed domain name convention to avoid collisions between independently developed schemas.
 
 **Rules:**
 
-- Dot-separated segments in reverse domain order
-- All segments lowercase
-- Each segment: 1–63 ASCII alphanumeric characters or hyphens
-- Hyphens not allowed at start or end of a segment
-- Top-level domain must not start with a digit
-- At least two segments required
-- Maximum 253 characters total (including dots)
+- Lowercase ASCII letters, digits, hyphens, and dots
+- Must start with a letter
+- Must not end with a dot or hyphen
+- At least one character
 
 ```
+person
 diy.cook
 io.gozala.person
 org.example.hr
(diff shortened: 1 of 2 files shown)


notes/notation/schema.json
```diff
@@ -0,0 +1,497 @@
...
+      "properties": {
+        "description": {
+          "type": "string",
+          "description": "Human-readable description of the attribute."
+        },
+        "the": {
+          "type": "string",
+          "description": "The relation in domain/name format (e.g. 'diy.cook/quantity'). Max 64 bytes total. Domain: lowercase letters, digits, hyphens, dots; must start with a letter. Name: lowercase kebab-case (no dots); must start with a letter.",
+          "maxLength": 64,
+          "pattern": "^[a-z][a-z0-9.-]*[a-z0-9]?/[a-z][a-z0-9-]*[a-z0-9]?$"
```
jackddouglas, line 20: i ran claude against these regex patterns and it showed the `?` on the final character allows for trailing dots/hyphens in domain and trailing hyphen in name. it suggested this instead which is inline with the spec:
```
^([a-z][a-z0-9.-]*[a-z0-9]|[a-z])/([a-z][a-z0-9-]*[a-z0-9]|[a-z])$
```

notes/notation.md
```diff
@@ -0,0 +1,1713 @@
...
+    },
+    "required": ["value", "is"]
+  }
+}
+</pre>
+</details>
+
+### Assertions and Claims
+
+Tools interact with the associative layer by submitting **assertions** and **retractions**. An assertion proposes that a relation holds; a retraction proposes that it no longer does. Once the transactor incorporates an assertion, it becomes a **claim**, the fundamental unit of information stored in the associative layer.
```
jackddouglas, line 1014: if we want to standardize around `Claim` here, we should probably clear that up more generally since we now have `Fact` `Claim` and `Artifact` floating around in different places

notes/notation.md
```diff
@@ -0,0 +1,1713 @@
...
+
+```yaml
+assert!:
+  the:   issue/assignee
+  of:    did:key:zIssue42
+  is:    did:key:zDana
+  cause:
+    by: did:key:zHome,
+    period: 3,
+    moment: 9
```
jackddouglas, line 1037: this hasn't been implemented yet right?

notes/notation.md
```diff
@@ -0,0 +1,1713 @@
...
+{ 
+  "assert": "math/sum", 
+  "where": { 
+    "of": { "?": { "name": "a" } },
+    "with": { "?": { "name": "b" } }, 
+    "is": { "?": { "name": "result" }
+    } 
+  } 
+}
+```
```
jackddouglas, line 701: the formula syntax could get confusing since they look like attribute selectors. and would these be reserved to prevent conflicts?

notes/notation.md
```diff
@@ -0,0 +1,1713 @@
...
+
+`name: .` expands to `io.gozala.person/name` by inheriting the field name and the concept's domain (`io.gozala/Person` normalizes to `io.gozala.person`).
+
+#### Inline attributes
+
+Attribute definitions can be inlined inside a concept in abbreviated form. The domain is derived by lowercasing the concept label and appending it as an additional segment:
+
+```
+diy.cook/RecipeStep  ->  diy.cook.recipe-step/
```
jackddouglas, line 1363: whats the algo for converting from pascal to kebab case?

jackddouglas (approved):
looks good on the whole, just some more technical questions about how parts of the notation would be implemented

notes/notation.md
Gozala: Fact is gone in https://github.com/dialog-db/dialog-db/pull/178

notes/notation.md
Gozala: It’s partially implemented. It’s what revision in repository abstraction is holding, but cause currently is a hash referring to prior clam.

notes/notation.md
Gozala: This is what I have implemented in https://github.com/dialog-db/dialog-db/pull/178 and yes there are currently closed set of identifiers.  That was intentional choice so query engine could be implemented in different language without much trouble.

we can do alternative and have something like:

```js
assert: { formula: "math/sum" }
where: { … }
```

That said I somewhat intended to make formula / constraint / concept (rule) applications indistinguishable in way it like if db had those facts, except it’s computed on demand like rules are. Just built-in rules effectively

notes/notation.md
Gozala: Every upper case character is substituted with a lowercase one & prefixed `-` if not followed by another upper case character

PR #233, opened 2026-03-11
1 ✅ feat: use borrowed storage in a tree

Gozala:
This pull requests changes prolly-tree implementation to no longer own storage but instead to receive it as a reference in functions that perform read / write operations.

Goal is to enable integration with capability system which also pass capability provider as reference to effects.


cdata (approved):
LGTM

Gozala:

PR #234, opened 2026-03-12
2 ✅ feat: remove exclusivity constraint imposed by capability system

Gozala:
Design of capability system where `Effect::perform` is passed `&mut Env` was misguided, imposing two constraints:

- **Non-retention**: provider can't be kept beyond execution. This is a lifetime invariant.
- **Exclusivity**: no concurrent access.

Intention was to have 1st and not the 2nd, but I naively assumed `&mut` was more accurate reflected fact that provider often mutates state.

---

This pull request changes interface to use `&Env` which correctly retains 1st invariant removing the second one.

### DIFF
--- a/rust/dialog-storage/src/storage/provider/volatile.rs
+++ b/rust/dialog-storage/src/storage/provider/volatile.rs
@@ -80,6 +80,15 @@ pub enum VolatileError {
     /// CAS condition failed.
     #[error("CAS condition failed: {0}")]
     Cas(String),
+    /// RwLock was poisoned by a panicking thread.
+    #[error("Lock poisoned: {0}")]
+    LockPoisoned(String),
+}
+
+impl<T> From<std::sync::PoisonError<T>> for VolatileError {
+    fn from(e: std::sync::PoisonError<T>) -> Self {
+        VolatileError::LockPoisoned(e.to_string())
+    }
 }
 
 #[cfg(test)]
--- a/rust/dialog-storage/src/storage/provider/volatile.rs
+++ b/rust/dialog-storage/src/storage/provider/volatile.rs
@@ -20,7 +20,7 @@
 //! use dialog_common::Blake3Hash;
 //!
 //! # async fn example() -> anyhow::Result<()> {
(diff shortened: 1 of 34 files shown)


cdata (commented):
It's not a hard requirement, but it might be good to "stress" the intended use cases that warranted this revision in a test e.g., demonstrate that `env` may be shared across threads while still maintaining the lifetime invariant.

rust/dialog-storage/src/storage/provider/volatile.rs
```diff
@@ -58,21 +59,19 @@ struct Session {
 ///
 /// A simple provider that stores all data in memory. Each subject DID gets its
 /// own session with separate archive and memory storage. Data is not persisted.
+///
+/// Uses `RwLock` for interior mutability so that `Provider::execute` can take
+/// `&self`. All lock guards are dropped before any `.await` points.
 #[derive(Default, Debug)]
 pub struct Volatile {
-    sessions: HashMap<Did, Session>,
+    sessions: RwLock<HashMap<Did, Session>>,
```
cdata, line 69: I tend to use `parking_lot::RwLock`, which has some perf / size advantages compared to the standard library and also has an infallible `read` (with a slightly different contract): https://docs.rs/parking_lot/latest/parking_lot/type.RwLock.html

Gozala: Addressed in [chore: switch to parking_lot::RwLock](https://github.com/dialog-db/dialog-db/pull/234/commits/d22c65231bd6987816e5924516326e761c1bd66c)

Gozala:
> It's not a hard requirement, but it might be good to "stress" the intended use cases that warranted this revision in a test e.g., demonstrate that `env` may be shared across threads while still maintaining the lifetime invariant.

Addressed in [chore: switch to parking_lot::RwLock](https://github.com/dialog-db/dialog-db/pull/234/commits/d22c65231bd6987816e5924516326e761c1bd66c)

Gozala:

PR #235, opened 2026-03-12
3 ✅ feat: gen router address

Gozala:
## Goal

Eliminates need for hand-written boilerplate needed when working with multiple network addresses. Without this change it requires manually creating enum and then having bunch of Provider impls that delegate to the Router impl, more annoyingly it required propagating cfg gates which are needed on type bounds.

## Overview

Changes here accomplish goal by updates `#[derive(Router)]` macro so it generates:

1. `{StructName}Address` enum with variant per route with same cfg gates.
2. `From` impls for every address variant in the enum so they can easily be turned into enums.
3. Unified `Provider<RemoteInvocation<Fx, StructAddress>>` that dispatches to corresponding route.

The `Router` was also updated gaining associated `Address` type that gets set to generated enum address.

The `Router` now also has a blanket impl for `ProviderRoute` so it can be used as is.

### DIFF
--- a/rust/dialog-macros/src/router.rs
+++ b/rust/dialog-macros/src/router.rs
@@ -27,6 +27,7 @@
 //! Generates per-field `Provider<RemoteInvocation<Fx, FieldAddress>>` impls,
 //! plus a unified `NetworkAddress` enum and dispatch impl.
 
+use convert_case::{Case, Casing};
 use proc_macro::TokenStream;
 use quote::{format_ident, quote};
 use syn::{DeriveInput, parse_macro_input};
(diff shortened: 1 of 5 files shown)


rust/dialog-macros/src/router.rs
```diff
@@ -50,11 +40,71 @@ pub fn generate(input: TokenStream) -> TokenStream {
...
+/// A routable field extracted from the struct.
+struct RoutableField<'a> {
+    field_name: &'a syn::Ident,
+    field_ty: &'a syn::Type,
+    cfg_attrs: Vec<&'a syn::Attribute>,
+    variant_name: syn::Ident,
+}
+
+/// Convert a snake_case identifier to PascalCase.
+fn to_pascal_case(s: &str) -> String {
```
cdata: I think I gave this feedback before, but there is a commonly used crate to implement case conversions that we should probably use unless we have a good reason not to: https://crates.io/crates/convert_case

cdata (changes requested):
LGTM except one minor thing

Gozala:

PR #239, opened 2026-03-17
feat(query): dialog rename macro

jackddouglas:
Sometimes the desired name of an `Attribute` or `Concept` in the query engine doesn't match what Rust allows or conventions produce. There is currently no way to decouple the Rust identifier from the query engine name.

This change adds on opt-in `#[dialog(rename = "...")]` attribute that overrides the string key without changing the Rust identifier, inspired by the `#[serde(rename = "...")]` convention.


rust/dialog-query/tests/dialog_rename_test.rs
```diff
@@ -0,0 +1,192 @@
...
+
+// -- Attribute rename tests --
+
+mod item {
+    use dialog_query::Attribute;
+
+    /// The type of item
+    #[derive(Attribute, Clone, PartialEq)]
+    #[dialog(rename = "type")]
+    pub struct Kind(pub String);
```
Gozala: As per comment above, this seems unnecessary and I would prefer to leave out unless there is a reason I'm overlooking.

rust/dialog-macros/src/query/attribute.rs
Gozala, line 1: Unless there is a reason to support this, I would prefer to leave it out. Unless I'm overlooking something attributes don't have a need for renaming as no words starting with capital case are reserved.

rust/dialog-query/tests/dialog_rename_test.rs
Gozala, line 1: Can we please move tests related to concepts into concept.rs to stay consistent with rest of the system.

Gozala (changes requested):
Generally looks good, but I'd like to make sure we're not introducing unnecessary abstractions like attribute renames as pointed out in the review.

Gozala:
Landed on `main` via #367 (a rebase of this branch onto current `main`, with the migrated tests adapted to the `Concept::descriptor()` API and the rename parser returning `syn::Error` instead of panicking). Closing as superseded — thanks @jackddouglas!

---

PR #240, opened 2026-03-19
📝 feat: design describing version control

Gozala:
Document alternative design for version control and provenance handling that addresses cross repository limitations of divergence-clocks

### DIFF
--- a/notes/version-control.md
+++ b/notes/version-control.md
@@ -1,0 +1,307 @@
+# Version Control
+
+## Context
+
+The [divergence clock] design encodes causal position as `{ since, at, drift }` where `since` increments at synchronization points, enabling cheap concurrency detection: two changes with the same `since` and different `at` values are concurrent by inspection, no traversal required.
+
+This works well for a single collaborative repository. The limitation surfaces when two repositories with independent histories need to interact. The `since` counter is local to a repo's synchronization history. There is no meaningful way to compare `since: 3` from repo A with `since: 3` from repo B. They count different sync events. A cross-repo merge would require either renumbering one history (losing provenance) or accepting that the clock values are simply incommensurable.
+
+A desired property of Dialog's collaboration model is that forks and follows work as first-class operations across independent repositories. This document describes a causal encoding grounded in the revision DAG that aims to preserve the divergence clock's fast concurrency detection while composing correctly across repo boundaries.
+
+## Idea
+
+Instead of deriving causal position from a logical counter, derive it from the structure of the revision DAG directly. Every revision has a natural position: the count of revisions in the causal chain leading to it. This is its `Edition`. An author increments their edition with each local revision and advances it to `max(seen) + 1` on sync. This is isomorphic to a Lamport timestamp, which gives it a useful property: a higher edition has seen at least as much causal history as any lower one, regardless of which repository it came from.
+
+`Edition` alone is not enough to identify a revision globally, because two authors could independently reach the same edition count. It is therefore paired with `Origin`, a repository-scoped identity derived as `Blake3(issuer + subject)`. Deriving origin from both the signing key and the repository DID ensures that the same principal acting on two different repositories produces two distinct origins. Without this, a principal whose histories in two separate repositories later merge could produce colliding identifiers.
+
+Together, `Origin` and `Edition` form a `Version`: a compact revision identifier that sorts naturally by causal depth and uniquely addresses any revision across all repositories.
+
+## Core Types
+
+```rust
+/// Root of the search tree at a given revision
+#[derive(Attribute, Debug, Clone)]
+#[domain("dialog.revision")]
+pub struct Tree(Blake3Hash);
+
+/// Count of revisions in the causal chain leading to this one.
+/// Increments locally on each revision; advances to max(seen) + 1 on sync.
+/// Isomorphic to a Lamport timestamp.
+#[derive(Attribute, Debug, Clone)]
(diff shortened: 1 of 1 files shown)


jackddouglas (commented):
While I understand how `Edition` is analogous to Lamport timestamps, I don't understand the claim that a larger edition guarantees seeing at least as much causal history. In the isolated sense, yes, but if two repos evolve from genesis separately then try to merge, if A has edition 100 and B has edition 5, A has seen more causal history but none of B's causal history. The inverse is trivially true: B has seen less causal history than A so it can't be a descendant, but I'm not sure how comparing separate causal histories in this way is useful.

I think this is just a language point, not sure it actually impacts the spec, but it's better to say that editions are useful because they indicate whether one repo might be a descendant/subset of another.

Where it does seem to matter is in cause chain traversal. If causes could potentially hold multiple merge nodes, then you might stop prematurely when finding a lower edition on one branch, while the actual target was on another branch further down. So the only way to definitively know the target doesn't exist in the DAG is to traverse it completely.

For `Origin`, what happens after key rotation? The origin would change and wouldn't be a stable identifier of repo membership.

`Version` should include content hash itself to be more specific, because if Alice has the same repo on two devices and makes concurrent edits locally then tries to sync, the versions will be the same. There's currently no path for surfacing conflict resolution between identical versions. Not sure this is possible if revisions are stored in the tree they describe, so they might need to be separate like Git, or this is an acceptable tradeoff.

As I understand it, `Version` is a workaround to a circularity problem when inserting a new revision: `T = hash(tree_containing(claim(is: hash(Revision(tree: T)))))`. To insert the revision you need the root hash of the tree, which you can't find until inserting the revision. `Version` gives you the ability to compare editions trivially, provenance baked in, and a workaround to the above problem, but it doesn't resolve conflicts around the content itself.

I'm assuming Dialog doesn't want to give up the ability to query revisions the same way as everything else. Claude told me Datomic has an interesting approach where transactions are still queryable as data, but transaction IDs are "assigned monotonically by a single transactor" rather than content-hashed. This avoids circularity because IDs are independent of tree state and a single transactor never reuses an ID. Obviously this won't work for Dialog because it's hardly local first.

Another interesting option is having claim `is:` point backward. A revision claim records "the parent of edition N is the content-hashed revision X" instead of "I am edition N." Essentially the same thing as `cause` but recursively applied to revisions. One tradeoff is that finding the current revision is more complex because you need to find the revision claim with an `is:` that no other claim's parent references. The first claim's `is:` would point to a genesis node or the repo DID.

I guess the challenge here though is that you still need to put the revision metadata somewhere, and that still affects the tree hash. So you could track two tree hashes, one before the current claim was added, and another that resolves after the revision was added, which means there's no single tree hash to look up.

Lastly, any ideas about how to optimize storage when causality is attached to every single transaction?

Gozala:
> While I understand how Edition is analogous to Lamport timestamps, I don't understand the claim that a larger edition guarantees seeing at least as much causal history. In the isolated sense, yes, but if two repos evolve from genesis separately then try to merge, if A has edition 100 and B has edition 5, A has seen more causal history but none of B's causal history. The inverse is trivially true: B has seen less causal history than A so it can't be a descendant, but I'm not sure how comparing separate causal histories in this way is useful.

I think you are correct and my claim should probably be tightened than stated, I think this would be more accurately stated version:

Within a **shared** causal history, a higher edition is either a descendant of, or concurrent with, any lower edition; it is never an ancestor.

> I think this is just a language point, not sure it actually impacts the spec, but it's better to say that editions are useful because they indicate whether one repo might be a descendant/subset of another.

Yeah the point I'm trying to make that this tells you that if you traverse history with higher edition you'll be able to identify if it includes one you're comparing to or if it is concurrent and you'd need at most to descend delta between editions + 1

Gozala:
> Where it does seem to matter is in cause chain traversal. If causes could potentially hold multiple merge nodes, then you might stop prematurely when finding a lower edition on one branch, while the actual target was on another branch further down. So the only way to definitively know the target doesn't exist in the DAG is to traverse it completely.

You are correctly pointing out problem, but I think conclusion is incorrect. Yes when history branches you have to explore each at most till target.edition - 1 point, which is greater than claimed delta between two revisions, but still amount of traversal needed is proportional to amount of divergence as opposed to amount of history in the repo which is what full DAG traversal would imply.

It is still worth noting that in linear history case which is the common one it's still amounts to delta between editions. In pathological case it's greater but still much smaller than having to traverse the whole DAG.

Gozala:
> For `Origin`, what happens after key rotation? The origin would change and wouldn't be a stable identifier of repo membership.

After rotation the new key produces a new origin and is treated as a distinct participant in the revision history. That said UCAN delegation chain will still trace it back to the membership delegation as in how they gained access, unless they mint a new on of course.

Do you think there is a problem with this or was it just for clarification ?

Gozala:
> `Version` should include content hash itself to be more specific, because if Alice has the same repo on two devices and makes concurrent edits locally then tries to sync, the versions will be the same. There's currently no path for surfacing conflict resolution between identical versions. Not sure this is possible if revisions are stored in the tree they describe, so they might need to be separate like Git, or this is an acceptable tradeoff.

I think this misses that each device will have unique key and paired with repo did itself unique origin so same person same repo across devices will not collide because origins will be different

jackddouglas:
> Do you think there is a problem with this or was it just for clarification ?

Not necessarily, just wasn't sure if it was meant to be stable or not.

PR #242, opened 2026-03-24
📬 feat: replace bit-batch rank derivation with threshold-based geometric distribution

Gozala:
The previous implementation grouped hash bits into k-bit batches where `k = ceil(log₂(BRANCH_FACTOR))`. For BRANCH_FACTOR=254, this gave k=7, making each trial probability `1/2^7 = 1/128` instead of `1/254`.  This made effective branch factor ~128, not the declared 254.
 
 > This is also why number of read/writes in tests went down

Proposed threshold approach offers IMO much easier to reason about algorithm, which is described in detail in the code comments. 

This also addresses prior upper limit on the possible branching factor.

### DIFF
--- a/rust/dialog-artifacts/src/artifacts.rs
+++ b/rust/dialog-artifacts/src/artifacts.rs
@@ -670,13 +670,11 @@ mod tests {
             )
         };
 
-        // The old prolly tree pinned its eagerly loaded root node in
-        // memory, so query read counts excluded the root. The search tree
-        // holds only the root hash and reads the root block on first use;
-        // the shared node cache absorbs that read on every subsequent
-        // query against the same tree (see it_uses_indexes_to_optimize_reads
-        // where the second select costs one read fewer than it used to).
-        assert_eq!(net_reads, 3);
+        // The threshold-based geometric distribution gives an exact 1/m
+        // split at every level, so the tree is flatter than the bit-batch
+        // distribution (whose upper levels averaged only 2-4 children) and a
+        // point query walks one fewer block to reach its leaf.
+        assert_eq!(net_reads, 2);
         assert_eq!(net_writes, 0);
 
         Ok(())
@@ -734,13 +732,11 @@ mod tests {
             )
         };
 
-        // The old prolly tree pinned its eagerly loaded root node in
-        // memory, so query read counts excluded the root. The search tree
-        // holds only the root hash and reads the root block on first use;
-        // the shared node cache absorbs that read on every subsequent
-        // query against the same tree (see it_uses_indexes_to_optimize_reads
-        // where the second select costs one read fewer than it used to).
-        assert_eq!(net_reads, 3);
+        // The threshold-based geometric distribution gives an exact 1/m
+        // split at every level, so the tree is flatter than the bit-batch
+        // distribution (whose upper levels averaged only 2-4 children) and a
+        // point query walks one fewer block to reach its leaf.
+        assert_eq!(net_reads, 2);
         assert_eq!(net_writes, 0);
 
         Ok(())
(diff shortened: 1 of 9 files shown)


rust/dialog-prolly-tree/src/distribution/geometric.rs
```diff
@@ -28,46 +81,70 @@ where
...
-        if batch != 0 {
-            return i + 1; // +1 because geometric distribution starts at 1
-        }
+    // Extract the first 8 bytes of the hash as a u64 prefix.
+    // Little-endian is used for consistency with the hash output byte order
+    // (the choice of endianness doesn't affect uniformity, but must be
+    // deterministic for consistent tree structure).
+    let prefix = u64::from_le_bytes(
+        bytes[0..8]
+            .try_into()
```
cdata: This is allocating an array per invocation. Can you try re-writing it without the allocation?

rust/dialog-prolly-tree/src/distribution/geometric.rs
```diff
@@ -28,46 +81,70 @@ where
...
+            .expect("hash must be at least 8 bytes"),
```
cdata: No panics, please. If it's fallible, return a `Result`.

rust/dialog-prolly-tree/src/distribution/geometric.rs
```diff
@@ -28,46 +81,70 @@ where
...
+    );
+
+    // Start with rank 1 (the "base" level — most keys land here).
+    let mut rank: Rank = 1;
+
+    // The first threshold: prefix values below this represent the ~1/m fraction
+    // of keys that should be promoted to rank 2 or higher.
+    let mut threshold = u64::MAX / u64::from(m);
```
cdata: Should `m` just be a `u64` so that we don't repeatedly convert it?

rust/dialog-prolly-tree/src/distribution/geometric.rs
```diff
@@ -4,11 +4,64 @@ use crate::KeyType;
...
-/// series of fair Bernoulli trials (p = 1 / 2). The number of trials is limited
-/// to 256 independent trials.
+/// Compute the maximum rank derivable from a u64 hash prefix for a given
+/// branch factor Q. This is `floor(log_Q(2^64))` — the number of times we
+/// can divide `u64::MAX` by Q before the threshold reaches zero.
 ///
-/// https://textile.notion.site/Flipping-bits-and-coins-with-hashes-205770b56418498fba4fef8cb037412d
+/// For Q=254 this gives 8, supporting trees with up to ~10^19 entries.
+const fn max_rank_for(branch_factor: u32) -> Rank {
+    let m = branch_factor as u64;
```
cdata: It seems like `Rank` should just be a `u64`.

Gozala:
Pushed ce4a33f7 addressing the review feedback:

- **No allocation, no panic** (both comments resolve together): the prefix is now extracted by destructuring the first 8 bytes of the fixed-size hash with an array pattern, so there is no slice conversion at all. On the question of returning a `Result` instead: rank derivation over a 32-byte hash cannot fail, and a fallible signature would force error plumbing through every caller (including the `Distribution` trait) for an unreachable case. Padding short inputs was also considered and rejected: zero-padding biases short inputs toward *higher* ranks (low prefix = promoted), which would silently turn a misuse into a wrong tree shape. Taking `&[u8; 32]` makes the invalid input unrepresentable instead, which seems strictly better than either erroring or papering over it.
- **`m` is now `u64`** end to end, including the `Distribution::BRANCH_FACTOR` trait const, so nothing converts per call.
- **`Rank` is now `u64`**, matching the threshold arithmetic.

Beyond the review comments:

- The explicit `max_rank_for` cap is gone: integer division drives the threshold to zero on its own, so the loop terminates naturally at `floor(log_m(2^64)) + 1` (rank 9 for Q=254, probability ~1/254^8). The cap was also where an `m = 1` input would loop forever; a `debug_assert!(m >= 2)` now documents the domain.
- The statistical tests use a fixed-seed `StdRng`, so they are deterministic rather than racing sigma tolerances against `thread_rng` (the factor=1000 case was running at roughly 3.5 sigma, which would have flaked eventually).
- New tests: exact threshold boundary goldens (including the equality case, which must NOT promote since the comparison is strict), a per-level conditional promotion test asserting `P(rank >= k+1 | rank >= k) = 1/m`, which is the direct regression test for the bit-batch straddling bug this PR fixes (the old code yielded 1/2, 1/4, 1/8 there), and a determinism test proving only the 8-byte prefix participates.
- The same test suite is ported to `dialog-search-tree`, which previously had no distribution tests.

The duplication of the implementation across the two crates is left as is on purpose: `dialog-prolly-tree` is slated for removal once `dialog-search-tree` replaces it, so the duplicate dies with it.

PR #245, opened 2026-04-01
5 ✅ feat: add Checksum types and #[derive(Attenuate)] macro

Gozala:
## Summary

- Add Checksum and Hasher types to dialog-common for content integrity verification using multihash-encoded SHA-256 digests
- Add #[derive(Attenuate)] proc macro to dialog-macros that generates projections for authorizations

The Claim trait allows effect types to declare how they should be represented during authorization. Fields annotated with `#[attenuate(into = Type)]` are projected via `From` conversion (e.g., replacing raw content bytes with their checksum). Fields can also use custom conversion functions via `#[attenuate(into = Type, with = path)]` and be renamed with `rename = name`.

For types with no annotated fields, `Attenuate::Attenuation = Self` (identity projection). For types with annotations, a parallel `{Name}Attenuation` struct is generated automatically.

### DIFF
--- a/rust/dialog-common/src/checksum.rs
+++ b/rust/dialog-common/src/checksum.rs
@@ -115,8 +115,8 @@ impl TryFrom<Vec<u8>> for Checksum {
 impl Checksum {
     /// Compute the SHA-256 checksum of the given data.
     ///
-    /// This is the primary constructor for use with `#[derive(Claim)]`:
-    /// `#[claim(with = Checksum::sha256, rename = checksum)]`
+    /// This is the primary constructor for use with `#[derive(Attenuate)]`:
+    /// `#[attenuate(with = Checksum::sha256, rename = checksum)]`
     pub fn sha256(data: impl AsRef<[u8]>) -> Self {
         Hasher::Sha256.checksum(data.as_ref())
     }
--- a/rust/dialog-common/src/checksum.rs
+++ b/rust/dialog-common/src/checksum.rs
@@ -1,0 +1,202 @@
+//! Checksum types for content integrity verification.
+//!
+//! Provides checksum types used for content-addressed storage and authorization.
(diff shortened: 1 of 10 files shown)


rust/dialog-common/src/checksum.rs
```diff
@@ -0,0 +1,202 @@
...
+    fn from(checksum: &Checksum) -> Self {
+        base64::engine::general_purpose::STANDARD.encode(checksum.as_bytes())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
```
cdata, line 155: Probably good to make sure that these tests are running under `wasm32-unknown-unknown` as well, just in case there is platform incompatibility via a transitive dependency.

rust/dialog-macros/src/lib.rs
```diff
@@ -394,3 +395,28 @@ pub fn derive_attribute(input: TokenStream) -> TokenStream {
...
+/// pub struct Put {
+///     pub digest: Blake3Hash,
+///     #[claim(into = Checksum)]
+///     pub content: Vec<u8>,
+/// }
+/// // Generates PutClaim { digest: Blake3Hash, content: Checksum }
+/// // and impl Claim for Put { type Claim = PutClaim; ... }
+/// ```
+#[proc_macro_derive(Claim, attributes(claim))]
+pub fn derive_claim(input: TokenStream) -> TokenStream {
```
cdata: This is new territory for me, but we're getting to the point where I wish we could author tests to assert the properties of our macro implementation. I wonder what the derive macro test strategy looks like for e.g., `serde` or `clap`.

cdata (approved):
LGTM although please consider the comments

Gozala:

PR #249, opened 2026-04-01
9 ✅ feat: redesign access capabilities

Gozala:
## Why

The old access module had authorization baked into Protocol::authorize as a monolithic async method requiring the full provider stack (Identify + Sign + List + Get) upfront. This made it impossible to separate proof discovery from signing, or to narrow time bounds after finding a chain. Every protocol implementation had to handle the entire flow, duplicating logic.

## Overview

Replaces monolithic Protocol::authorize with a multi-step flow:
1. Prove: walk the certificate store, find a delegation chain, return a Proof with effective time bounds (intersection of all certificate ranges)
2. claim(signer): bind a signer to produce an Authorization
3. not_before()/expires(): optionally narrow (never widen) the time window
4. delegate()/invoke(): produce a delegation or signed invocation

New foundation types: Site (remote endpoint marker), Fork/ForkInvocation (remote dispatch), Subject::any() for powerline authorization, StorageError in error.rs.

Adds #[derive(Claim)] to all effect types in dialog-effects to satisfy the new Effect: Claim bound.

Removes the old ucan module and feature from dialog-capability, dialog-effects, dialog-credentials, and CI. The old authority, authorization, router modules are removed. UCAN support will be reintroduced as the standalone dialog-ucan crate.

Integration tests for the Prove/CertificateStore flow come with the storage providers in a later PR.


rust/dialog-capability/src/authority.rs
```diff
@@ -1,14 +1,148 @@
...
+/// # use dialog_capability::authority::{Profile, Operator, Authority};
+/// # let authority: Authority = todo!();
+/// let repo_subject = authority.subject();
+/// let profile = Profile::of(&authority);
+/// let operator = Operator::of(&authority);
+/// ```
+pub type Authority = Capability<Operator>;
+
+/// Sign operation — signs a payload using the operator's key.
+#[derive(Debug, Clone, Serialize, Deserialize, crate::Claim)]
```
cdata: Nit: if there isn't a specific reason for the path qualification, I would just as soon prefer `Claim` over `crate::Claim`.

rust/dialog-capability/src/fork.rs
```diff
@@ -0,0 +1,142 @@
+//! Fork — remote execution of capabilities at a site address.
+//!
+//! A [`Fork`] pairs a [`Capability`] with a site address for remote execution.
```
cdata, line 3: [AFAIK "Fork" has no direct analog in Git](https://git-scm.com/docs/gitglossary). Instead, it's a Github term which refers to a replica of the project repository, hosted in a _different_ project that observes a _different_ authority from the source.

Another case of "Fork" in use that I'm aware of is when you spawn a process on Unix-likes by duplicating the calling process.

I'm not sure which definition to apply here, but as I read it neither of those definitions apply - is that right?

cdata (approved):
LGTM, although I'm not sure about "Fork" as a name

rust/dialog-capability/src/fork.rs
Gozala: > Another case of "Fork" in use that I'm aware of is when you spawn a process on Unix-likes by duplicating the calling process.
>
> I'm not sure which definition to apply here, but as I read it neither of those definitions apply - is that right?

I was mimicking unisons notion of [spawning a task in a remote node](https://www.unison-lang.org/blog/visualizing-remote/#drawing-remote-primitive-operations) which they call fork. I'm open to renaming it because it's likely others may get confused between those two meanings also

rust/dialog-capability/src/authority.rs
Gozala: fixed

Gozala:

PR #251, opened 2026-04-01
8 ✅ feat: add credential newtypes and key export

Gozala:
## Why

Downstream code (profiles, repositories, storage providers) needs to handle credentials at a higher level than raw Ed25519 signers. A profile always has a signer, but a repository might receive a credential from a delegation that only carries the public key. Without newtypes, every consumer has to match on key material directly and handle the signer/verifier distinction ad hoc.

SignerCredential wraps a signer with Issuer capability. VerifierCredential wraps a public key for verification only. Credential is the enum for contexts where either is valid. Key export is extended with platform-specific formats (multicodec bytes on native, CryptoKey on wasm) so credentials can be persisted and restored across sessions.

### DIFF
--- a/rust/dialog-credentials/src/credential/export.rs
+++ b/rust/dialog-credentials/src/credential/export.rs
@@ -3,14 +3,20 @@
 //! These are plain data containers for serialized credential material.
 //! Import/export logic lives on the credential types themselves.
 
+use thiserror::Error;
+
+#[cfg(not(target_arch = "wasm32"))]
+use super::constants::{
+    ED25519_PRIV_TAG, ED25519_PUB_TAG, PUB_KEY_OFFSET, SIGNER_EXPORT_SIZE, VERIFIER_EXPORT_SIZE,
+};
+
 #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
 use js_sys::Uint8Array;
 #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
 use wasm_bindgen::{JsCast, JsValue};
 
-
 /// Error type for credential export/import operations.
-#[derive(Debug, thiserror::Error)]
+#[derive(Debug, Error)]
 pub enum CredentialExportError {
     /// Key export/import operation failed.
     #[error("key operation failed: {0}")]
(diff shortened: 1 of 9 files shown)


rust/dialog-credentials/src/credential/export.rs
```diff
@@ -0,0 +1,192 @@
+//! Platform-specific credential export types.
+//!
+//! These are plain data containers for serialized credential material.
+//! Import/export logic lives on the credential types themselves.
+
+/// Error type for credential export/import operations.
+#[derive(Debug, thiserror::Error)]
```
cdata: nit: I prefer to use fully-qualified paths like this only in the case where there is potential ambiguity.

rust/dialog-credentials/src/credential/export.rs
```diff
@@ -0,0 +1,192 @@
...
+
+/// Platform-specific serialized form of a credential (signer or verifier).
+#[derive(Debug, Clone)]
+pub enum CredentialExport {
+    Signer(SignerCredentialExport),
+    Verifier(VerifierCredentialExport),
+}
+
+/// Multicodec varint for ed25519 private key (0x1300).
+pub(crate) const ED25519_PRIV_TAG: &[u8] = &[0x80, 0x26];
```
cdata: nit: I think these might make more sense in a `constants.rs` module that isn't exported publicly (but can be used by internal library code). It would spare you from `pub(crate)` and make it clear that these are shared values.

rust/dialog-credentials/src/key.rs
```diff
@@ -89,3 +132,103 @@ pub trait ExtractableKey: Sized {
...
+        match (&original, &restored) {
+            (KeyExport::Extractable(a), KeyExport::Extractable(b)) => assert_eq!(a, b),
+            #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+            _ => panic!("expected extractable"),
+        }
+    }
+}
+
+#[cfg(all(test, target_arch = "wasm32", target_os = "unknown"))]
+mod wasm_tests {
```
cdata: A few nits here:

1. `web_tests` (not `wasm_tests`)
2. Some token comments to explain why these need to be web specific
3. BDD naming throughout, please

cdata (approved):
Some small things but otherwise LGTM

rust/dialog-credentials/src/credential/export.rs
Gozala: done

rust/dialog-credentials/src/credential/export.rs
Gozala: done

Gozala:

PR #252, opened 2026-04-01
7 ✅ feat: add derive(Provider) compose macro, Issuer, and Command

Gozala:
## Why

Composite providers (like Operator, MountedSpace) need to dispatch different effects to different inner providers based on type. Without this, every composite type requires hand-written Provider impls for each effect it handles, which is verbose and error-prone.

The Issuer trait is needed by SignerCredential (PR 8) to express that a signer can issue capabilities. Command is a minimal trait pairing input/output types, used by the fork dispatch system for remote invocations.

## Overview

Adds `#[derive(Provider)]` proc macro that generates `Provider<Fx>` impls for composite structs. Each field is annotated with `#[provide(Fx1, Fx2, ...)]` to declare which effects route to it. The macro generates a forwarding impl for each listed effect.

Supports enum variants with `#[cfg]` gating for platform-specific dispatch (e.g., FileStore on native, IndexedDb on wasm).

Also adds:
- `Issuer` trait: marks types that can issue capabilities (Principal + Signer)
- `Command` trait: pairs Input and Output types for non-capability dispatch

### DIFF
--- a/rust/dialog-capability/src/command.rs
+++ b/rust/dialog-capability/src/command.rs
@@ -1,0 +1,22 @@
+use crate::{Capability, Constraint, Effect};
+
+/// Trait for types that can be executed via a Provider.
+///
+/// This trait connects a command type to what the provider receives
+/// as input and what it produces as output.
+pub trait Command {
+    /// The input type passed to Provider::execute.
+    type Input;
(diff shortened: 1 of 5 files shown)


cdata (approved):

Gozala:

PR #265, opened 2026-04-08
9 🏗️ feat: redesign access capability system

Gozala:
## Why

The old access module had authorization baked into the Protocol trait as a single async method that required the full provider stack (Identify + Sign + List + Get) upfront. This made it impossible to separate proof discovery from signing, or to narrow time bounds after finding a chain. It also meant every protocol implementation had to handle the entire authorization flow, duplicating logic.

The old ucan module in dialog-capability was coupled to this design and to the storage KV hierarchy being removed. It is deleted here and will be cleanly reimplemented as the standalone dialog-ucan crate against the new traits.

## Overview

Replaces the monolithic Protocol::authorize with a multi-step flow:

1. Prove: walk the certificate store, find a delegation chain, return a Proof with effective time bounds (intersection of all certificate ranges)
2. claim(signer): bind a signer to the proof, producing an Authorization
3. not_before()/expires(): optionally narrow (never widen) the time window
4. delegate()/invoke(): produce a delegation bundle or signed invocation

This separation means proof discovery, signing, and time scoping are independent steps that can be composed differently depending on context (operator bootstrap, user delegation, remote invocation).

Integration tests for the Prove/CertificateStore flow are added later with the storage capability providers that implement CertificateStore. This PR defines the traits, the default BFS chain-walking algorithm, and unit tests for TimeRange logic.


Gozala:
Superseded by reordered PR stack

PR #268, opened 2026-04-08
11 ✅ chore: rename dialog-ucan to dialog-ucan-core

Gozala:
## Why

The dialog-ucan name will be used for dialogs embedding of ucan, implementing capability protocol  integration (the bridge between dialog-capability's access system and UCAN tokens). The low-level UCAN spec implementation moves to dialog-ucan-core to avoid entangling two

## Overview

Pure rename: dialog-ucan becomes dialog-ucan-core. No code changes, just crate name and identifier updates across the workspace. The container module (DelegationChain, InvocationChain) is added in the next PR.


cdata (approved):
Soft bikeshed, but I think this makes more sense the other way around e.g., the core implementation is `dialog-ucan` and the integration is `dialog-ucan-integration` or something like that.

Gozala:

PR #246, opened 2026-04-01
6 ✅ chore: remove S3 backend, network provider, and dialog-s3-credentials

Gozala:
## Summary

- Delete `dialog-s3-credentials` crate entirely (replaced by `dialog-remote-s3` + `dialog-remote-ucan-s3` in upcoming PRs)
- Delete S3 backend and network provider from `dialog-storage`
- Delete `dialog-effects::remote` module (uses old `Invocation` trait)
- Delete S3 integration test workflow and tests

This unblocks the capability system restructure by removing all external consumers of the old `Authorization`, `Authorized`, `Claim<C>`, and `Invocation` types.

PR #269, opened 2026-04-08
12 ✅ feat: add UCAN container format

Gozala:
## Overview

This migrates code we have in [https://github.com/dialog-db/dialog-db/tree/main/rust/dialog-s3-credentials/src/ucan] that was removed under https://github.com/dialog-db/dialog-db/pull/246 into `dialog-ucan-core` specifically `DelegationChain` and `InvocationChain` implementations:

- `DelegationChain`: ordered sequence of signed delegations from subject to audience. Principal alignment is enforced (each delegation's audience must match the next delegation's issuer). Tracks effective time bounds as the intersection of all delegation ranges. Serializes to DAG-CBOR container format.

- `InvocationChain`: pairs a signed invocation with the delegation proofs authorizing it. Verifies the invocation's proofs field references the correct delegation CIDs.

Both support roundtrip serialization via the [ctn-v1 container format](https://github.com/ucan-wg/container).

### DIFF
--- a/rust/dialog-ucan-core/src/container.rs
+++ b/rust/dialog-ucan-core/src/container.rs
@@ -1,0 +1,194 @@
+//! UCAN Container format utilities.
+//!
+//! This module provides [`Container`], a type that represents a UCAN container
+//! following the [UCAN Container spec](https://github.com/ucan-wg/container).
+//!
+//! The container format is:
+//! ```text
+//! { "ctn-v1": [token_bytes_0, token_bytes_1, ..., token_bytes_n] }
+//! ```
+//!
+//! Where tokens are DAG-CBOR serialized UCANs.
+//!
+//! # Usage
+//!
+//! `Container` can be converted to/from:
+//! - [`DelegationChain`] - A chain of delegations
+//! - [`InvocationChain`] - An invocation with its delegation chain
+
+pub mod delegation;
+pub mod invocation;
+
+mod check_failed;
+pub use check_failed::check_failed_to_container_error;
+
+use ipld_core::ipld::Ipld;
+use std::collections::BTreeMap;
+use thiserror::Error;
+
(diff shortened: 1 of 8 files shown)


rust/dialog-ucan-core/src/container.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// Parse a container from DAG-CBOR bytes.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// - The bytes are not valid DAG-CBOR
+    /// - The container is missing the "ctn-v1" key
+    /// - The tokens array is invalid
+    /// - The container is empty
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, ContainerError> {
```
cdata, line 81: Is there a reason why serde wasn't the right tool for (de)serialization here?

rust/dialog-ucan-core/src/container/delegation.rs
```diff
@@ -0,0 +1,783 @@
...
+        DelegationChain::try_from(bytes.as_slice()).map_err(serde::de::Error::custom)
+    }
+}
+
+/// Helper functions for testing delegation chains.
+///
+/// These are exported when the `helpers` feature is enabled so that
+/// other crates can use them in their tests.
+#[cfg(test)]
+pub mod helpers {
```
cdata, line 348: This should probably just be its own module

cdata (approved):
LGTM

rust/dialog-ucan-core/src/container.rs
Gozala: I mean we use serde inside aren't we ? with just more invariant checks. Perhaps we could do more declarative way like with field renaming, which I guess might be doable but not something that occured to me

Gozala:

PR #279, opened 2026-04-09
14 ✅ feat: add credential providers, FileSystemHandle API, Space, and preludes

Gozala:
## Summary

Storage providers need credential persistence so that spaces can save and
load their identity on bootstrap. Each backend (FileSystem, Volatile,
IndexedDb) needs to store credentials in its native format and expose them
through the capability effect system.

The filesystem provider was also coupled to raw PathBuf and tokio calls,
making the code fragile (a trailing-slash bug caused PidlockGuard to
busy-spin indefinitely). FileSystemHandle now encapsulates all filesystem
operations so provider modules work through a clean abstraction.

Space composes the per-concern providers (archive, memory, credential,
certificate) into a single unit that Storage can mount per DID. Preludes
give tests and consumers a fluent API instead of verbose attenuate/invoke
chains.


rust/dialog-credentials/src/credential/signer.rs
```diff
@@ -97,9 +97,22 @@ impl SignerCredential {
         let seed: &[u8; KEY_SIZE] = data[PRIV_TAG_SIZE..PUB_KEY_OFFSET]
             .try_into()
             .map_err(|_| CredentialExportError::InvalidFormat("invalid seed".into()))?;
+        let stored_pubkey: &[u8; KEY_SIZE] = data[PUB_KEY_OFFSET + PUB_TAG_SIZE..]
```
cdata, line 100: The ship sort of sailed here, but...

nit: prefer full words over abbreviations in identifiers e.g., `PUBLIC_KEY_OFFSET`, `PRIVATE_TAG_SIZE`

rust/dialog-storage/src/storage/provider/fs/archive.rs
```diff
@@ -86,82 +77,96 @@ impl Provider<Put> for FileSystem {
...
-        .parse()
-        .unwrap();
-        Subject::from(did)
+    use crate::resource::Resource;
+    use dialog_capability::Did;
+    use dialog_effects::prelude::*;
+    use dialog_effects::storage::{Directory, Location as StorageLocation};
+
+    fn unique_name(prefix: &str) -> String {
+        use dialog_common::time;
```
cdata, line 86: Is this a re-export of `web_time`?

(I have a strong bias against re-exports like this, as we don't control the API of 3P crates)

rust/dialog-storage/src/storage/provider/fs/archive.rs
```diff
@@ -86,82 +77,96 @@ impl Provider<Put> for FileSystem {
...
+        format!("{prefix}-{ts}-{seq}")
+    }
+
+    async fn unique_did() -> Did {
+        let signer = dialog_credentials::Ed25519Signer::generate().await.unwrap();
+        dialog_varsig::Principal::did(&signer)
     }
 
     #[dialog_common::test]
     async fn it_returns_none_for_missing_content() -> anyhow::Result<()> {
```
cdata, line 103: `use anyhow::Result`

rust/dialog-storage/src/storage/provider/fs/credential.rs
```diff
@@ -0,0 +1,217 @@
...
+use dialog_effects::credential::{self, CredentialError};
+
+use super::{FileSystem, FileSystemError, FileSystemHandle};
+
+const CREDENTIAL: &str = "credential";
+const KEY: &str = "key";
+
+impl FileSystem {
+    /// Returns the handle for a key credential at the given address.
+    pub(super) fn credential_key(
```
cdata, line 16: Two things are smelly about this:

1. `pub(super)`
2. Impl outside of the module's source file

2 is allowed, but can be quite confusing because you can't intuit where the implementation came from

1 is smelly to me because if some API ought to be public outside of the module that implements it, there is a fair chance it might be useful to a consumer beyond the scope of `super` or the crate.

rust/dialog-storage/src/storage/provider/fs/credential.rs
```diff
@@ -0,0 +1,217 @@
...
+mod tests {
+    use super::*;
+    use crate::resource::Resource;
+    use dialog_capability::Did;
+    use dialog_credentials::{Ed25519Signer, SignerCredential};
+    use dialog_effects::prelude::*;
+    use dialog_effects::storage::{Directory, Location as StorageLocation};
+    use dialog_varsig::Principal;
+
+    fn unique_name(prefix: &str) -> String {
```
cdata, line 71: Was this copy-pasted? Can we make a `helpers` module?

rust/dialog-storage/src/storage/provider/fs/error.rs
```diff
@@ -19,3 +19,18 @@ pub enum FileSystemError {
     #[error("Containment violation: {0}")]
     Containment(String),
 }
+
+use dialog_capability::access::AuthorizeError;
+use dialog_effects::credential::CredentialError;
```
cdata, line 24: nit: put `use` at the top of the module

rust/dialog-storage/src/storage/provider/fs/archive.rs
```diff
@@ -1,14 +1,22 @@
...
+use super::{FileSystem, FileSystemError, FileSystemHandle};
 use async_trait::async_trait;
 use base58::ToBase58;
 use dialog_capability::{Capability, Provider};
+
+const ARCHIVE: &str = "archive";
+
+impl FileSystem {
+    /// Returns the handle for this space's archive directory.
+    pub fn archive(&self) -> Result<FileSystemHandle, FileSystemError> {
```
cdata, line 14: Similar comment to https://github.com/dialog-db/dialog-db/pull/279/changes#diff-61be54c4c2110492028a9aacfdf4007b01587df3c49ec08d3db4ad879e067c53R16

rust/dialog-storage/src/storage/provider/fs/memory.rs
```diff
@@ -1,12 +1,23 @@
...
 //! content hashing for edition tracking.
 
-use super::{FileSystem, FileSystemError, Location};
+use super::{FileSystem, FileSystemError, FileSystemHandle};
 use async_trait::async_trait;
 use base58::ToBase58;
+
+const MEMORY: &str = "memory";
+
+impl FileSystem {
```
cdata, line 15: Similar comment to https://github.com/dialog-db/dialog-db/pull/279/changes#diff-61be54c4c2110492028a9aacfdf4007b01587df3c49ec08d3db4ad879e067c53R16

rust/dialog-storage/src/storage/provider/fs/memory.rs
```diff
@@ -270,45 +263,46 @@ impl Provider<Retract> for FileSystem {
...
-        )
-        .parse()
-        .unwrap();
-        Subject::from(did)
+    use crate::resource::Resource;
+    use dialog_capability::Did;
+    use dialog_effects::prelude::*;
+    use dialog_effects::storage::{Directory, Location as StorageLocation};
+
+    fn unique_name(prefix: &str) -> String {
```
cdata, line 279: ditto: make it a common helper

rust/dialog-storage/src/storage/provider/indexeddb/archive.rs
```diff
@@ -75,38 +66,35 @@ impl Provider<Put> for IndexedDb {
...
 
 #[cfg(test)]
 mod tests {
+    wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_dedicated_worker);
+
     use super::*;
     use dialog_capability::{Did, Subject};
     use dialog_effects::archive::{Archive, Catalog};
 
+    fn unique_name(prefix: &str) -> String {
```
cdata, line 94: Is the uniqueness requirement less strict here, or could this be the same implementation as the other `unique_name` helpers that use `time` and a counter?

(I assumed the counter was helpful for debugging, but maybe it's over-engineered?)

rust/dialog-storage/src/storage/provider/fs/memory.rs
```diff
@@ -796,4 +814,131 @@ mod tests {
...
+
+        std::fs::remove_dir_all(&dir)?;
+        Ok(())
+    }
+
+    #[dialog_common::test]
+    async fn it_fails_lock_when_directory_exists_at_lock_path() -> anyhow::Result<()> {
+        // If a directory exists where the lock file should be (e.g. from
+        // a previous buggy run), acquire should fail, not spin or panic.
+        let dir = std::env::temp_dir().join(unique_name("fs-lock-dir-exists"));
```
cdata, line 925: Just wondering: was it inconvenient to use the `tempfile` crate for this? It supports creating directories IIRC.

rust/dialog-storage/src/storage/provider/indexeddb/credential.rs
```diff
@@ -0,0 +1,161 @@
...
+use dialog_capability::{Capability, Policy, Provider};
+use dialog_credentials::Credential;
+use dialog_credentials::credential::CredentialExport;
+use dialog_effects::credential::{self, CredentialError};
+use wasm_bindgen::JsValue;
+
+const CREDENTIAL: &str = "credential";
+
+#[async_trait(?Send)]
+impl Provider<credential::Load<Credential>> for IndexedDb {
```
cdata, line 14: nit: `use credential::Load`

rust/dialog-storage/src/storage/provider/indexeddb/credential.rs
```diff
@@ -0,0 +1,161 @@
...
+                    .await
+                    .map_err(|e| CredentialError::Corrupted(e.to_string()))
+            }
+            None => Err(CredentialError::NotFound(idb_key)),
+        }
+    }
+}
+
+#[async_trait(?Send)]
+impl Provider<credential::Save<Credential>> for IndexedDb {
```
cdata, line 47: nit: `use credential::Save`

rust/dialog-storage/src/storage/provider/indexeddb/credential.rs
```diff
@@ -0,0 +1,161 @@
...
+    async fn execute(
+        &self,
+        input: Capability<credential::Save<Credential>>,
+    ) -> Result<(), CredentialError> {
+        let address = credential::Key::of(&input).address.clone();
```
cdata, line 52: nit: `use ...`

rust/dialog-storage/src/storage/provider/indexeddb/memory.rs
```diff
@@ -43,189 +43,161 @@ impl Provider<Resolve> for IndexedDb {
...
 
 #[cfg(test)]
 mod tests {
+    wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_dedicated_worker);
+
     use super::*;
     use dialog_capability::{Did, Subject};
     use dialog_effects::memory::{Cell, Memory, Space};
 
+    fn unique_name(prefix: &str) -> String {
```
cdata, line 197: ditto re: other comments

rust/dialog-storage/src/storage/provider/volatile/credential.rs
```diff
@@ -0,0 +1,66 @@
...
+use dialog_common::{ConditionalSend, ConditionalSync};
+use dialog_credentials::Credential;
+use dialog_effects::credential::{self, CredentialError};
+use dialog_varsig::Principal;
+
+use super::Volatile;
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+impl Provider<credential::Load<Credential>> for Volatile
```
cdata, line 13: `use ...`

rust/dialog-storage/src/storage/provider/volatile/credential.rs
```diff
@@ -0,0 +1,66 @@
...
+                .await
+                .map_err(|e| CredentialError::Corrupted(e.to_string())),
+            None => Err(CredentialError::NotFound(key)),
+        }
+    }
+}
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+impl Provider<credential::Save<Credential>> for Volatile
```
cdata, line 43: `use ...`

rust/dialog-storage/src/storage/provider/fs.rs
```diff
@@ -67,58 +35,81 @@ const MEMORY: &str = "memory";
...
-        Ok(Self(s.try_into()?))
-    }
-}
+    fn try_from(location: &Location) -> Result<Self, FileSystemError> {
+        let base = match &location.directory {
+            Directory::Profile => {
+                let data_dir = dirs::data_dir().ok_or_else(|| {
+                    FileSystemError::Io("could not determine platform data directory".into())
+                })?;
+                data_dir.join("dialog")
```
cdata, line 73: This looks like it should be a constant.

rust/dialog-storage/src/storage/provider/fs.rs
```diff
@@ -67,58 +35,81 @@ const MEMORY: &str = "memory";
...
+        Ok(Self(FileSystemHandle::try_from(location)?))
     }
 }
 
-impl TryFrom<&str> for FileSystem {
+/// Resolve a `Location` (Directory + name) to a filesystem `Location`.
+///
+/// In test mode, all directories resolve under the platform temp dir
+/// to avoid polluting real profile or workspace directories.
+#[cfg(not(test))]
```
cdata, line 63: This is a rather smelly pattern esp. considering the inner implementation is non-trivial. I would recommend refactoring so that the root can be passed in as some kind of configuration.

rust/dialog-storage/src/storage/provider/fs.rs
```diff
@@ -194,18 +185,32 @@ impl TryFrom<PathBuf> for Location {
...
 
-impl TryFrom<Location> for PathBuf {
+impl TryFrom<FileSystemHandle> for PathBuf {
     type Error = FileSystemError;
 
-    fn try_from(location: Location) -> Result<Self, Self::Error> {
-        location
+    fn try_from(location: FileSystemHandle) -> Result<Self, Self::Error> {
+        let path = location
             .0
```
cdata, line 193: In general it is kind of smelly to be reaching into an unnamed field like this and then invoking that field's method (can it be an impl-specific method on `FileSystemHandle`?)

rust/dialog-storage/src/storage/provider/fs.rs
```diff
@@ -194,18 +185,32 @@ impl TryFrom<PathBuf> for Location {
...
             .to_file_path()
-            .map_err(|_| FileSystemError::Io("Failed to convert URL to path".to_string()))
+            .map_err(|_| FileSystemError::Io("Failed to convert URL to path".to_string()))?;
+
+        // Strip trailing slash added by FileSystemHandle for URL semantics.
+        // Filesystem operations (read, write, rename) need clean file paths.
+        let s = path.to_string_lossy();
```
cdata, line 199: This is a bug. Different paths can potentially resolve to the same string (because their non-UTF-8 chars are replaced with the same substitute).

`path.to_str()` and the `None` case should be an error.

rust/dialog-storage/src/storage/provider/indexeddb.rs
```diff
@@ -90,66 +93,24 @@ impl Session {
...
-            self.upgrade(stores).await?;
-        }
-
-        Ok(Store {
-            db: self.db.as_ref().unwrap(),
-            name: self.stores.get(store_path).unwrap(),
-        })
+    /// Returns a clone of the database handle.
+    fn db(&self) -> Rc<Rexie> {
+        self.db.as_ref().expect("connection not open").clone()
```
cdata, line 102: No panic-ing at runtime please. I like lazy initialization primitives e.g., `OnceCell` for this because access typically implies initialization.

rust/dialog-storage/src/storage/provider/space.rs
```diff
@@ -0,0 +1,134 @@
+//! Space: a composed set of providers for a mounted space.
```
cdata, line 1: A little more detail about when / why this is helpful would be welcome.

cdata (changes requested):
Cool to see it coming together.

rust/dialog-storage/src/storage/provider/fs/archive.rs
Gozala: Not exactly, because you have to do this on wasm

https://github.com/dialog-db/dialog-db/blob/af442cac90d72c9da8be9c71799f497bddc62f0b/rust/dialog-common/src/time.rs#L22-L23

and following on native

https://github.com/dialog-db/dialog-db/blob/af442cac90d72c9da8be9c71799f497bddc62f0b/rust/dialog-common/src/time.rs#L11-L14

And having to remember to add `.to_std()` and gate it by target is pain which is why we added `dialog_common::time`

rust/dialog-storage/src/storage/provider/fs.rs
Gozala: We can pass root which is done through `Directory::At` where's all the other `Directory::` variants are mapped to specific directory through `dirs`. 

I initially used `Directory::At` in tests, but then higher level APIs like operator use `Directory::Profile` internally which was easy to miss causing tests to pollute system.

I have considered env variable or some other way to specify how to interpret `Directory::Profile` and others but in the end it seemed best in tests to simply map all Directories to `/tmp/` avoiding any potential for pollution.

I personally find it to be nice, but open to suggestions how to to do differently. 

P.S.: Directory variants don't share root which is why this enum exists so passing root is not obvious for that reason.

rust/dialog-credentials/src/credential/signer.rs
Gozala: fixed by 1de6574971d7e89189c1aba8c030c145210c5187

rust/dialog-storage/src/storage/provider/fs/archive.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/fs/credential.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/fs/credential.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/fs/error.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/fs/archive.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/fs/memory.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/fs/memory.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/indexeddb/archive.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/fs/memory.rs
Gozala: fixed by switching to tempfile

rust/dialog-storage/src/storage/provider/indexeddb/credential.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/indexeddb/credential.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/indexeddb/credential.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/indexeddb/memory.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/volatile/credential.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/volatile/credential.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/fs.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/fs.rs
Gozala: fixed by adding helper methods instead

rust/dialog-storage/src/storage/provider/fs.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/indexeddb.rs
Gozala: fixed

rust/dialog-storage/src/storage/provider/space.rs
Gozala: This got moved to https://github.com/dialog-db/dialog-db/pull/282/changes#diff-6a0745832243b523f729484ef52624cf1c9b9e25ad0200c9827810529435c447 and I added bunch of comments

rust/dialog-storage/src/storage/provider/fs.rs
Gozala: After discussing this on the call and thinking more about this I came up with solution https://github.com/dialog-db/dialog-db/pull/326

Gozala:

PR #280, opened 2026-04-09
15 ✅ feat: add CertificateStore implementations for all backends

Gozala:
## Summary

The access layer needs to persist delegation certificates so authorization
chains can be verified across sessions. Each backend stores encoded
certificate bytes keyed by audience/subject/issuer: FileSystem as files,
Volatile in a HashMap, IndexedDb using IDBKeyRange prefix queries.

Tests live in dialog-ucan which provides the concrete Ucan protocol type.
open_store() returns the platform backend (FileSystem on native, IndexedDb
on wasm) so the same assertions run everywhere. 20 tests cover the full
prove flow including powerline chains, time bounds, and edge cases.

### DIFF
--- a/rust/dialog-storage/src/storage/provider/fs/certificate.rs
+++ b/rust/dialog-storage/src/storage/provider/fs/certificate.rs
@@ -1,0 +1,124 @@
+//! CertificateStore for filesystem storage.
+//!
+//! Layout: `{space_root}/certificate/{audience}/{subject}/{issuer}.{hash}`
+
+use base58::ToBase58;
+use dialog_capability::access::{
+    AuthorizeError, Certificate, CertificateStore, Delegation, Protocol, Prove, Retain,
+};
+use dialog_capability::{Capability, Policy, Provider};
+use dialog_varsig::Did;
+
+use super::{FileSystem, FileSystemError, FileSystemHandle};
+
+const CERTIFICATE: &str = "certificate";
+
+impl FileSystem {
+    /// Returns the handle for this space's certificate directory.
+    pub fn certificate(&self) -> Result<FileSystemHandle, FileSystemError> {
+        self.resolve(CERTIFICATE)
+    }
+}
+
+#[async_trait::async_trait]
+impl<P: Protocol> CertificateStore<P> for FileSystem
+where
+    P::Certificate: Send + Sync,
+{
+    /// List certificates where `audience` is the recipient and `subject`
+    /// is either the specific subject DID or, when `None`, a
+    /// [powerline](https://github.com/ucan-wg/delegation?tab=readme-ov-file#powerline)
+    /// delegation that applies to any subject.
+    async fn list(
+        &self,
+        audience: &Did,
+        subject: Option<&Did>,
+    ) -> Result<Vec<P::Certificate>, AuthorizeError> {
+        let subject_segment = match subject {
+            Some(did) => did.to_string(),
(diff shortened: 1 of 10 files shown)


rust/dialog-storage/src/storage/provider/fs/certificate.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+#[async_trait::async_trait]
+impl<P: Protocol> CertificateStore<P> for FileSystem
+where
+    P::Certificate: Send + Sync,
+{
+    /// List certificates where `audience` is the recipient and `subject`
+    /// is either the specific subject DID or, when `None`, a
+    /// [powerline](https://github.com/ucan-wg/delegation?tab=readme-ov-file#powerline)
```
cdata, line 30: Given that a "powerline" delegation may be seen as a more powerful scope than specifying some subject, I wonder if it would be more defensive to make the subject a domain-specific enum (rather than `Option<...>`).

rust/dialog-storage/src/storage/provider/volatile/certificate.rs
```diff
@@ -0,0 +1,110 @@
...
+    ///
+    /// Each certificate is stored keyed by
+    /// `{audience}/{subject}/{issuer}.{hash}` where the hash is the
+    /// base58-encoded BLAKE3 hash of the encoded certificate bytes.
+    async fn save(&self, delegation: &P::Delegation) -> Result<(), AuthorizeError> {
+        let mut sessions = self.sessions.write();
+
+        for cert in delegation.certificates() {
+            let bytes = cert.encode()?;
+            let id = base58::ToBase58::to_base58(blake3::hash(&bytes).as_bytes().as_slice());
```
cdata, line 62: You should be able to `use base58::ToBase58` above and then invoke this as a method on any slice of `u8`

rust/dialog-ucan/src/certificate_store.rs
```diff
@@ -0,0 +1,556 @@
+#[cfg(test)]
+mod tests {
+    #[cfg(target_arch = "wasm32")]
+    wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_dedicated_worker);
+
+    use dialog_capability::access::{Certificate, CertificateStore};
+    use dialog_credentials::Ed25519Signer;
+    use dialog_effects::storage::{Directory, Location};
+    use dialog_storage::resource::Resource;
```
cdata, line 9: IIUC this constitutes a circular dependency (via `dev-dependencies`, so technically not a problem) between `dialog-ucan` and `dialog-storage`. There may not be a better way, but I tend to try to avoid these kinds of cycles even via `dev-dependencies`.

cdata (approved):
LGTM

rust/dialog-ucan/src/certificate_store.rs
Gozala: I think only real way to break the cycle here is create separate crate that depends on `dialog-ucan` and `dialog-storage` in order to test interop. crate just for tests seemed like a too much, so it kind of ended up with this admittedly annoying situation.

Perhaps you have recommendation for addressing this kind of situation ?

rust/dialog-storage/src/storage/provider/volatile/certificate.rs
Gozala: fixed

Gozala:

PR #289, opened 2026-04-09
17 ✅ feat: adapt artifacts to capability system

Gozala:
All of the new abstractions work with borrowed capability provider, where's query engine index storage. Changes here introduce `Select` capability and `ArtifactStream` with lifetimes that enables query engine to receive reference to `Provider<Select>` and produce `ArtifactStream` overcoming a need for the owned index storage.

PR also moves some pieces from the [query engines transactor](https://github.com/dialog-db/dialog-db/blob/main/rust/dialog-query/src/session/transaction.rs) in preparation for exposing transaction API in #294 to avoid dependency cycle between `dialog-query` and `dialog-repository` allowing both to import it from `dialog-artifacts`.

P.S.: I would like to repurpose / rename `dialog-artifacts` to either `dialog-facts` or `dialog-claims` or something like that so it would hold definitions for value types supported by the [associative layer](https://gozala.io/dialog/anatomy#associative-memory), but I'm deferring that to the future change


cdata (approved):
Soft preference for `dialog-facts`, that is what I originally called this crate before you suggested `dialog-artifacts`. I get what you're going for with `dialog-claims`, which seems more pedantically correct as far as names go. I'm not sure how important it is to connect this with the corpus of prior art (which naively I assume uses "facts" over "claims," but I don't really know).

Gozala:
> I'm not sure how important it is to connect this with the corpus of prior art (which naively I assume uses "facts" over "claims," but I don't really know).

I got swayed 

> Soft preference for `dialog-facts`, that is what I originally called this crate before you suggested `dialog-artifacts`. I get what you're going for with `dialog-claims`, which seems more pedantically correct as far as names go. I'm not sure how important it is to connect this with the corpus of prior art (which naively I assume uses "facts" over "claims," but I don't really know).

@bumbleblue made me think that "fact" is [a misleading term](https://github.com/tonk-labs/rfc/pull/6#issuecomment-3896237382) and adopted "claim" instead. That said I have had second thoughts since mainly because prior art uses term "fact" there and it feels like a pain to go against it.

Gozala:

PR #294, opened 2026-04-09
24 📬 feat: adapt query engine and add repository query integration

Gozala:
## Why

The query engine dispatched through a Source trait that directly accessed artifact stores. With the capability system, all data access routes through Provider<Select> on the operator. The query macros (Formula, Concept) need to generate code that works with this provider-based dispatch instead.

## Overview

Updates dialog-query to use capability-based dispatch:
- Application::evaluate takes a generic Env with Provider<Select> + Provider<SelectRules> bounds instead of a Source
- Formula/Concept derive macros updated to generate matching signatures with proper lifetime bounds
- TestEnv in source.rs uses Operator<VolatileSpace> for tests

Adds query integration to dialog-repository:
- session.rs: transaction and query builder API for Branch
- branch.query().select(Query).perform(&operator) for typed concept queries
- branch.transaction().assert(concept).commit().perform(&operator) for typed writes
- branch.query().install(rule).select(Query) for deductive rule queries

Updates dialog-credentials with credential module re-exports.


Gozala:

PR #297, opened 2026-04-13
25 📬 feat: add pluggable import/export for artifacts with CSV support

Gozala:
Add Exporter/Importer traits to dialog-artifacts for format-agnostic artifact serialization. Exporter uses write/close, Importer is a Stream.

Add dialog-csv crate implementing CSV format with columns: the, of, as (value type), is (value), cause. Value types use human-readable names (text, natural, integer, boolean, float, bytes, entity, attribute, record).

Add Branch::export/import methods to dialog-repository that wire the traits into the capability-based commit pipeline.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1018,6 +1018,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "dialog-csv"
+version = "0.1.0"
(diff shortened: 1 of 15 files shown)


Gozala:

PR #302, opened 2026-04-15
feat: adapt query engine and add repository query integration

Gozala:
The query engine used a Source trait for data access. With the capability system, queries dispatch through Provider<Select> and Provider<SelectRules> instead. The Formula/Concept derive macros are updated to generate code matching the new Application trait signature with lifetime-bounded environments.

Adds session.rs back to dialog-repository with transaction/query builder API for typed concepts and deductive rules. Updates dialog-credentials with credential module re-exports.


Gozala:
Superseded — force-pushed onto #294.

PR #303, opened 2026-04-18
🔁 chore: remove S3 backend, network provider, and dialog-s3-credentials

Gozala:
Parallel-chain rebase of #246 onto the updated #245 (which now has the Claim→Attenuate rename). No content changes vs #246; just a rebase to validate the chain.

Will replace #246 once parallel chain is validated end-to-end.


Gozala:
Superseded — force-pushed onto #246.

PR #304, opened 2026-04-18
🔁 feat: add #[derive(Provider)] compose macro

Gozala:
Parallel-chain rebase of #252 onto the updated #246 (which carries the #245 Claim→Attenuate rename forward). Trivial conflict in dialog-macros/src/lib.rs (mod listing) resolved.

Will replace #252 once parallel chain is validated end-to-end.


Gozala:
Superseded — force-pushed onto #252.

PR #305, opened 2026-04-18
🔁 feat: add credential newtypes and key export improvements

Gozala:
Parallel-chain rebase of #251 onto the updated #252. No content changes vs #251.


Gozala:
Superseded — force-pushed onto #251.

PR #306, opened 2026-04-18
🔁 feat: redesign access capabilities

Gozala:
Parallel-chain rebase of #249 with one additional commit on top: rename the Claim trait/derive to Attenuate (carrying forward the rename from the Attenuate parallel chain).

Will replace #249 once parallel chain is validated end-to-end.


Gozala:
Superseded — force-pushed onto #249.

PR #267, opened 2026-04-08
10 ✅ feat: add effects hierarchy modules

Gozala:
## Why

Effects previously re-exported storage KV types (Get/Set/Delete/List) from a flat module. The new storage design uses location-based bootstrap operations (load/create a space at a directory + name), and the operator layer needs space-level name resolution, credential persistence, and identity effects. These capabilities had no home.

## Overview

Replaces the old storage KV re-exports with location-based operations and adds the missing capability domains:

- storage: Directory/Location/Load/Create for bootstrap. LocationExt sugar for the capability chain. Storage::profile()/current()/temp() convenience builders.
- space: Space Load/Create for operator-level name resolution (resolves names against a base directory)
- authority: identity chain (Profile with DID + optional account, Operator with session DID, Identify effect)
- credential: Load/Save effects for credential persistence with CredentialError
- access: re-exports from dialog-capability::access for convenience
- archive/memory preludes: fluent builder extension traits (SubjectExt, ArchiveExt, CatalogExt, etc.)

PR #307, opened 2026-04-18
🔁 feat: add effects hierarchy modules

Gozala:
Parallel-chain rebase of #267 on top of #249-v2. Conflicts in archive.rs, memory.rs, storage.rs (taken from incoming, then re-renamed Claim->Attenuate).


Gozala:
Superseded — force-pushed onto #267.

PR #312, opened 2026-04-19
🔁 feat: add CertificateStore implementations for all backends

Gozala:
## Summary

The access layer needs to persist delegation certificates so authorization
chains can be verified across sessions. Each backend stores encoded
certificate bytes keyed by audience/subject/issuer: FileSystem as files,
Volatile in a HashMap, IndexedDb using IDBKeyRange prefix queries.

Tests live in dialog-ucan which provides the concrete Ucan protocol type.
open_store() returns the platform backend (FileSystem on native, IndexedDb
on wasm) so the same assertions run everywhere. 20 tests cover the full
prove flow including powerline chains, time bounds, and edge cases.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1188,6 +1188,7 @@ dependencies = [
  "dialog-common",
  "dialog-credentials",
  "dialog-effects",
+ "dialog-storage",
  "dialog-ucan-core",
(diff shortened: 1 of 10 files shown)


Gozala:
Superseded — force-pushed onto #280.

PR #282, opened 2026-04-09
16 ✅ feat: add Storage runtime with DID-routed dispatch

Gozala:
## Summary

Spaces need a runtime that routes capability effects to the correct
backend by subject DID and handles space lifecycle (load existing or
create new at a given location with credential persistence).

`Storage<S>` composes Router (blanket Provider dispatch by subject DID)
with Loader (storage::Load/Create with location-to-DID caching).
Platform type aliases and Default impls live in cfg-gated submodules
so the same Storage API works on native (FileSystem) and web (IndexedDb).

PR #313, opened 2026-04-19
🔁 feat: add Storage runtime with DID-routed dispatch

Gozala:
## Summary

`Storage<S>` composes `Router` (blanket `Provider` dispatch by subject DID) with `Loader` (storage::Load/Create handling with location caching). When a space is loaded or created, `Loader` opens the backend, saves or loads the credential, registers the DID-to-store mapping, and `Router` dispatches all subsequent effects to the correct store.

`Space` groups the four per-DID storage concerns (archive, memory, credential, certificate) behind a single composed provider. Each field is independently typed so different concerns can be backed by different implementations (e.g. OPFS for archive, IndexedDB for memory). The module doc walks through the repository/profile framing, the mix-and-match composition, and how `Storage` hides the composition from consumers so they can stay in capability-land — the space's DID is the root of any capability chain, and the runtime routes effects to the right backend without consumers needing to think about it.

Platform type aliases (`NativeSpace`, `WebSpace`) and `Default` impls live in cfg-gated `native.rs`/`web.rs` submodules. `Storage::volatile()` provides an in-memory instance for testing.


Gozala:
Superseded — force-pushed onto #282.

PR #314, opened 2026-04-19
🔁 feat: adapt artifacts to capability system

Gozala:
## Summary

Adds a `Select` `Command` that yields a boxed `ArtifactStream` tied to the provider's lifetime, so query results can borrow from the environment rather than always being `'static`.

Adds an `Update` trait with `associate` / `dissociate` / `associate_unique` methods plus a `Statement` trait so domain types can describe assertions and retractions. `Changes` batches pending writes by entity and attribute and exposes them as either a `Vec<Instruction>` or a `ChangeStream` for the existing commit path.

Wires `From<ArchiveError>` / `From<MemoryError>` into `DialogArtifactsError` so providers from `dialog-effects` can flow through.


Gozala:
Superseded — force-pushed onto #289.

PR #290, opened 2026-04-09
18 ✅ feat: add dialog-remote-s3 crate

Gozala:
## Summary

Extracts S3 storage backend from the feature-gated code removed in #246 into a standalone `dialog-remote-s3` crate. This separates the S3 concerns (SigV4 signing, bucket operations, fork invocations) from the local storage layer. It also removes need for various feature flags avoid complex matrix to navigate.

PR #315, opened 2026-04-19
🔁 feat: add dialog-remote-s3 crate

Gozala:
## Summary

Extracts the S3 storage backend from the feature-gated code removed in #246 into a standalone `dialog-remote-s3` crate. This separates the S3 concerns (SigV4 signing, bucket operations, fork invocations) from the local storage layer and removes the need for the various feature flags / matrices that came with it.

The first 25 commits replay the original PR #290 history rebased onto `feat/artifacts-adaptation-v2`. The trailing `chore: sync …` commit brings the crate forward to the cleanup-branch state so it compiles against the v3 stack: the `capability/` module is renamed to `request/` (with `IntoRequest` blanket alias and per-capability `From<&Capability<Fx>>` impls), the `*Capability` ext traits are replaced with the v3's `*Ext` traits, qualified `dialog_capability::{site,fork,command}::` paths collapse to the re-exports, `S3Authorization` captures the `Request` at authorize-time, and `Secret` is constructed/inspected via its public API.

#290 review feedback is addressed by the cleanup state: `Version::from(String)` replaces the lossy bytes conversion in the memory provider; the missing-ETag case errors via `?` rather than panicking; and `s3/credential.rs` no longer panics. The integration-tests workflow already uses the nix flake pattern.


Gozala:
Superseded — force-pushed onto #290.

PR #317, opened 2026-04-19
✅ 20 feat: dialog-network providing network capability composition

Gozala:
## Summary

Adds the `#[derive(Site)]` proc macro and uses it to extract the composite `Network` site (S3 + UCAN-over-S3 dispatch) into its own `dialog-network` crate. This lets `dialog-operator` depend on the abstraction without pulling the concrete transport crates as regular deps.

`#[derive(Site)]` takes a struct of site fields and generates the address enum, authorization enum, fork enum, the `Site` impl, the `SiteAddress` back-pointer, the `From<S::Address>` adapters, and the `Provider<ForkInvocation>` dispatch impl. cfg-gated fields are rejected for now (would require marker-trait routing for per-variant where-clause bounds). Drops the obsolete `Router` derive while we're here.

`dialog-network` itself is just `struct Network { s3: S3, ucan: UcanSite }` plus the derive — all the boilerplate the macro now writes for you.


rust/dialog-macros/src/site.rs
```diff
@@ -0,0 +1,490 @@
...
+            ) -> <Fx as ::dialog_capability::Effect>::Output {
+                let ::dialog_capability::ForkInvocation {
+                    capability,
+                    address,
+                    authorization,
+                } = input;
+                match (authorization, address) {
+                    #(#dispatch_arms)*
+                    _ => unreachable!("authorization/address type mismatch"),
+                }
```
jackddouglas, line 486: seems like this is the fault of how `ForkInvocation` is constructed. something like `Authorized` could pair authorization/address together and have stronger type enforcement

```rs
pub trait Site {
    type Address: SiteAddress;
    type Authorized; // paired address+authorization
    type Fork<Fx: Effect>;
}

pub struct ForkInvocation<S: Site, Fx: Effect> {
    pub(crate) capability: Capability<Fx>,
    pub(crate) authorized: S::Authorized,
}
```

jackddouglas (approved):

Gozala:

PR #293, opened 2026-04-09
23 🏗️ feat: add dialog-repository crate

Gozala:
## Why

The repository is the primary user-facing abstraction: a git-like container for structured data with branches, remotes, and merge. Without it, consumers have to wire up the operator, storage, and artifact layers themselves.

## Overview

New crate providing the repository layer:

- Repository<C>: credential-parameterized wrapper. Repository<SignerCredential> can delegate access, Repository<Credential> is read-only or read-write depending on the credential. Opened via profile.repository(name).open()/load()/create().
- Branch: named branch with revision tracking, commit/select/push/pull/fetch/merge/reset operations. Claims select via artifact selectors with automatic remote fallback for missing blocks.
- Remote: named remote endpoints with S3 or UCAN site addresses. Remote branches for upstream tracking.
- Archive: content-addressed storage adapter bridging capabilities with prolly tree storage. FallbackStore transparently fetches missing blocks from remote on cache miss.
- Memory: capability-scoped transactional memory cells with edition tracking.
- 52 unit tests covering repository open/create/load, branch operations, commits, selects, retractions, remotes, and UCAN delegation.

Query integration (transaction/session API with typed concepts and deductive rules) is deferred to a later PR with dialog-query adaptation.

PR #319, opened 2026-04-20
🔁 feat: add dialog-repository crate

Gozala:
## Summary

Adds the `dialog-repository` crate — the user-facing abstraction on top of Operator + Storage. A Repository owns a Branch (the local writable head) and a set of Remotes (addressable peers it can push/pull/fetch from). Local archive + memory adapters route through the Operator's capability dispatch; networked archive uses a fallback store for fetching missing blocks from remotes.

Branch operations include commit, select, fetch, publish, retract, reset, and upstream tracking. Remote types cover address, branch, repository, name, reference, and revision. Helpers include a credentials roundtrip fixture and integration test setup.

The crate ports straight from the cleanup branch's final state in a single commit (rather than replaying the original PR #293's intermediate refactors, which used APIs that didn't survive the cleanup). Also adds `From<ArchiveError>` / `From<MemoryError>` impls to `DialogStorageError` so the repository's local adapters can use `?` against the v3 stack's effect errors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1261,6 +1261,43 @@ dependencies = [
 ]
 
 [[package]]
+name = "dialog-repository"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "async-stream",
+ "async-trait",
+ "base58",
+ "dialog-artifacts",
+ "dialog-capability",
+ "dialog-common",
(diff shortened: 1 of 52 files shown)


Gozala:
Superseded — force-pushed onto #293.

PR #327, opened 2026-04-22
fix: synthesize docs on Concept/Formula derive-generated items

Gozala:
The `#[derive(Concept)]` and `#[derive(Formula)]` macros emitted `pub` items (mirror structs, terms accessors, field positions) with no `#[doc]` attribute, which tripped `#![warn(missing_docs)]` in consumer crates on code they did not author.

Synthesize fallback doc attributes on every generated item, forwarding the user's field-level docs when provided. The user's original struct is left untouched so rustc still warns them for their own missing docs.

Adds `rust/dialog-query/tests/missing_docs.rs` as a regression guard: compiled under `#![deny(missing_docs)]`, it stops building if a future macro change emits a new undocumented `pub` item. Exercises `Concept` mirror types plus the three `Attribute` domain paths (derived from module path, string literal, identifier).

### DIFF
--- a/rust/dialog-macros/src/query/concept.rs
+++ b/rust/dialog-macros/src/query/concept.rs
@@ -156,9 +156,20 @@ pub fn derive(input: TokenStream) -> TokenStream {
         field_types.push(field_type);
         field_name_lits.push(field_name_lit.clone());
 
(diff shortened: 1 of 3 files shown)


Gozala:

PR #328, opened 2026-04-23
feat: add From<Profile> for Repository<SignerCredential>

Gozala:
Lets a Profile be treated as its own repository whose subject is the profile DID. Because the operator's space providers gate on the profile DID and the Loader registers the profile DID at Directory::Profile during Profile::open, repo operations route to the profile's storage root (~/Library/Application Support/dialog/<name>/ on native, the <name>.profile IndexedDB on web).


Gozala:

PR #329, opened 2026-04-23
fix(dialog-ucan): preserve chain order in `UcanProof::from_chain`

jackddouglas:
`UcanProof::from_chain` built its `proofs: Vec<UcanCertificate>` by iterating `chain.delegations().values()` which is unordered due to the HashMap. `Proof::claim` then rebuilt the delegation chain by calling `DelegationChain::new(first)` followed by `chain.push(next)` for each subsequent proof, and `push` rejects anything where the new delegation's issuer doesn't match the current chain's leaf audience.

For single-hop chains this was doesn't matter because there's only one element but for multi-hop chains it was more random, depending on however the entries got hashed. Since `RandomState` is seeded once per process, I could fail to claim an invite once, wipe the state and try again, then it works.

Simple fix by iterating `DelegationChain::proof_cids` instead, added regression tests as well.

### DIFF
--- a/rust/dialog-ucan-core/src/container/delegation.rs
+++ b/rust/dialog-ucan-core/src/container/delegation.rs
@@ -103,9 +103,29 @@ impl DelegationChain {
         &self.proof_cids
     }
 
-    /// Get the delegations map for building InvocationChain.
-    pub fn delegations(&self) -> &HashMap<Cid, Arc<Delegation<Ed25519Signature>>> {
-        &self.delegations
+    /// Iterate the chain's delegations in root-to-leaf order.
+    pub fn proofs(&self) -> impl Iterator<Item = &Delegation<Ed25519Signature>> {
+        self.proof_cids
+            .iter()
+            .map(|cid| self.proof_for_cid(cid).as_ref())
+    }
+
+    /// Iterate `(cid, arc)` pairs in root-to-leaf order, cloning the
+    /// inner `Arc`s.
+    ///
+    /// For constructing a delegation lookup table — most commonly to
+    /// hand to `InvocationChain::new`. Collect into a `HashMap` at
+    /// the call site.
+    pub fn export(&self) -> impl Iterator<Item = (Cid, Arc<Delegation<Ed25519Signature>>)> + '_ {
+        self.proof_cids
+            .iter()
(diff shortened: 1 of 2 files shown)


rust/dialog-ucan/src/access.rs
```diff
@@ -103,10 +103,22 @@ impl UcanProof {
...
+    /// alignment error — so the iteration order here has to match
+    /// the canonical order the chain documents.
     pub fn from_chain(chain: &DelegationChain, scope: Scope) -> Self {
+        let delegations = chain.delegations();
         let proofs = chain
-            .delegations()
-            .values()
+            .proof_cids()
+            .iter()
+            .filter_map(|cid| delegations.get(cid))
```
Gozala: Can we instead fix this at the `DelegationChain` level e.g. maybe it could have proofs iterator so method. I'm also starting to think that perhaps we should loose delegations method (can't recall if there are good reasons to have it) as it seems like a foot gun. If we do need `delegations` maybe we could at least add warning in the method docs calling out to not to do this.

P.S. filter_map could have a comment explaining that we're guaranteed to have delegation for every cid by construction. Actually maybe it should not be filter map but rather utilize `unreachable!`.

Gozala (changes requested):
Thanks for the fix!

rust/dialog-ucan/src/access.rs
jackddouglas: `access.rs` needs `delegations()` to create `InvocationChain` so we can just steer people to `proofs()` instead

rust/dialog-ucan-core/src/container/delegation.rs
```diff
@@ -103,9 +103,31 @@ impl DelegationChain {
...
+            .map(|cid| self.proof_for_cid(cid).as_ref())
+    }
+
+    /// Iterate `(cid, arc)` pairs in root-to-leaf order, cloning the
+    /// inner `Arc`s.
+    ///
+    /// For constructing a delegation lookup table — most commonly to
+    /// hand to `InvocationChain::new`. Collect into a `HashMap` at
+    /// the call site.
+    pub fn proof_arcs(
```
Gozala: Nit: I'd probably call it `export` instead

Gozala (approved):
Looks good thanks!

PR #331, opened 2026-04-24
fix: ucan proof chain order

Gozala:
Restores #329


Gozala:

PR #332, opened 2026-04-24
fix: Search tree ranking and deletion

cdata:
This fix addresses two major issues with the new search tree.

1. Ranking was incorrectly adapted in the first pass. The diff reveals the discrepency, but the short explanation is that we assumed all keys in a grouping share the same rank as the boundary and everything past that was built on an incorrect assumption.
2. When deleting a boundary, we were not collapsing nodes to properly reparent the entries of the boundary's segment.

Both of these issues are addressed in this change. Additionally:

- Replace `Tree::get_node` with `Accessor` which serves as a `Sync` shared lookup cache for nodes
- Introduce localized caching of key ranks (there is headroom for further improvement down the road)
- Added a bunch of missing test coverage for the tree and other constructs

### DIFF
--- a/rust/dialog-search-tree/src/shaper.rs
+++ b/rust/dialog-search-tree/src/shaper.rs
@@ -5,9 +5,10 @@
 //! from the read-only [`Tree`] interface, we achieve clearer separation of
 //! concerns and make the codebase more maintainable.
 
-use std::marker::PhantomData;
+use std::{collections::VecDeque, marker::PhantomData};
 
 use dialog_common::Blake3Hash;
+use hashbrown::HashMap;
 use nonempty::NonEmpty;
 use rkyv::{
     Deserialize, Serialize,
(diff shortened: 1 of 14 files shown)


rust/dialog-search-tree/src/shaper.rs
```diff
@@ -198,68 +250,226 @@ where
...
-            ));
+        // Boundary-delete with non-empty orphans and a right-adjacent segment
+        // triggers overflow: the orphans must fold into the next segment so
+        // the resulting tree matches a from-scratch build.
+        if deleted_boundary
+            && !segment.entries.is_empty()
+            && let Some(right_neighbor) = search_result.right_neighbor
+        {
+            let main_leaf_hash = search_result.leaf.hash().clone();
+            return self.absorb_orphans_into_right_neighbor(
```
Gozala: > Seems like it should be adopting orphans not absorbing them 😅

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -198,68 +250,226 @@ where
...
+                segment.entries,
+                main_leaf_hash,
+                search_result.path,
+                right_neighbor,
+            );
+        }
 
-            let mut delta = self.delta.branch();
-            delta.subtract(search_result.leaf.hash());
+        match NonEmpty::from_vec(segment.entries) {
```
Gozala, line 303: Maybe add comment here, I'm bit lost on what is going on here. Did not we let right neighbor adopt these entries already ?

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -198,68 +250,226 @@ where
...
-                delta,
+    /// Resolves a boundary-delete overflow using the prefetched right-adjacent
+    /// segment.
+    ///
+    /// See case (3) in [`Self::delete`] for the high-level contract.
+    ///
+    /// Key invariant: when a rank-R entry is deleted, it was simultaneously a
+    /// boundary at levels 0..R-2, and each of those boundaries dissolves. At
+    /// every level below the LCA, the main subtree's rebuild and the
+    /// right-adjacent subtree's rebuild therefore fuse into a *single* index
```
Gozala: index node I assume right ?

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -198,68 +250,226 @@ where
...
-                search_result.path,
-                delta,
+    /// Resolves a boundary-delete overflow using the prefetched right-adjacent
+    /// segment.
+    ///
+    /// See case (3) in [`Self::delete`] for the high-level contract.
+    ///
+    /// Key invariant: when a rank-R entry is deleted, it was simultaneously a
+    /// boundary at levels 0..R-2, and each of those boundaries dissolves. At
+    /// every level below the LCA, the main subtree's rebuild and the
```
Gozala: What is LCA ?

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -198,68 +250,226 @@ where
...
+    /// and handed off to the standard merge routine.
+    ///
+    /// By construction of the search (main descent follows the boundary
+    /// rightward at every level; the right-adjacent descent takes leftmost
+    /// children) we can rely on:
+    /// - `main_layer.right_siblings == None` for every layer below the LCA, and
+    /// - `right_layer.left_siblings == None` for every layer below the LCA.
+    ///
+    /// The fold at each level is therefore just
+    /// `[main_layer.left_siblings | unified | right_layer.right_siblings]`.
```
Gozala: I have a hard time following this, I think some asci visuals would really help here

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -198,68 +250,226 @@ where
...
+        })?;
+        let above_lca = above_and_lca;
+
+        // Fuse the two subtrees level-by-level below the LCA. At every such
+        // level the boundary that separated them has dissolved, so their
+        // children flatten into one unified index.
+        //
+        // Nodes live at level 0 after the entry merge; building a level-1
+        // parent uses minimum_rank 2 (level L uses minimum_rank L + 1).
+        let mut level_minimum_rank: Rank = 2;
```
Gozala: Maybe 2 should be a named constant

rust/dialog-search-tree/src/shaper.rs
```diff
@@ -402,33 +620,27 @@ where
...
+        let ranked_links = into_ranked_links(links, &mut context);
 
         // Create nodes from the remaining links
         let nodes = Self::collect::<Link<_>>(ranked_links, 1)?;
 
-        // Merge up the remaining path
-        Self::merge_with_path(nodes, path, delta)
+        // Merge up the remaining path. `merge_with_path` pops from the back,
+        // so hand it back a `Vec` of the still-pending ancestors.
+        Self::merge_with_path(nodes, path.into(), context, 2)
```
Gozala: Seems like 2 should be a constant

Gozala (approved):
Provided some feedback, I did not caught any issues but have to admit I had a hard time following the code and I still don't know what LCA is 😅 All that said I'm very excited and all the added tests further reassure that it should behave as expected.

I would however prefer if this could wait and land after large PR stack lands as I think it would be pretty straightforward to rebase this over it as there almost no overlap and it would save me from having to rebase 20+ prs

PR #333, opened 2026-05-12
feat: cardinality one assertions

Gozala:
## Summary

Cardinality-one assertions did not actually supersede prior values across commits. Originally the design imagined callers would populate `cause` on a new assertion to point at the prior fact they intended to retract — and supersession would key off that. In practice you often *don't* know the prior value at write time, so nothing ever set `cause`, and cardinality-one wasn't behaving as cardinality-one. I keep forgetting this myself and getting surprised by it; the cardinality declaration on the attribute is a better expression of intent than caller-supplied `cause`.

This PR makes `cardinality: one` actually mean "this attribute holds one value per entity," which also lines up more naturally with the planned [version-control design](https://github.com/dialog-db/dialog-db/pull/240/changes) — where `Cause` becomes a `Vec<Version>` of *automatically captured* superseded priors rather than something the caller hand-fills.

## What changed

Add `Change::Replace` / `Instruction::Replace` as the single signal for cardinality-one supersession. `Statement::assert` already routes cardinality-one writes through `Update::associate_unique`, which now lowers to `Replace`. Both commit handlers (`dialog-repository/src/repository/branch/commit.rs` and the in-memory store in `dialog-artifacts/src/artifacts.rs`) implement `Replace` by scanning `(entity, attribute)`, deleting every prior whose value differs, and skipping the insert when a same-value prior is already present (so reasserting the same value doesn't change the tree hash).

`Instruction::Assert` is now purely additive. The dead `cause`-gated branch is gone, along with `Artifact::update` and the WASM `Artifact.update(value)` closure that depended on it. `Artifact.cause` is left as a vestigial field; the version-control work will repurpose it.

## Tests

- `it_resolves_only_latest_name_target_via_name_concept` — was failing; now passes.
- `it_accumulates_two_cardinality_many_values_in_one_tx` — regression guard; cardinality-many still accumulates.
- `it_is_noop_when_reasserting_same_cardinality_one_value` — `r1.tree == r2.tree` on identical reassert.

### DIFF
--- a/rust/dialog-artifacts/src/artifacts.rs
+++ b/rust/dialog-artifacts/src/artifacts.rs
@@ -441,60 +441,85 @@ where
 
                         let datum = Datum::from(artifact);
 
-                        if let Some(cause) = &datum.cause {
-                            let ancestor_key = {
-                                let search_start = <EntityKey<Key> as KeyViewConstruct>::min()
-                                    .set_entity(entity_key.entity())
-                                    .set_attribute(entity_key.attribute())
-                                    .into_key();
-                                let search_end = <EntityKey<Key> as KeyViewConstruct>::max()
-                                    .set_entity(entity_key.entity())
-                                    .set_attribute(entity_key.attribute())
-                                    .into_key();
+                        // TODO: Make it concurrent / parallel
+                        index
+                            .set(
+                                entity_key.into_key(),
+                                State::Added(datum.clone()),
+                                &mut self.storage,
+                            )
+                            .await?;
+                        index
+                            .set(
+                                attribute_key.into_key(),
+                                State::Added(datum.clone()),
+                                &mut self.storage,
+                            )
+                            .await?;
+                        index
+                            .set(value_key.into_key(), State::Added(datum), &mut self.storage)
+                            .await?;
+                    }
+                    Instruction::Replace(artifact) => {
+                        let entity_key = EntityKey::from(&artifact);
 
-                                let search_stream =
-                                    index.stream_range(search_start..search_end, &self.storage);
+                        // Scan priors at this (entity, attribute). Same-valued
+                        // priors already represent the desired state; only
+                        // different-valued ones need superseding.
+                        let mut superseded_keys: Vec<Key> = Vec::new();
+                        let mut found_same_value = false;
+                        {
+                            let search_start = <EntityKey<Key> as KeyViewConstruct>::min()
+                                .set_entity(entity_key.entity())
+                                .set_attribute(entity_key.attribute())
+                                .into_key();
+                            let search_end = <EntityKey<Key> as KeyViewConstruct>::max()
+                                .set_entity(entity_key.entity())
+                                .set_attribute(entity_key.attribute())
+                                .into_key();
 
-                                let mut ancestor_key = None;
+                            let search_stream =
+                                index.stream_range(search_start..search_end, &self.storage);
+                            tokio::pin!(search_stream);
 
-                                tokio::pin!(search_stream);
-
-                                while let Some(candidate) = search_stream.try_next().await? {
-                                    if let State::Added(current_element) = candidate.value {
-                                        let current_artifact = Artifact::try_from(current_element)?;
-                                        let current_artifact_reference =
-                                            Cause::from(&current_artifact);
-
-                                        if cause == &current_artifact_reference {
-                                            ancestor_key = Some(candidate.key);
-                                            break;
-                                        }
+                            while let Some(candidate) = search_stream.try_next().await? {
+                                if let State::Added(current_element) = candidate.value {
+                                    let current = Artifact::try_from(current_element)?;
+                                    if current.is == artifact.is {
+                                        found_same_value = true;
+                                    } else {
+                                        superseded_keys.push(candidate.key);
                                     }
                                 }
-
-                                ancestor_key
-                            };
-
-                            if let Some(key) = ancestor_key {
-                                // Prune the old entry from the indexes
-                                let entity_key = EntityKey(key);
-                                let value_key = ValueKey::from_key(&entity_key);
-                                let attribute_key = AttributeKey::from_key(&entity_key);
-
-                                // TODO: Make it concurrent / parallel
-                                index
-                                    .delete(&entity_key.into_key(), &mut self.storage)
-                                    .await?;
-                                index
-                                    .delete(&value_key.into_key(), &mut self.storage)
-                                    .await?;
-                                index
-                                    .delete(&attribute_key.into_key(), &mut self.storage)
-                                    .await?;
                             }
                         }
 
-                        // TODO: Make it concurrent / parallel
+                        for key in superseded_keys {
+                            let entity_key = EntityKey(key);
+                            let value_key = ValueKey::from_key(&entity_key);
+                            let attribute_key = AttributeKey::from_key(&entity_key);
+
+                            index
+                                .delete(&entity_key.into_key(), &mut self.storage)
+                                .await?;
+                            index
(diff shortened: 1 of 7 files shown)


rust/dialog-artifacts/src/artifacts/artifact.rs
```diff
@@ -41,19 +41,6 @@ pub struct Artifact {
     pub cause: Option<Cause>,
 }
 
-impl Artifact {
-    /// Change the value of the [`Artifact`], assigning the hash of its
-    /// antecedent as the `cause`.
-    pub fn update(self, value: Value) -> Self {
-        let cause = Some(Cause::from(&self));
```
jackddouglas, line 48: `cause` is dangling now and isn't updated anywhere. it still makes sense to me to update for causality so perhaps `Replace` should do that?

rust/dialog-artifacts/src/artifacts/instruction.rs
```diff
@@ -10,8 +10,16 @@ use crate::ArtifactStoreMut;
...
 pub enum Instruction {
-    /// Assert a [`Artifact`], persisting it in the [`ArtifactStoreMut`]
+    /// Add this [`Artifact`] to the [`ArtifactStoreMut`]. Purely additive:
+    /// any prior entries at the same `(entity, attribute)` are left in
+    /// place. Use [`Instruction::Replace`] for cardinality-one supersession.
     Assert(Artifact),
+    /// Replace any prior artifact at the same `(entity, attribute)` with this
+    /// one, regardless of value. Every existing entry for the pair is
+    /// removed from all three indexes; the new artifact's `cause` is
+    /// populated from the first prior found. Asserting the same value that
```
jackddouglas, line 20: this isn't true but probably should be

rust/dialog-artifacts/src/artifacts/instruction.rs
```diff
@@ -10,8 +10,16 @@ use crate::ArtifactStoreMut;
...
+    /// already exists is a no-op.
+    Replace(Artifact),
```
jackddouglas, line 22: `Artifact` could include `cause` which `Replace` doesn't currently use. either scope should be reduced so `cause` isn't silently dropped or we use it

rust/dialog-artifacts/src/artifacts/update.rs
jackddouglas, line 32: maybe `cardinality-one` could be enforced stronger at the type level because if an `Update` impl doesn't override this `cardinality-one` is no longer true

rust/dialog-artifacts/src/artifacts.rs
```diff
@@ -441,60 +441,85 @@ where
...
+
+                            while let Some(candidate) = search_stream.try_next().await? {
+                                if let State::Added(current_element) = candidate.value {
+                                    let current = Artifact::try_from(current_element)?;
+                                    if current.is == artifact.is {
+                                        found_same_value = true;
+                                    } else {
+                                        superseded_keys.push(candidate.key);
                                     }
                                 }
```
jackddouglas, line 493: could be a helper method

rust/dialog-repository/src/repository/branch/commit.rs
```diff
@@ -96,53 +96,74 @@ where
...
+                            if let State::Added(current_element) = candidate.value {
+                                let current = Artifact::try_from(current_element)?;
+                                if current.is == artifact.is {
+                                    found_same_value = true;
+                                } else {
+                                    superseded_keys.push(candidate.key);
                                 }
                             }
+                        }
+                    }
```
jackddouglas, line 145: yeah this logic should be extracted and shared with `artifact.rs`

rust/dialog-artifacts/src/artifacts.rs
```diff
@@ -441,60 +441,85 @@ where
...
+                            let search_stream =
+                                index.stream_range(search_start..search_end, &self.storage);
+                            tokio::pin!(search_stream);
+
+                            while let Some(candidate) = search_stream.try_next().await? {
+                                if let State::Added(current_element) = candidate.value {
+                                    let current = Artifact::try_from(current_element)?;
+                                    if current.is == artifact.is {
+                                        found_same_value = true;
+                                    } else {
```
jackddouglas, line 490: so two entities with identical values but different causal histories are identical? maybe better to drop `cause` entirely if so?

rust/dialog-artifacts/src/artifacts.rs
```diff
@@ -441,60 +441,85 @@ where
...
+                            index
+                                .delete(&entity_key.into_key(), &mut self.storage)
+                                .await?;
+                            index
+                                .delete(&value_key.into_key(), &mut self.storage)
+                                .await?;
+                            index
+                                .delete(&attribute_key.into_key(), &mut self.storage)
+                                .await?;
+                        }
```
jackddouglas, line 511: claude has this to say:

> Each Replace runs 3*N independent await? deletes then 3 sets. A failure in any later op leaves the three indexes (entity/attribute/value permanently desynchronized - entity index gone while value index still points at the superseded datum produce phantom query results forever. Assert is structurally identical (3 sets) and has the same exposure; Replace amplifies it. No rollback, no compensation, no log identifying which keys were partially mutated. At minimum, log the partial-failure boundary; ideally apply atomically.

so the op could partially fail and leave mixed, out-of-sync state. can this be done atomically?

rust/dialog-artifacts/src/artifacts.rs
```diff
@@ -441,60 +441,85 @@ where
...
+
+                        // Skip insert if a same-value prior is already in place.
+                        if found_same_value {
+                            continue;
+                        }
```
jackddouglas, line 516: should this be before we delete superseded keys or do we want to clean up prior values regardless?

jackddouglas (approved):
Looks good though I'm confused about the purpose of `claim` now and think we could improve on how cardinality can be enforced strictly rather than implicitly through the called instruction:

My first observations are that `claim` is vestigial now and silently dropped in various places which could lead consumers to think causal chains are built when there are none. Misleading docstrings don't help.

Also makes me think `cardinality` could be strongly enforced by typing because there a number of ways `cardinality-one` in particular could degrade if implementers aren't careful to override or respect the specific semantics.

The schema should be the source of truth in both the type system and storage. I think it'd be cleaner if `Instruction::Assert(artifact)` looked up cardinality and routed itself or if `Replace` became a parameter in `Assert` (though first feels purer to me).

The exception where asserting the same value is identical implies that two entities with the same value and different causal histories are identical which further implies causality doesn't matter at all. Either this is a mistake or better to drop `cause` entirely.

rust/dialog-artifacts/src/artifacts.rs
Gozala: No failure will never mint a new block that replaces tree root so tree will remain as is and caller will get an error

rust/dialog-artifacts/src/artifacts.rs
Gozala: I believe this reflects intended behavior, if writer asserts value as cardinality one they expect that after transaction that value to be the only value present. Leaving others behind would produce unexpected result.

rust/dialog-artifacts/src/artifacts/update.rs
Gozala: I think you're right this method should not have default implementation which I do not believe is used right now.

rust/dialog-artifacts/src/artifacts.rs
Gozala: I think this illustrates real issue with current design of `cause` field. Current undocumented design should identify which of the values is a winner and use that as causal reference implying it is superseding it. Which is perhaps what I should have done instead.

Gozala:

PR #334, opened 2026-05-12
fix(dialog-macros): strip raw-identifier prefix from field names

Gozala:
## Summary

Rust field names leaked into derive-macro output verbatim, breaking two conventions the formal notation relies on:

1. **Raw-identifier prefix.** A field declared `r#type` produced descriptor key `"r#type"` instead of `"type"`.
2. **Casing.** Fields like `display_name`, `firstName`, or `LastName` produced descriptor keys identical to the Rust source instead of the kebab-case form (`display-name`, `first-name`, `last-name`) used elsewhere (attribute names, concept names, …).

Both surface in `ConceptDescriptor.with` map keys, `Term::var(...)` query variable names, and the parameter map populated by the generated `From<...> for Parameters` impls.

## Fix

Normalize once at derive time in both `Concept` and `Formula` derives:

```rust
let name = field.ident.as_ref().unwrap()
    .unraw()                    // drop r# prefix
    .to_string()
    .to_case(Case::Kebab);      // snake/camel/Pascal → kebab
```

`syn::ext::IdentExt::unraw` + `convert_case::Case::Kebab` (workspace dep already in use by `dialog-macros`). Single-word lowercase names are unchanged.

## Tests

In `dialog-query/src/concept/descriptor.rs`:

- `it_strips_raw_identifier_prefix_from_concept_field_names` — `r#type` field round-trips through JSON as `"type"`.
- `it_kebab_cases_concept_field_names` — `display_name`, `firstName`, `LastName`, `email` fields serialize as `display-name`, `first-name`, `last-name`, `email` respectively.

Both verify round-trip preserves the concept's content hash (`descriptor.this()`).

### DIFF
--- a/rust/dialog-macros/src/query/concept.rs
+++ b/rust/dialog-macros/src/query/concept.rs
@@ -71,6 +71,7 @@
 //! // PersonQuery → Parameters, Premise, Proposition, ConceptQuery
 //! ```
 
+use convert_case::{Case, Casing};
 use proc_macro::TokenStream;
 use quote::quote;
 use syn::ext::IdentExt;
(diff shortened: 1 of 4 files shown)


Gozala:

PR #339, opened 2026-05-18
feat(dialog-query): add InductiveRule sibling of DeductiveRule

Gozala:
InductiveRule shares the head + body shape and compile-time analysis (planner + unbound-variable check) of DeductiveRule via a new Compile trait. The two kinds differ only at evaluation time: deductive yields tuples on query, inductive asserts head facts during a transaction. Wire-format descriptor uses `assert!:` for the inductive head, mirroring the Scheme/Clojure `!` mutation marker.

Error variants that previously carried `Box<DeductiveRule>` now carry `Box<Rule>` (new enum over both kinds) so compile-time diagnostics report uniformly across the two.


Gozala:

PR #340, opened 2026-05-18
feat(dialog-repository): implement layer abstraction enabling layered

Gozala:
Implements `.with` method similar to `d/with` in datomic, but instead of just taking facts it takes more general `QueryLayer` implementation which can be another branch or in-memory search-tree. 

This allows you to query multiple branches as if they were merged or branch with some local facts as if they were merged.

In addition Transaction now also gained .query method just like Branch allowing you to query with pending changes if they were already applied.


Gozala:

PR #341, opened 2026-05-18
feat(dialog-remote-fs): FS remote backend

jackddouglas:
## Summary

Adds the `dialog-remote-fs` crate: a capability-based FS-remote backend that
lets a service-worker shim or a native tool expose a user-picked directory
as a sync route. Layout is byte-compatible with `dialog-storage::FileSystem`
so a directory written by this crate is a valid `slide` vault and vice versa.

Two backends share a single trait (`FsHandle`) selected at compile time:

- **Native** — `tokio::fs` + `PathBuf`, `pidlock` for cross-process CAS
  serialization against a concurrent `slide` CLI writing the same vault.
- **WASM** — `web_sys::FileSystemDirectoryHandle` via the File System Access
  API; `navigator.locks` for cross-tab CAS serialization. Writes commit
  atomically via `createWritable().close()` — no temp+rename dance.


rust/dialog-network/src/lib.rs
```diff
@@ -36,6 +37,7 @@ use dialog_remote_ucan_s3::UcanSite;
 pub struct Network {
     s3: S3,
     ucan: UcanSite,
+    fs: Fs,
```
Gozala, line 40: It may be worth considering if we should add cfg gate here. Only reason I did not add this to above two is because UcanSite is the primary one and it depends on S3 anyway so I though there was no point in having to gate it since it's going to be included anyway.

I think `derive(Site)`  should respect cfg gates and generate dispatchers that will also be gated.

rust/dialog-remote-fs/src/fs/provider/archive.rs
```diff
@@ -0,0 +1,97 @@
...
+        // declared digest. dialog-storage's native FS provider does the
+        // same check; doing it here too means a corrupt capability can't
+        // poison the on-disk content-addressed store.
+        let actual = Blake3Hash::hash(&content);
+        if actual != digest {
+            return Err(ArchiveError::DigestMismatch {
+                expected: digest.as_bytes().to_base58(),
+                actual: actual.as_bytes().to_base58(),
+            });
+        }
```
Gozala: 🤔 Maybe we should be doing this on read actually, this way something else altering block will not deceive the layers, above.

rust/dialog-remote-fs/src/fs/provider/memory.rs
```diff
@@ -0,0 +1,203 @@
...
+            .await
+    }
+}
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+impl Provider<FsInvocation<Retract>> for Fs {
+    async fn execute(&self, input: FsInvocation<Retract>) -> Result<(), MemoryError> {
+        let (parent, file_name) =
+            split_target(&input.permit.handle_id, &input.permit.request.path).await?;
```
Gozala: I wonder if we should just have method on permit itself that gives you this pair

rust/dialog-remote-fs/src/fs/authorization.rs
```diff
@@ -0,0 +1,40 @@
...
+    pub fn request(&self) -> &FsRequest {
+        &self.request
+    }
+
+    /// Redeem this authorization for a permit against the given address.
+    ///
+    /// For FS there is no signing step — this is a passthrough constructor
+    /// that pairs the captured request with the registered handle id.
+    pub fn redeem(&self, address: &FsAddress) -> FsPermit {
+        FsPermit::new(address.id().to_string(), self.request.clone())
```
Gozala: Have not gone through the rest so this comment may not hold, but it seems to me that perhaps what this should be returning is actual handle to directory wrapped in permit.

rust/dialog-remote-fs/src/handle/native.rs
```diff
@@ -0,0 +1,189 @@
...
+
+impl NativeHandle {
+    /// Construct a handle for the given path. Callers should pass an
+    /// absolute path — relative paths are not normalised here.
+    pub(crate) fn new(path: PathBuf) -> Self {
+        Self { path }
+    }
+
+    /// The underlying filesystem path this handle points at.
+    #[allow(dead_code)]
```
Gozala: If it's dead maybe we don't need a method

rust/dialog-remote-fs/src/handle/native.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub(crate) fn path(&self) -> &Path {
+        &self.path
+    }
+}
+
+fn validate_segment(segment: &str) -> Result<(), FsError> {
```
Gozala: The reason FS provider uses Uris is to avoid having to guard against containment & avoid OS specific delimiters etc...  I ink it would be better to do the same here instead.

rust/dialog-remote-fs/src/handle/native.rs
Gozala, line 1: Can we please reuse https://github.com/dialog-db/dialog-db/blob/af442cac90d72c9da8be9c71799f497bddc62f0b/rust/dialog-storage/src/storage/provider/fs.rs#L125-L131 here which I think offers most of the functionality ? It does not offer read, write, etc.. but we could add those

rust/dialog-remote-fs/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+//! provider:
+//!
+//! ```text
+//! {root}/archive/{catalog}/{base58(digest)}
+//! {root}/memory/{space}/{cell}
+//! {root}/credential/key/{address}
+//! {root}/certificate/{audience}/{subject}/{issuer}.{hash}
+//! ```
+
+#![warn(missing_docs)]
```
Gozala, line 65: Please add docs instead

rust/dialog-remote-fs/src/registry.rs
```diff
@@ -0,0 +1,77 @@
...
+            .ok_or_else(|| FsError::UnregisteredHandle(id.to_string()))
+    })
+}
+
+#[cfg(all(test, not(target_arch = "wasm32")))]
+mod tests {
+    use super::*;
+    use tempfile::tempdir;
+
+    #[test]
```
Gozala: dialog_common::test please and we need that wasm_bingen thing to opt-in into testing in browsers

rust/dialog-remote-fs/tests/archive.rs
```diff
@@ -0,0 +1,222 @@
...
+
+async fn execute_put(
+    handle_id: &str,
+    capability: dialog_capability::Capability<Put>,
+) -> Result<(), ArchiveError> {
+    let request = capability.to_request();
+    let auth = FsAuthorization::new(request);
+    let permit = auth.redeem(&dialog_remote_fs::FsAddress::new(handle_id));
+    permit.invoke(capability).perform(&Fs).await
+}
```
Gozala: Typically this kind of stuff lives in helpers.rs with feature gate enabled by default in test.

rust/dialog-remote-fs/tests/archive.rs
```diff
@@ -0,0 +1,222 @@
...
+
```
Gozala, line 20: I think without wasm_bindgen opt-in this will be running in node instead of browser

Gozala (changes requested):
I am excited to see this happening, thanks for taking on this! I think there are some things to address before we this is approved:

1. It seems to me that native implementation duplicates lot of functionality provided by [fs provider](https://github.com/dialog-db/dialog-db/blob/main/rust/dialog-storage/src/storage/provider/fs.rs) and I would very much prefer if we evolve it instead of have two implementations of same capabilities.
2. Ideally we make [fs provider](https://github.com/dialog-db/dialog-db/blob/main/rust/dialog-storage/src/storage/provider/fs.rs) backend isomorphic that is have native and web variants and one you get depends on target. This would also mean that capabilities that will be implemented universally without worrying about the target.
3. If above two things are addressed than dialog-remote-fs becomes just a simple wrapper managing credential resolution.
4. Currently no real tests seem to run in the wasm target, I think we should have tests, although that would require figuring out permissions but I suspect there must be a way. In worst case scenario we could probably use OPFS handle in tests.

PR #343, opened 2026-05-22
fix(dialog-macros): carry concept doc comment into the descriptor

Gozala:
The Concept derive captured the struct doc comment for the Concept::description trait method but the From<Struct>/From<Query> conversions that build the ConceptDescriptor dropped it, leaving description as None. A concept: query reads the descriptor, so the description never reached consumers.

Add ConceptDescriptor::with_description and have both From impls chain it with the doc comment. An empty doc comment is treated as absent so an undocumented concept serializes no blank field.


Gozala:

PR #344, opened 2026-05-22
feat: type inference

Gozala:
## Overview

Two layered concerns are landing together here:

1. **Set-widening optional concept fields.** A concept field declared `Option<T>` resolves to `None` at projection time when no fact exists, instead of dropping the row entirely. Absence is row-level, not stored — we never persist a "Nothing" value.

2. **Rule-level type inference.** Every named variable a rule mentions places a constraint on its type (from the slot it appears in). Inference unifies those constraints once, at rule-compile time, and the planner uses the result to narrow premise terms before evaluation.

The second concern grew out of a hole the first one exposed: optional attribute queries were emitting `Absent` fallback rows even when a sibling premise required the same variable to be present, leaving the join to filter the spurious rows. With inference, the planner sees the rule-level constraint and the optional query stops emitting them in the first place.

## Why

Without rule-level inference, every premise reasoned only about its own term's static kind:

```rust
// Optional attribute query for ?nickname:
//   - missing fact → emit Absent-fallback row
// Sibling premise requires ?nickname to be String:
//   - downstream join filters Absent rows
```

Two failure modes:

- **Wasted work.** Absent rows emitted, then immediately filtered.
- **Surprising errors.** Constraints like `?x == ?y` couldn't tell whether to propagate `Absent`, infer `Absent`, or filter — without rule context, the answer wasn't well-defined.

Inference makes the rule-level constraint visible to every part of the pipeline that wants it. The optional query consults the inferred kind for its `is` variable; if the rule has narrowed it to non-optional, the fallback row is suppressed at the source.

## How it works

A rule moves through three phases:

```
Parse ──► Analyze ──► Plan
```

**Parse** produces a `Vec<Premise>` from the user's `DeductiveRuleDescriptor`. No type info beyond what the user wrote at the term layer (`Term<Option<String>>` carries `String | Nothing`; bare `Term::var("x")` carries nothing).

**Analyze** (`rule::analyzer::analyze`) does three things:

1. **Type inference** — `TypeEnv::infer` walks every positive premise's slots, allocates a unifier variable per variable name, and unifies each slot's kind. Negations don't contribute; they filter on existing bindings.

2. **Required-head check** — every conclusion variable's inferred kind must not admit `Nothing`. Otherwise the rule could produce `Absent` in a required slot.

3. **Coalesce contract validation** — each `coalesce(source, fallback) = is` checks `source: Optional<α>`, `fallback: α`, `is: α`.

The output is an `AnalyzedRule` carrying the inferred `TypeEnv` (shared via `Arc`) and a per-premise dependency graph (precomputed `binds`/`needs`/`requires[]`).

**Plan** runs the existing greedy cost-based planner over the analyzed rule, then **narrows each premise's variable terms via `apply_types`**:

```rust
// Before planning: user-supplied
is: Term<Option<String>>      // kind = String | Nothing

// Rule-level inference narrows ?name to String (some sibling premise
// has a required slot for it). After narrowing:
is: Term<Any>                 // kind = String  (no Nothing bit)
```

Narrowing happens **once at plan time**. The evaluator is unchanged — its `is.is_optional()` check just sees the narrowed term and returns the right answer. No `TypeEnv` is threaded through evaluation. User-supplied premises stay untouched; only the in-flight working copy in the plan reflects the narrowing.

**Replanning** (`Conjunction::plan(&new_scope)`, used for adornment-based optimization in concept queries) reruns the same pipeline against a different scope. Inference is idempotent over already-narrowed premises.

## What the reviewer should look at

- `rust/dialog-query/src/rule/analyzer.rs` — analysis orchestration, `AnalyzedRule`, dependency graph.
- `rust/dialog-query/src/rule/types.rs` — `TypeEnv::infer`, the actual inference pass (uses the existing `type_system::unifier::Context`).
- `rust/dialog-query/src/planner.rs` and `rust/dialog-query/src/planner/plan.rs` — where narrowing is applied to premises.
- `rust/dialog-query/src/rule.rs` — the `Compile` trait whose default impl threads planning + analysis together; shared by `DeductiveRule` and `InductiveRule`.
- `notes/rule-pipeline.md` — design note covering the three phases, error attribution, and rationale for narrowing at plan time.

## Testing

The behavior change is covered end-to-end:

- `it_suppresses_absent_fallback_under_rule_inference` — confirms the planner rewrites the optional `is` term to be non-optional when rule-level inference strips `Nothing`.
- `it_strips_nothing_when_a_required_binding_also_exists` / `it_strips_nothing_when_any_premise_is_required` — multi-premise inference cases.
- `it_preserves_nothing_when_only_optional_bindings_exist` — a single optional premise keeps its local optionality through planning.
- `it_rejects_planning_when_variable_kinds_disagree` — inference contradictions raise `TypeError::TypeInference` instead of being silently swallowed.
- `it_ignores_negation_contributions_during_inference` — negations are filters, not constraints.
- `it_preserves_narrowing_across_replans` — replanning is idempotent.

Plus tests covering `Equality` over `Absent` (both-absent yields, absent-vs-present filters), the type-inference conflict path, and the dependency graph.

1485 workspace tests pass, clippy and fmt clean.

## What's intentionally not in scope

- **Consuming the dependency graph in the planner.** Today the planner still uses `Candidate::update`'s schema-walking loop; the dependency graph is built but unread. Replacing the walk with the graph is a meaningful planner refactor that I judged out of scope for this PR. The graph infrastructure is in place for that follow-up.
- **Narrowing nested concept queries.** `apply_types` rewrites `AttributeQuery::is` and walks into negated attributes. Other premise kinds (concept queries with nested terms) aren't narrowed yet because their kinds are schema-fixed today; this becomes relevant only if future premise types gain inferrable slots.

### DIFF
--- a/notes/optional-fields.md
+++ b/notes/optional-fields.md
@@ -1,0 +1,543 @@
+# Optional Fields — v2 Design
+
+This document specifies the second design pass for set-widening
+optional concept fields in dialog-db. It supersedes a prior
+implementation (commit `d6629bf7` on `feat/optional-fields`) which
+shipped a working but structurally awkward solution. The lessons
+from v1 are folded into v2.
+
+This document is the design contract for `feat/optional-fields-v2`.
+Implementation follows in subsequent commits.
+
+## Motivation
+
+Concepts model entities as records with named, typed fields. Today
+every field is required: a query for a concept with field `name`
+finds only entities that have a `name` fact. Some real-world
+attributes are inherently optional — a `Person` may or may not have
+a `nickname`, an order may or may not have a `discount_code`. Users
+want to write:
+
+```rust
+#[derive(Concept)]
+struct Person {
+    this: Entity,
+    name: Name,
+    nickname: Option<Nickname>,  // optional
+}
+```
+
+and have `Query::<Person>::default()` return entities that have a
+`name` fact, with `nickname` populated as `Some(value)` when the
+fact exists and `None` when it doesn't. The fact for a `None` field
+is **never persisted** — absence is realized at query time, not
+stored.
+
+This is the same pattern Datomic calls `get-else` and Hickey
+described as "set widening": `Optional<T>` is the set
+`T ∪ {Absent}`, with the subtype rule `T ⊆ Optional<T>`. A `T` value
+flows freely into an `Optional<T>` slot.
+
+## v1 retrospective
+
+v1 (commit `d6629bf7`) shipped working set-widening across nine
+slices: `Binding<Value>` at the Match layer, a schema-level `Type`
+enum with `Required`/`Optional`/`Any`, `AttributeQueryOptional`
+streaming wrapper, `Constraint::Coalesce`, marker-trait taxonomy
+with structural rejection of `Term<Option<Option<U>>>`, and end-to-end
+tests proving both `Some(...)` and `None` realize correctly.
+
+It works. The native test suite is 1361 green. The user-visible
+behavior is correct.
+
+But the implementation accumulated structural debt that motivates
+v2:
+
+1. **Two parallel type taxonomies.** A schema-layer `Type` enum
+   (`Any | Required(ValueType) | Optional(ValueType)`) and a
+   descriptor-layer `Option<ValueType>`. The same concept expressed
+   twice in different shapes, with conversion sites between them.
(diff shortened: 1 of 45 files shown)


Gozala:
Superseded by https://github.com/dialog-db/dialog-db/pull/347

PR #346, opened 2026-05-31
feat(dialog-query): rule-level type inference and optional concept fields

Gozala:
## Overview

Rebased successor to #344 (based on `fix/tr-metadata` instead of `fix/concept-description`). Lands two layered concerns plus the review fixes from #344.

1. **Set-widening optional concept fields.** A concept field declared `Option<T>` (or a `maybe` attribute) resolves to `Absent` at the row layer when no fact exists, instead of dropping the row. Absence is row-level, never persisted — no `Nothing` value is stored.

2. **Rule-level type inference.** Every named variable a rule mentions places a type constraint (from the slot it appears in). A per-rule unifier folds those via a meet lattice at rule-compile time, and the planner narrows premise terms before evaluation. The optional query consults the inferred kind for its `is` variable; if the rule narrowed it to non-optional, the Absent fallback is suppressed at the source.

## How it works

A rule moves through three phases: **Parse → Analyze → Plan**.

- **Analyze** (`rule::analyzer::analyze`): type inference (`TypeEnv::infer` unifies each slot's kind per variable; negations don't contribute), a required-head check (no conclusion variable may admit `Nothing`), and Coalesce contract validation.
- The unified type system is `Type = Primitive | Composite` with a meet/intersect lattice; the unifier resolves per-rule type variables. The inferred `TypeEnv` is projected onto each plan step to narrow terms before evaluation.

## Review fixes (carried in)

These address findings raised in review of #344:

- **Coalesce** reconciles a pre-bound `is` as a filter instead of binding unconditionally — previously a pre-bound `is` (to a different value or to Absent) made `?` abort the whole result stream.
- **Optional attribute queries** no longer emit a phantom `Absent` fallback on a value mismatch (a fact exists with a different value) or for an unbound entity; the fallback now requires a determined entity and a genuine "no fact" miss (in both `only.rs` and `all.rs`).
- **`ConceptDescriptor` is valid-by-construction.** The collection constructors are fallible `TryFrom` (`Error = TypeError::EmptyConcept`) and `Deserialize` rejects an empty attribute map, so a concept with no required (`with`) attribute — which would match every entity — is unconstructable. Concept types expose their descriptor via `Descriptor<ConceptDescriptor>` (cached, mirroring attributes) and an inherent `descriptor()`; the redundant derive-generated `From<Struct>/From<Query> for ConceptDescriptor` is removed. `#[derive(Concept)]` emits a compile-time assertion (over the `ConceptField::OPTIONAL` const, not syntactic `Option` detection) that a concept declares at least one required field — all-optional and no-attribute structs fail to compile, covered by `compile_fail` doctests.

## Compatibility

- `#[derive(Concept)]` now rejects all-optional / no-attribute concept structs at compile time.
- `ConceptDescriptor::from(<collection>)` is now `try_from` (returns `Result`).

### DIFF
--- a/notes/optional-fields.md
+++ b/notes/optional-fields.md
@@ -1,0 +1,395 @@
+# Optional Fields & Type System — v2 Design
+
+This document was written as a *design contract* before
+implementation. The implementation took a smaller path than the
+contract proposed. The doc has been split:
+
+- **What shipped** — the set-widening type system, the unifier-backed
+  rule-level type inference, the `Resolution` policy, and the
+  `Coalesce` constraint. These are now in code under
+  `rust/dialog-query/`. The shipped pipeline is documented in
+  [`rule-pipeline.md`](./rule-pipeline.md); the type-system shape is
+  in `src/type_system.rs` and `src/type_system/unifier.rs`.
+
+- **What didn't ship** — rank-1 polymorphic *formulas* with
+  `TypeScheme`/`SchemeBody`/`SchemeType`, and the `instantiate` /
+  `generalize` operations that go with them. The schemes were the
+  design's answer to "generic formulas like `math/sum`"; we didn't
+  have a concrete consumer that needed them, so the work was
+  deferred to a follow-up.
+
+The remainder of this document is the original design contract,
+preserved so the historical intent is visible. **Names like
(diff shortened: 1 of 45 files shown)


Gozala:
Superseded by #347

PR #348, opened 2026-06-02
test(dialog-query): optional field sorted before required drops rows

Gozala:
## The bug

A `this`-unbound concept query whose **alphabetically-first field is optional** drops every entity that lacks that optional fact, instead of set-widening it to `Absent`.

A concept's fields live in a `BTreeMap`, and the concept-query planner leads the `this`-unbound scan with the first field. When that lead premise is optional, its `Absent` fallback is suppressed while `this` is still unbound (the `entity_known = selection.of().is_constant()` guard in `attribute/query/all.rs`), so the lead scan yields only entities that *have* the optional fact. Entities missing it never produce a row.

`it_executes_concept_with_optional_field` passes only because its optional field (`nickname`) happens to sort *after* the required `name`. Rename the optional field so it sorts first and the same query silently drops rows.

## This PR

Adds `it_set_widens_optional_field_sorted_before_required` — the exact sibling of the existing optional test, with the optional field renamed `bio` (sorts before `name`). Two entities are asserted (Alice with only `name`, Bob with `name` + `bio`); the query should return both (Alice's `bio` Absent, Bob's Present) but returns only Bob.

The test is `#[ignore]`d so it documents the bug without breaking CI. Remove the attribute once the planner leads an unbound scan with a required field (never an optional one).

## Reproduce

```
cargo test -p dialog-query it_set_widens_optional_field_sorted_before_required -- --ignored
```

Found while integrating optional concept attributes downstream in tonk: a plain `person:` query (required `name` + optional `age`) returned only the people who had an `age`, because `age` < `name`.

### DIFF
--- a/rust/dialog-query/src/concept/query.rs
+++ b/rust/dialog-query/src/concept/query.rs
@@ -549,6 +549,137 @@ mod tests {
         Ok(())
     }
 
+    /// Regression: the set-widening of an optional field must not
+    /// depend on the field's alphabetical position relative to the
+    /// required fields.
+    ///
+    /// This is the sibling of [`it_executes_concept_with_optional_field`]
+    /// with one change: the optional field is named `bio`, which
+    /// sorts BEFORE the required `name`. Because the concept's
+    /// fields live in a `BTreeMap` and the planner leads the
+    /// `this`-unbound scan with the first field, an optional lead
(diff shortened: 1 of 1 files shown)


Gozala:
Incorporated into https://github.com/dialog-db/dialog-db/pull/347

PR #355, opened 2026-06-12
perf: bulk-build empty-tree batches and validate node buffers once

Gozala:
Stacked on #354. Targets the seed-commit numbers measured in the tonk app (51-expression scaffold: ~1416 ms at `1bf3f48c`), which are CPU-bound in the per-insert mutation path: each insert rewrites the whole touched segment (~176 entries deserialized, re-ranked, re-serialized, re-hashed per level), three inserts per instruction, and `Node::body` ran a full bytecheck validation on every access.

- `Tree::from_entries`: one-pass bottom-up bulk construction. Every key ranked once, every node serialized and hashed exactly once, zero reads. A property test pins that the result is byte-identical to the canonical tree an insert loop produces (reverse insertion order, duplicate keys included).
- `ArtifactTreeExt::apply` empty-tree fast path: batches applied to an empty tree (the seeding path) stage the final key/value set in memory, including `Replace` cardinality-one supersession against the staged state, then bulk-build. An equivalence test pins chunked applies == single bulk apply.
- Validate-once node access: buffers carry a shared validated marker; after the first successful `rkyv::access` validation, body access is a pointer cast. Note for review: this introduces the crate's first `unsafe` block (`access_unchecked`), with the safety argument documented inline (immutable aligned buffers, marker only set by the validating path, content addressing pins the bytes).
- `benches/bulk.rs`: criterion comparison of `insert_loop` vs `from_entries` at 10/100/1k/10k entries, sequential and random keys, for the graphs validating the change.

### DIFF
--- a/rust/dialog-artifacts/src/tree.rs
+++ b/rust/dialog-artifacts/src/tree.rs
@@ -23,9 +23,11 @@
 //! orphan rule rules out inherent methods — the operations are exposed as
 //! an extension trait instead.
 
+use std::collections::BTreeMap;
+
 use async_stream::try_stream;
 use async_trait::async_trait;
-use dialog_common::{Blake3Hash as NodeHash, ConditionalSend, ConditionalSync};
+use dialog_common::{Blake3Hash as NodeHash, ConditionalSend, ConditionalSync, NULL_BLAKE3_HASH};
 use dialog_search_tree::{ContentAddressedStorage, Entry, Tree, Value as TreeValue};
 use dialog_storage::{Blake3Hash, DialogStorageError, StorageBackend};
 use futures_util::{Stream, StreamExt};
@@ -144,6 +146,16 @@ impl ArtifactTreeExt for ArtifactTree {
             + ConditionalSync,
         I: Stream<Item = Instruction> + ConditionalSend,
     {
+        // A batch applied to an empty tree (the seeding path) stages
+        // entirely in memory and builds the tree bottom-up in one pass:
+        // every key is ranked once and every node is serialized and hashed
+        // exactly once. The per-instruction loop below instead rewrites
+        // each touched segment once per insert, which dominates seed
+        // commits.
+        if self.root() == NULL_BLAKE3_HASH {
+            return apply_to_empty(self, instructions).await;
+        }
+
         let storage = ContentAddressedStorage::new(TreeStorageBridge(store.clone()));
 
         tokio::pin!(instructions);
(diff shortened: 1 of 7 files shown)


Gozala:
Benchmark results from `benches/bulk.rs` (M-series native, criterion; HTML report with plots under `target/criterion` after `cargo bench -p dialog-search-tree --bench bulk`):

| entries | insert loop | from_entries | speedup |
|---|---|---|---|
| 10 | 21.6 µs | 3.9 µs | 5.5x |
| 100 | 822–884 µs | 16–18 µs | ~50x |
| 1,000 | 24.2–25.6 ms | 168–171 µs | ~145x |
| 10,000 | 357–426 ms | 1.7–1.8 ms | ~220x |

Sequential and random key orders within a few percent of each other. The insert loop is superlinear (every insert rewrites its whole touched segment per level); the bulk build is essentially linear. Both produce byte-identical roots (pinned by property tests), so the gap is purely redundant work.

The validate-once unsafe was reworked per review discussion: the proof now lives on `Node<Key, Value>` (shared across clones) instead of the type-erased `Buffer`, binding it to the exact archived type by construction.

Gozala:
Status after the queue cleanup (2026-07-05): this can't be rebased onto `main` — it predates the #354/#357 restructuring. Where each piece stands:

- **Empty-tree bulk apply / per-insert rewrite cost**: superseded. `ArtifactTreeExt::apply` on `main` now opens one `TransientTree` edit batch over the spine and persists once per instruction stream (#357), which removes the full-segment-rewrite-per-insert cost this PR targeted. The `shaper.rs` and old `Tree` code the first two commits patch no longer exist.
- **Validate-once node access**: still open on `main` — `PersistentNode::body()` runs full `rkyv::access` bytecheck validation on every call (`node/persistent.rs`). Worth doing, but it needs a fresh design because the node cache stores raw `Buffer`s, so the shared validated marker has to live on the buffer, not the node. That's a new PR against the current architecture, ideally with the criterion evidence re-measured.
- **`Tree::from_entries` bottom-up bulk build**: possibly still a win for the empty-tree seed path on top of the transient batch, but it would also be a fresh implementation.

Leaving open as the tracking item for the validate-once + bulk-seed follow-up rather than landing a stale tree.

---

Gozala:
The port landed as #369: `TransientTree::seed` (bulk bottom-up build for empty trees), the empty-tree fast path in `ArtifactTreeExt::apply` with staged `Replace` supersession, validate-once `PersistentNode::body()`, and the bulk benchmark. This PR can close once #369 merges.

---

PR #359, opened 2026-06-16
fix(storage): Windows filesystem path portability

s0lness:
The native filesystem storage provider builds paths from DIDs (`did:key:z6Mk...`), which are used directly as directory names in the certificate and credential layouts. Two of these are invalid on Windows, so any operation that writes such a path fails.

### Repro

On Windows, building an operator / saving a delegation hits:

1. `os error 123` (ERROR_INVALID_NAME) creating `.../certificate/{audience}/{subject}/...`, because `:` (and `< > " \ | ? *`) are illegal in Windows path components and the segments are DIDs.
2. `os error 267` (ERROR_DIRECTORY) on the `FileSystemHandle -> PathBuf` round trip, because only a trailing `/` was trimmed, leaving a trailing `\` on a Windows-formatted path.

Surfaced while running the `slide` CLI on Windows ARM64; reproduces on any Windows host.

### Fix

Both changes are in `rust/dialog-storage/src/storage/provider/fs.rs`:

- A `sanitize_path_segment` helper maps reserved characters to `_`, per `/`-separated component, applied at the two segment chokepoints (`FileSystemHandle::resolve` and the `Location -> handle` builder). The mapping is deterministic, so a write and a later read resolve to the same on-disk name.
- The `PathBuf` conversion now trims a trailing `\` as well as `/`.

### Compatibility

- The sanitizer is `#[cfg(windows)]`, so the on-disk layout on Linux/macOS is unchanged.
- The separator trim additionally strips `\`, which is a no-op where paths don't use it.
- Adds two Windows-only tests. `cargo test -p dialog-storage` passes (135) on this branch.

### One open question

The sanitizer maps reserved characters to `_`, which is lossy: two segments differing only in reserved characters would collide. In practice the segments are DIDs that differ in their base58 body, so this doesn't arise, but if you'd prefer a reversible encoding or a different sink character I'm glad to change it. Note that percent-encoding at the URL layer doesn't survive, since `to_file_path()` decodes it back.


Gozala:
Hey @s0lness thanks for taking time to look into this and writing up a PR. I do wonder though if there is a different / better way to address this specifically handling lossiness, I'll provide more details inline

rust/dialog-storage/src/storage/provider/fs.rs
```diff
@@ -92,7 +125,7 @@ impl TryFrom<&Location> for FileSystemHandle {
         let path = if location.name.is_empty() {
             base
         } else {
-            base.join(&location.name)
+            base.join(sanitize_path_segment(&location.name).as_ref())
```
Gozala: Rational for `FileSystemHandle` being a wrapper of the `Url` was to avoid having to worry OS specifics and basically use `file:` URIs which can be converted to os specific paths via `From<FileSystemHandle> for PathBuf` here

https://github.com/dialog-db/dialog-db/blob/7a85a4d8dcdfcfae1755a1d78979e0e4de0d5414/rust/dialog-storage/src/storage/provider/fs.rs#L207-L231

However looking at fixes below it does not seems to work correctly either. That said turns out there is  [to_file_path](https://docs.rs/url/latest/url/struct.Url.html#method.to_file_path) exists on Urls and perhaps we could just use that instead which I expect takes care escaping in standardized way and probably handles various other edge cases.

Gozala (changes requested):
Thanks for taking time to investigate and fix the issue, it is a great feeling to get new people contributing ❤️

I think there might be a better way to address this without lossiness issue you've raised. Would you be willing to try that and maybe also report whether that works on windows as expected. I provided more details inline in the diff

s0lness:
Thanks so much for the fast and generous feedback, it genuinely helped. Quick context: i'm not really a Rust or a very technical person in general, just wanted to get things running on a Windows machine, so please don't hesitate to correct anything that looks off.

Your point about `FileSystemHandle` wrapping a `Url` to stay OS-agnostic was the thing that reframed it for me.  I've moved everything to the boundary instead.

One note on `to_file_path`: the code already routes through it in `From<FileSystemHandle> for PathBuf`, and that's the path that was failing. It percent-decodes and handles the structural conversion (drive letters, UNC), but it doesn't escape characters that are illegal in Windows filenames, so a `:` in a DID passes straight through into an invalid path (os error 123). So I kept `to_file_path` and layered the escaping on top of it rather than swapping it out.

What the new version does:

- Leaves the URL/logical layer untouched (DIDs keep their colons).
- At the `FileSystemHandle -> PathBuf` boundary, percent-encodes the Windows-reserved characters per `Normal` path component, and decodes them back in `list()` so callers always work in logical names. The drive-letter colon is a `Prefix` component, so it is left alone.
- It is reversible (`:` -> `%3A`), which addresses the lossiness concern directly: distinct names can't collide, and reads recover the original.
- Rebuilding the path from `components()` also drops the trailing separator that was causing a second Windows failure (os error 267), so both fixes collapse into one change.
- All `#[cfg(windows)]`, so other platforms keep their exact on-disk layout.

Scoped to reserved characters. I left Windows reserved device names (`CON`, `NUL`) and `MAX_PATH` out, since DIDs don't hit those, but happy to add them if you'd prefer.

Verified on Windows: `cargo test -p dialog-storage` is green, including a reversibility test and a round-trip that writes under a `did:key:...` path, lists it, and recovers the original logical name. The downstream UCAN push / pull / delegation flow works too. On disk the DID directories now look like `did%3Akey%3A...`, with no raw colons anywhere.

Let me know if the encoding approach or anything else looks off, glad to adjust.

Gozala:
Heads up @s0lness — this was merged during a queue cleanup and then reverted in #368: the per-component escaping at the path-conversion boundary isn't the shape we want this solved in; the preference is for the file-URI → path layer to own the concern rather than Windows-specific character mapping sprinkled at the `PathBuf` boundary. The Windows breakage itself is real and worth fixing, so let's iterate on the approach here. One constraint for the next round (from your own investigation, worth keeping front and center): `Url::to_file_path()` percent-decodes, so escaping that lives only in the URL doesn't survive conversion.

---

s0lness:
Thanks for the heads-up, and for #370. Since the `cfg(windows)` code could only be type-checked under a forced-cfg build, I ran it on a real Windows machine: on Windows 11 (ARM64) at f777fe7c, `cargo test -p dialog-storage` passes (137 passed, 0 failed), including the escaping round-trip tests. The downstream UCAN flow works too: with all dialog crates patched to f777fe7c, tonk's `ucan_integration` tests (`it_pushes_and_pulls_via_ucan`, `it_collaborates_via_ucan_delegation`) both pass. Happy to re-test on Windows whenever useful.

PR #362, opened 2026-06-22
fix(repository): make pull fail-safe against a racing commit

Gozala:
## Why

`Branch::pull` reads the local head, merges over several `await`s (fetch, block hydration, tree integrate), then publishes the merged revision. The final publish is a compare-and-swap on the revision cell — but `Publish::perform` read its *expected* version **live from the shared cell cache at publish time**, not from the snapshot the merge was computed against.

So when a commit advanced the head *during* a pull (the everyday case in the service worker, where the auto-sync pull races a local commit on the same branch), the pull published a tree built from its now-stale snapshot while presenting the *post-commit* version as the CAS precondition. The swap **succeeded**, and the committed change was **silently dropped** — no error.

## What

- **`Cell::checkpoint`** — captures the cell's current version and returns a `Checkpoint`. `Checkpoint::publish` CAS's against the checkpointed version rather than the live cache, so a write computed from a value read earlier fails with `VersionMismatch` instead of clobbering a concurrent write. Unlike a detached copy it shares the cell's cache, so a successful publish updates the originating cell — nothing to copy back.
- **`pull`** checkpoints the head up front (CAS'd at publish → a racing commit makes it fail rather than overwrite) and checkpoints the upstream sync-base marker just before its write. The head publish and the marker write are not atomic; if a concurrent pull landed its own consistent (head, sync-base) pair in between, the marker write's mismatch is **swallowed** and pull yields the head that pull left in place, rather than clobbering a consistent pair back to a stale base.
- **`Branch::refresh`** — re-resolves the head and upstream cells from storage, the consumer's recovery path: on a mismatch, refresh then re-pull. The re-pull merges from the now-current snapshot and reuses the blocks the failed attempt already imported locally, so it does not re-hit the network for what was already fetched.

## Test

The reframed pull test (`it_fails_a_pull_racing_a_commit_then_reconciles_on_refresh`) drives a commit through one handle while another pulls, and asserts the pull fails loudly with a `VersionMismatch` (no silent loss), then that `refresh` + re-pull reconciles both the upstream change and the raced commit. All 109 `dialog-repository` tests pass; clippy + fmt clean.


Gozala:
This work is already fully on `main`: `pull.rs`, `memory/cell.rs` (`Cell::checkpoint`), and `commit.rs` on this branch are byte-identical to `main`, `Branch::refresh` is present, and the `it_fails_a_pull_racing_a_commit_then_reconciles_on_refresh` test is there too — the content landed as part of the #357/#358/#364 sequence. The only remaining diff is `main` being newer (the #364 rule/plan caches). Closing as already landed.

---

PR #372, opened 2026-07-07
feat: content-addressed blob storage, replication, and sync

Gozala:
## What

Blob storage, replication, and sync: a content-addressed blob layer spanning the
capability surface (`dialog-effects`), storage providers (FileSystem/OPFS,
in-memory, S3, UCAN-S3), the artifacts blob index, and repository-level
`write_blob`/read/list with lazy remote hydration and push-time replication.

## Stacking

This PR is **stacked on #371** (`feat(dialog-remote-fs): FS remote backend`) and
targets `feat/dialog-remote-fs-main` as its base. It will merge cleanly once
#371 lands; retarget to `main` after that.

## Grafting notes

The original blob-sync branch sat on top of the orphaned `feat/dialog-remote-fs`
stack (see #371). These 21 commits were replayed onto the main-based fs-remote
branch. A few adjustments were needed to build against current `main`:

- `Space::load` gained the `blob` field the branch added to `Space::open`
  (main's #365 introduced the separate `load` path after the branch was written).
- `dialog-repository` blob paths switched from absolute `dialog_effects::blob::*`
  to `use` imports (main's `clippy::absolute_paths` lint).
- Two `dialog-storage` blob clippy nits (`ok_or`, needless borrow) and formatting.
- Refreshed two stale `dialog-remote-s3` doctests to the current
  `S3Request::from(&cap).attest(cred).redeem(&address)` API (these were already
  broken on `main`).

## Verification

- Full workspace `cargo clippy --all-targets --all-features -- -D warnings` clean.
- `cargo fmt --all --check` clean.
- Native tests pass across storage, remote-fs, remote-s3, remote-ucan-s3,
  repository, artifacts, and operator; wasm target compiles.


Gozala (commented):
Reviewed the full stack. Commits 1–10 are the replayed blob-sync foundation (design record, blob index + `blob_changes`, the `dialog-effects` Blob surface, FS/S3/UCAN providers, true-streaming OPFS); commits 11–21 are the integration layer on top. Overall this is solid and I'm comfortable with it landing once web CI is green. Notes below — the two "Decision" items are choices worth confirming rather than defects.

### Strengths in the integration layer
- **`write_blob` takes a fallible streaming source** (`Stream<Item = Result<Vec<u8>, BlobError>>`) and hashes-as-it-writes — the right call: it can't silently ingest a truncated file under a valid hash, and it doesn't force callers to buffer.
- **UCAN authorizer generalized to variadic command segments** (01c473a) closes a real gap — the S3/UCAN provider side existed, but `/archive/blob/{read,import}` weren't dispatchable, so blob sync would've failed authorization. Good that the tonk e2e caught it.
- **Removing `dialog-blobs`** matches the direction (blobs via the unified FileSystem provider, not the old crate).
- `read_blob`'s local → remote-hydrate → digest-verify → serve, and push-time propagation over the BLOB-tag differential, are sound and match the design record.

### Decision 1 — Space blob backend: dedicated field vs composite archive
`0f635a6` splits blobs into a 5th `Space<A, M, C, D, B>` field. It works and is green. The alternative is a composite `Archive<Blocks, Blobs>` under the existing archive slot (`WebSpace = Space<Archive<IndexedDb, FileSystem>, …>`), which reaches the same runtime placement (blocks on IndexedDB, blobs on OPFS) without adding a type parameter that ripples through all five space aliases + the two manual `Prove`/`AccessRetain` impls + `Resource::{open,load}`. Both are fine; flagging so it's a conscious choice. No change requested.

### Decision 2 — API is hash-addressed; design record is entity-addressed
The repository surface is `write_blob → Blake3Hash`, `read_blob(&hash, range)`, `blob_size(&hash)`, `list_blobs()`. The design record (notes §Identity / §Query) references a blob by a content-derived entity `blob:<hash>`, and `Entity::from_blob` / `blob_hash` already exist in this PR — but nothing in the repository API returns or takes an `Entity`. I'll put up a small follow-up layering an entity-addressed facade (`blobs().add(source) → Entity`, `get(entity).{read, slice, size}`) over these commands so both coexist; the hash commands stay as the lower layer.

### Remaining / deferred
- **Query-engine formulas** `blob/size` / `blob/slice` (notes §Query) are implemented here as imperative `Branch` methods, not datalog bindings — that phase is still open (separate follow-up).
- **In-browser OPFS coverage**: the WebSpace write→read round-trip test is great; the ranged-read and import-verify streaming variants aren't yet exercised in-browser (easy to add — the `fs/blob.rs` tests just need to build for wasm32).
- **Multipart `Import`**, **BAO verified range reads**, **History index (tag 3)**, and **GC of unreferenced blobs** remain intentionally deferred.

---

PR #385, opened 2026-07-09
fix(dialog-repository): skip push when nothing new to push

Gozala:
## Why

`Branch::push().perform()` re-published the revision pointer to the upstream on **every** call, even when the local head already equaled the recorded upstream sync point — i.e. when there was nothing new to push.

The tonk service worker drains sync every ~10s while a tab holds a live subscription, and each drain calls `push()`. With no no-op guard, an idle branch (no commits since the last push) still:

- re-fetched the upstream branch,
- computed the tree diff against the recorded base (empty novelty),
- **PUT `branch/<b>/revision` to the remote anyway** (`upstream.publish(revision)`),
- re-published the local upstream sync point.

The visible symptom is a stream of `.../branch/main/revision` PUTs to R2/S3 with no transactions behind them — ongoing pushes on a quiet, idle space.

## What

Short-circuit at the top of `perform`, after resolving the local `revision` and the recorded `base` (`upstream_state.tree()`):

```rust
if revision.tree == base {
    return Ok(Some(revision));
}
```

When the local head already equals the recorded sync point there is nothing to upload and nothing to re-point, so an idle branch now does zero push I/O. The first push after a commit is unaffected (`revision.tree != base`), and the non-fast-forward divergence check still runs for real pushes.

Covered by a new test (`it_is_a_noop_when_nothing_new_to_push`): a second push with no intervening commit returns the current revision without re-publishing. The existing 5 push tests still pass.

Stacked on `feat/inductive-self-negation`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1406,6 +1406,7 @@ dependencies = [
  "tower-http",
  "url",
  "wasm-bindgen",
+ "wasm-bindgen-futures",
  "wasm-bindgen-test",
  "web-time",
 ]
--- a/notes/claim-based-serialization.md
+++ b/notes/claim-based-serialization.md
@@ -1,0 +1,104 @@
+# Claim-Based Serialization: Remove Serialize from Capabilities
+
+## Status
+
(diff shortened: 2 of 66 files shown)


Gozala:
Closing: this change (skip push when nothing new to push) already landed on main via #386. During the stack rebase onto the scrubbed main history, the commit was detected as already-upstream (identical patch-id) and dropped, leaving this PR empty.

PR #388, opened 2026-07-11
Feat/automerge records

goblinoats:


Gozala:
We have reconsidered approach for automerge integration

PR #389, opened 2026-07-12
Version control: revision DAG, signed heads, observed-remove merge

Gozala:
Grounds repository version control in a revision DAG and retires tombstones as the representation of deletion. (This PR originally carried the Revision/TreeReference move to dialog-capability; the branch now carries the full version-control stack, which redesigns Revision around the DAG instead.)

## Version-control primitives

- Revisions carry a (Origin, Edition) version: the edition is a Lamport timestamp derived from the DAG (max of parents + 1), the origin is the lineage-scoped identity of the issuer (profile, subject, branch, session key). Two revisions with the same edition but different origins are concurrent by inspection.
- Revision records live in the tree's history region as atomic signed facts carrying parents, skip links, and attribution; the `dialog.` attribute namespace is reserved for them. Skip links give logarithmic ancestor search.
- Heads and records are Ed25519-signed and verified at every trust boundary (a pull refuses a forged or tampered head before walking any of its blocks).
- Ancestry, causal ordering, and conflict detection are projected into queries via formulas and derived concepts, with memoized causal verdicts.
- Branches support multiple tracked upstreams; pull can target any of them and starts tracking a new one on first use.

## Observed-remove merge (no tombstones)

The log (history region) is the truth; the active index is a cache of the live fold. A retraction deletes the fact's index keys and appends a signed retract record; a replica's causal context (per-origin watermark of its head's ancestry, exact by the sequential-origin invariant) is what stops a stale peer from resurrecting the fact. Pull integrates the upstream delta onto the local tree in one screened pass, history region before data regions:

- R1: an incoming live claim the receiver has observed is never re-applied (re-applying would resurrect a deletion).
- R2: incoming removes stay byte-guarded, which is already the correct observed-remove rule.
- R3: incoming covering records (retraction cause, replace supersedes) retire the covered claims still live locally, by scanning the record's (entity, attribute) slot and matching versions. Coverage matches by version, not value: a replace supersedes claims of other values, which live at other keys, so probing the record's own keys would silently miss them (found and fixed during review, with a stale-peer replace acceptance test).

Convergence is pinned by acceptance tests: deterministic resurrection guard (20/20, previously a hash-race coin flip), quiescence under mutual pulls, the Alice/Bob/Mallory/Jordan OR-set scenario, resurrection surviving an empty-base pull, and the replaced-value stale-peer case.

Spec: `notes/version-control.md`, a ground-up explainer that assumes no CRDT background and doubles as the spec of the shipped design: every pull scenario with a worked example, the deletion scenarios, winner selection end to end on values, measured costs including the commit path, the invariants, and the open follow-ups. It moves in the same commits as algorithm changes.

Verification: 1814 native tests, gated integration tests, clippy clean with `-D warnings`, wasm target compiles.

## Performance

Pull cost is pinned by a read-amplification harness (`dialog-repository`, `read_amplification`, an ignored test run in release; costs in archive `Get` block reads and wall time). Two mechanisms keep pull cost proportional to what actually needs merging rather than to history depth or upstream churn:

- **Context memo**: the receiver's causal context (per-origin watermark) is memoized by head version and maintained incrementally (commits extend it by one version; pulls fold the revision records riding the delta). An invariant test pins memo-equals-walk across adopt, commit, and merge.
- **Graft merge**: tracked pulls merge by partition and graft. Node-level diffs against the sync base yield each side's divergence spans without enumerating entries; the merged tree is stitched from whole subtrees of unilaterally-changed spans (adopted by hash, unread, even for content held only by reference; a new canonical `TransientTree::stitch` primitive with oracle property tests pins stitched roots byte-equal to from-scratch builds); contested spans integrate the smaller side's scoped delta through the usual screens; coverage repair enumerates both sides' covering records from a compact coverage region (a mirror holding no value bytes) and retires covered claims by version. Cost: the intersection plus coverage plus seams. Pinned: a three-commit tracked pull after adopting ten thousand commits of foreign bulk merges in 22 block reads.
- **Reverse replay**: when both sides moved and the upstream publishes its watermark, the merge adopts the upstream tree as the substrate and replays the replica's own delta onto it, screened by the upstream's watermark and tree (the same R1/R2/R3 rules with the roles swapped). Pull cost becomes proportional to the replica's divergence, independent of upstream churn: one local commit replayed against two hundred commits of foreign-namespace churn stays under thirty block reads (pinned by test). The merged head's watermark is the union of the two published ones, derived with zero reads.
- **Published watermarks**: every head carries its context inside the signed payload. Pull compares the two watermarks before any differential: an upstream whose context is included in ours is skipped outright (nothing it has can be news), and an upstream whose context includes ours, when we have no local novelty, is adopted by root with no diff and no import, blocks hydrating lazily on demand. When knowledge has genuinely diverged the screened merge runs unchanged, which is what keeps deletion safe (a dedicated test pins the refused-adoption counterexample).

At a ten-thousand-commit history (60,000 entries in a 249-node tree): no-op sync tick, fast-forward, and initial adoption are all 0 reads; a symmetric two-commit merge reads 32 nodes in 9ms (the write-amplification floor of the key layout: six content-scattered keys per one-fact commit, both sides, plus index and write paths; tiny merges route past the graft to the cheaper direct direction); a tracked merge after adopting the full bulk reads 22. Commits run at about 3ms at depth (revision records dropped their dead value ordering, and verified records are memoized by version, so ancestry steps skip the read, decode, and Ed25519 verify).

Measured against the pre-history-index baseline: merge reads roughly double (18 to 32, tracking the 3-to-6 per-fact key amplification) and a single-fact commit goes from ~0.6ms to ~3ms. That overhead buys signed attribution, exact watermarks, convergent deletion, and the queryable causal log, and it is per commit rather than per fact, so batching amortizes it. Remaining levers, recorded in the spec: collapsing revision records to a single clustered ordering, leaf-size tuning via the format manifest, a durable novelty buffer for write-ahead-style single-fact commits, and roundtrip structure for the passive read-by-hash backend (concurrent level fetches plus push-time delta manifests; the backend is storage only, nothing server-side computes deltas). Full tables and the observation recipe (the `Counting` helpers provider) are in `notes/version-control.md`.


Gozala:
Was superseded by #397

PR #390, opened 2026-07-12
Add retain_entities to Changes and Overlay

Gozala:
Tonk's service worker stamps per-client session facts (`site:<client-id>`) into branch overlays. When a page reloads, the browser assigns a new client id, so the old client's facts must be garbage-collected — but the overlay offered no primitive for it: `retract` appends tombstones (the overlay grows on every sweep) and `clear` drops unrelated sessions' facts along with the dead ones. In practice this made every reload permanently enlarge the overlay every read folds, degrading load times until the worker restarted.

`retain_entities` removes an entity's recorded changes outright, asserts and retracts alike. The `Overlay` wrapper bumps its epoch only when something was actually removed, so subscriptions re-evaluate exactly when the readable contents changed.

Based on `feat/inductive-self-negation` (the rev tonk pins) so the consumer can bump its pin directly on merge.

### DIFF
--- a/rust/dialog-artifacts/src/artifacts/update.rs
+++ b/rust/dialog-artifacts/src/artifacts/update.rs
@@ -87,6 +87,19 @@ impl Changes {
         ChangeStream::from(self)
     }
 
+    /// Drop every change recorded for entities that fail `keep`,
+    /// asserts and retracts alike. Returns `true` when anything was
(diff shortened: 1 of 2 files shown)


Gozala:
Absorbed into #391, which now carries this commit plus the Revision/TreeReference move on the same branch. Tonk needs both, and one branch means one tag to pin instead of a stack — closing this in favour of the combined PR.

PR #405, opened 2026-07-22
Reserve the rule namespace: db.rule/* -> dialog.rule/*, privileged write only

Gozala:
Renames the deductive-rule attributes from the unreserved `db.rule/*` to `dialog.rule/*`, moving them under the reserved `dialog.` prefix, and closes the spoofing gap that rename exposes.

## Why

`dialog.` is a reserved namespace: the write gate in `dialog-artifacts` refuses any public Assert/Replace/Retract whose attribute begins with `dialog.`, so an application cannot forge version-control lineage. Rules, however, lived under `db.rule/*` — outside the reserve — and were written as ordinary asserts, so **any app could forge a deductive rule** by hand-writing `db.rule/conclusion` / `db.rule/source` facts. This was the one system-owned namespace with the vaguer `db.` root and no protection.

Moving rules to `dialog.rule/*` makes the reservation gate cover them, and standardizes on a single reserved prefix (`dialog.`) with one taxonomy: `revision`, `branch`, `session`, `replica`, `rule`.

## What changed

Because `dialog.` is refused on the public write path, rule writes now travel the **privileged record rail** — the same surface revision records use — instead of ordinary asserts:

- **dialog-artifacts**: `BufferedArtifactTree::install` / `uninstall` write a rule's facts (all three EAV/AEV/VAE index orderings, spilling a large body to the block store) directly into the buffered tree, off the instruction stream, so the `starts_with("dialog.")` gate never sees them and a rule-only commit still mints a revision.
- **dialog-repository**: a `RuleWrite { Install, Remove }` side channel on the commit, and a sanctioned `Transaction::install_rule` / `remove_rule` API. Apps install rules through this; they can no longer forge one through `assert`. The gate itself is untouched (still a plain `starts_with` rejection, no allowlist).

The uncommitted overlay path (a transaction's pending rules) is unchanged — it stages into in-memory `Changes`, never hitting the tree write gate.

## Tests

- `it_rejects_public_writes_to_the_reserved_rule_namespace` — public Assert/Replace/Retract of both `dialog.rule/*` attributes are all refused with `ReservedAttribute` (proves the gap is closed).
- `it_round_trips_a_privileged_rule_install` — a rule installed via the sanctioned path is found by the `conclusion` discovery scan and its `source` body hydrates.
- Existing committed- and overlay-rule resolution tests adapted to the new install/remove API and passing.

Full native suite: 2062 passed, 0 failed.

## Migration

Attribute-name space, not on-disk format — no format version bump. Pre-existing `db.rule/*` data won't be discovered under the new names, but rules are re-installable, so regeneration is the migration.

### DIFF
--- a/notes/layered-rule-resolution.md
+++ b/notes/layered-rule-resolution.md
@@ -10,7 +10,7 @@ the same way. This note records how that works and why.
 Each layer in the stack is a query source:
 
 - **Durable layer** — one per branch in scope. Facts come from the
-  branch's committed tree; rules come from `db.rule/*` facts on that
+  branch's committed tree; rules come from `dialog.rule/*` facts on that
   tree.
 - **Transient layer** — the per-query `Changes` overlay (`.with(...)`
   and a transaction's pending writes). Facts and rules both come from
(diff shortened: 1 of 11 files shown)


Gozala:
Superseded by #420: rules install/uninstall through plain assert/retract (Statement impls), with the reserved-namespace integrity enforced by the hydration-time content-address check rather than a privileged write rail. The db.rule/* -> dialog.rule/* rename is included there.

PR #410, opened 2026-07-29
feat(dialog): preserve service errors and install staged revisions

jackddouglas:
## Summary
- preserve bounded structured HTTP service errors from UCAN redeem through S3, memory/archive/blob, storage/artifact, and repository pull/push errors
- add `Branch::install` to copy an exact staged revision, every reachable tree/spilled-value block, and referenced blob into another mounted storage environment without publishing a branch head or minting a commit
- cover native redeem/error propagation and volatile-to-durable exact installation; keep focused tests wasm-compatible

## Stack
- base: `feat/storage-clone` (PR #407)

## Verification
- `cargo test -p dialog-effects -p dialog-storage -p dialog-search-tree -p dialog-artifacts -p dialog-remote-s3 -p dialog-remote-ucan-s3 -p dialog-repository`
- `cargo check -p dialog-remote-s3 -p dialog-remote-ucan-s3 -p dialog-repository --target wasm32-unknown-unknown --tests`
- `cargo clippy --all --all-targets --all-features -- -D warnings`
- `cargo fmt --all --check`

Wasm test binaries compile, but browser execution in this sandbox could not start the shared wasm-bindgen driver (`failed to bind port during startup`).


rust/dialog-repository/src/repository/branch/install.rs
```diff
@@ -0,0 +1,416 @@
...
+}
+
+impl InstallRevision<'_> {
+    /// Copy the exact revision and all content it references from `source` to
+    /// `destination` without minting a new revision.
+    pub async fn perform<Source, Destination>(
+        self,
+        source: &Source,
+        destination: &Destination,
+    ) -> Result<Revision, InstallRevisionError>
```
Gozala: Unstated contract for `perform` methods is that it only takes `<Env: Provider>(env: &env) → Result<Ok, Error>` if you want any other configuration that should be done through builder like interface. E.g I would prefer if we extended our export functionality so we could export repo into non CSV format and then import it. If I recall correctly we even make export format pluggable.

Gozala: See details here https://github.com/dialog-db/dialog-db/blob/main/rust/dialog-repository/src/repository/branch/export.rs

rust/dialog-repository/src/repository/branch/install.rs
```diff
@@ -0,0 +1,416 @@
...
+
+impl Branch {
+    /// Prepare to install `revision` from this branch's storage into another
+    /// storage environment.
+    ///
+    /// The destination must already have the same repository DID mounted. The
+    /// caller may publish the returned revision only after this operation
+    /// succeeds; installation itself deliberately leaves branch memory cells
+    /// untouched.
+    pub fn install(&self, revision: Revision) -> InstallRevision<'_> {
```
Gozala: I find name very confusing, maybe we can call it `duplicate` or something like that ? Better yet if we can build this functionality through existing export / import API.

rust/dialog-effects/src/archive.rs
```diff
@@ -253,11 +254,21 @@ pub enum ArchiveError {
     #[error("Storage error: {0}")]
     Storage(String),
 
+    /// A remote service returned a non-success HTTP response.
+    #[error("{0}")]
+    ServiceResponse(#[source] ServiceResponseError),
+
```
Gozala: This feels like leaking provider implementation details, instead of mapping errors into existing set. 
Specifically archive implementation has no service or HTTP semantics. I think proper solution is to extend error with a semantic variants so errors could be mapped to them.

I think what we should be doing is mapping status code range to one of the errors that already exist. We could potentially add more error categories if we lack coverage. We could serialize status code as string or if we do need it for handling specific errors we can consider adding something to the errors also.

rust/dialog-effects/src/memory.rs
```diff
@@ -261,6 +262,10 @@ pub enum MemoryError {
     #[error("Storage error: {0}")]
     Storage(String),
 
+    /// A remote service returned a non-success HTTP response.
+    #[error("{0}")]
+    ServiceResponse(#[source] ServiceResponseError),
+
```
Gozala: Same as https://github.com/dialog-db/dialog-db/pull/410/changes#r3754029240

Gozala (changes requested):
I think this PR bundles two things worth decoupling:

1. Extending capability provider errors
2. Adding functionality to export blocks from one archive and import them into another

I think 1st one need redesigning, specifically it's not clear how consumer could handle `ServiceResponse` error type other than just printing it out. Not all providers have HTTP semantics or services. I think better approach would be to extend error variant into categories of errors that can occur that when encountered caller can choose to do one thing or the other. It's fare to say that existing error variants are kind of not great already because difference between storage and io error is unclear, but I would much rather improve it than make it worse.

On the 2nd I'm mostly happy with it, except I find name confusing and would very much prefer if we can fit this flow into import / export functionality that already exists, which was kind of designed to allow exporting / importing in different formats.

Gozala (approved):

Gozala:
Digging into the error side, I found evidence that sharpens what I asked for. Also noting where the rest of my review stands.

## `ServiceResponse` is the fourth escape hatch, and tonk proves it

There *are* real consumers downstream — `classify_pull` (`rust/tonk-worker/src/router/join.rs:1889`), `classified_service_failure` (`router/sync.rs:295`) — so the fidelity is not unused. But look at what every one of them does with it: collapse status/code into a small enum and discard the rest.

```rust
// rust/tonk-workspace/src/sync.rs:89
Some("CREDENTIAL_REVOKED" | "DEVICE_REVOKED")       => SyncFailure::Revoked,
Some("SYNC_CONFLICT")                               => SyncFailure::Conflict,
Some("SYNC_UNAVAILABLE" | "REVOCATION_UNAVAILABLE") => SyncFailure::Unavailable,
_                                                   => SyncFailure::Unknown,
```

Nobody keeps the HTTP detail. Three crates each re-derive that `CREDENTIAL_REVOKED | DEVICE_REVOKED` means revoked; `"CREDENTIAL_REVOKED"` appears as a bare literal 16 times across 4 files with no shared definition, so a typo compiles and a service-side rename silently degrades every match to `_`. And `join.rs:1903` matches on `response.status == 401 || response.status == 403` — HTTP semantics in a UI-facing worker, exactly what we are trying to avoid.

Tonk's empirically-derived union is small: revoked, conflict, unavailable, audience-mismatch, claim-failed, unknown. That is close to the variant set these errors should have had.

## The organizing axis

Not a variant per status code — the opposite. Roughly 4XX/5XX: **is the caller's input wrong, or did we fail?**

| 4XX — caller's input is wrong | 5XX — we failed |
| --- | --- |
| `DigestMismatch`, `VersionMismatch`, `NotFound`, everything in `AuthorizeError` | `Storage` — the backend broke |

Two things fall out. `ExecutionError` sits on neither side, which is the tell that it is not a failure mode — it wants deleting and its call sites redistributing. And `ArchiveError::Io` is 5XX but duplicates `Storage`: it is constructed both from `S3Error` (`remote-s3/error.rs:49`, a socket) and from a filesystem read (`fs/blob.rs:55`, a disk). Dumping ground, not a category.

So: map `ServiceResponseError` at the `dialog-remote-ucan-s3` boundary — `CREDENTIAL_REVOKED`/`DEVICE_REVOKED` → `Revoked`, `AUDIENCE_MISMATCH` → `InvalidAudience`, 401/403 → the authorize variants, and **5XX plus anything unrecognized → `Storage`**. I considered a separate `Unclassified` case and it collapses: if we cannot attribute a failure to the caller, we cannot blame the caller for it. That is also the safe direction to be wrong in. The raw code rides inside the 5XX variant rather than justifying a third category.

Then `find_service_response` and the three `ServiceResponse` variants go away, and tonk's classifiers become enum matches with no string literals.

## Scope — this should not block the PR

#423 already landed the `AuthorizeError::Denied(String)` split (10 named variants, mirroring `dialog_ucan_core`'s `CheckFailed`), and #422 deleted the dead `StorageError::Io`. So the targets to map *into* now exist.

What I would ask of this PR is only: **do not add the fourth escape hatch.** Map into those variants. The effect-error rework (`Io`, `ExecutionError`, ~50 sites) lands separately.

## Where the rest of my review stands

`install` is superseded by #427, which I am merging into this branch: it replaces `install` with `repo.snapshot(rev).export()` / `repo.import(items)`, which addresses all three of my other comments — the two-environment `perform` (each half now takes the one environment its side needs), the name, and the export/import framing. Export additionally handles spills and blobs, which the artifact exporter did not.

One thing I flagged there and did not resolve: sharing `Exporter`/`Importer` generically between the artifact path and the block path. Their `perform` bodies genuinely differ — artifact import commits and mints a revision, block import deliberately does neither — so it is follow-up, not a precondition.

That leaves the error modelling as the only live item from my review.

PR #413, opened 2026-07-31
perf(operator): memoize proven delegation chains

jackddouglas:
Every authorized remote read builds its UCAN invocation from scratch. On a
wasm deployment that measures at ~130-170ms of client-side work per request:
a certificate-store BFS issuing two IndexedDB prefix queries per chain hop
(each a fresh transaction, decoding and re-verifying every certificate), a
chain rebuild that re-encodes and re-hashes every delegation for its CID,
a container encode that re-serializes the same delegations again, and a
fresh `reqwest::Client` per request at two call sites. During a cold read of
a synced branch — hundreds of node fetches — this dominates wall clock.

The proven chain only changes when a certificate is retained or the
operator rotates. Everything beneath the per-request nonce is cacheable.

## Changes

One commit per crate; each intermediate state builds.

**`perf(dialog-operator): memoize proven delegation chains`** — `Operator`
gains a `ProofMemo` consulted by its `Provider<Authorize<P>>` impl, generic
over `Protocol`. Key: (capability subject routing to the certificate store,
operator DID, access subject, command segments) — scope parameters are
deliberately excluded, since they change per request (e.g. the digest on
`/archive/get`) and would give a 0% hit rate. A hit does not skip
verification: every cached certificate is re-verified in memory against
*this* request's access (command attenuation, policy predicates, intersected
validity ranges), and the chain's window must still contain now + 60s. This
matters because `Certificate::verify` never looks at the subject — a test
pins that with a powerline root-command grant that would verify for any
subject. Invalidation: `Retain` drops the memo whole (a powerline
certificate can complete a chain for any subject, so no guessing), rotation
misses via the key, and a 60s retention TTL bounds staleness toward
revocation. Time is injected (`recall`/`remember` take `now`) so retention
and expiry are tested deterministically.

**`perf(dialog-ucan-core): compute a delegation's encoding once`** — a
delegation is immutable after construction, so its encoded bytes and CID are
computed once behind a `OnceLock` instead of per `DelegationChain::push` and
again per `Container::from(&InvocationChain)`.

**`perf(dialog-remote-s3): share one HTTP client across requests`** — one
lazily-built client replaces the per-request `Client::new()` in
`UcanAuthorization::redeem` and `Permit::send`. `thread_local` on wasm
(the wasm client is not `Send`), a static on native.

## Tests

New, all `#[dialog_common::test]`:

- `it_proves_once_across_repeated_authorizations`
- `it_reproves_after_retaining_a_new_certificate`
- `it_reproves_when_the_cached_chain_nears_expiry`
- `it_scopes_the_memo_by_principal_and_certificate_store`
- plus memo-internal retention/expiry cases and a counting-encoder check
  that a delegation encodes once

`cargo test --workspace --lib --tests` green; per-crate clippy
`--all-targets --all-features` clean; `cargo fmt --all`; wasm32 check clean
including the `thread_local` client arm. The
`dialog-capability` doctest failure at `attenuate.rs:30` is pre-existing —
it fails identically at the base revision.

## Review notes

- **No 403-drop of the memo.** `CREDENTIAL_REVOKED` surfaces in
  `UcanAuthorization::redeem`, which holds no handle on the operator that
  produced the authorization; threading one through the Site/Fork machinery
  was judged not worth it. The 60s TTL is the staleness bound — the service
  remains the enforcement point regardless.
- `Retain` clears the whole memo rather than affected entries. Retains are
  rare (operator build, explicit delegation); narrowing would mean reasoning
  about powerline reach.
- Recall clones cached certificates because `Proof::push` takes ownership —
  a few hundred bytes per hop. If it shows in a profile, `push` taking
  `Arc<P::Certificate>` removes it.
- The wasm-side latency win is not measured here; native tests pin the
  mechanics (one prove per window, invalidation, single encode).

### DIFF
--- a/rust/dialog-operator/src/operator.rs
+++ b/rust/dialog-operator/src/operator.rs
@@ -5,11 +5,14 @@
 mod access;
 mod builder;
 mod fork;
+mod memo;
 mod space;
 #[cfg(test)]
 mod test;
 
 pub use builder::{OperatorBuilder, OperatorError};
+
+use memo::ProofMemo;
 
 use crate::Authority;
 use dialog_capability::{Capability, Provider};
@@ -58,6 +61,10 @@ pub struct Operator<S: Clone> {
 
     /// Network dispatch for fork invocations.
     network: Network,
+
+    /// Chains already proven against `storage`, so that repeated
+    /// authorizations do not walk the certificate store again.
+    memo: ProofMemo,
 }
 
 impl<S: Clone> Operator<S> {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1280,7 +1280,6 @@ dependencies = [
  "hyper",
  "hyper-util",
  "ipld-core",
- "reqwest",
  "serde",
  "serde_bytes",
  "serde_ipld_dagcbor",
--- a/rust/dialog-capability/src/access.rs
+++ b/rust/dialog-capability/src/access.rs
@@ -28,6 +28,13 @@ use thiserror::Error;
 pub trait Scope {
     /// The subject (resource) this scope applies to.
     fn subject(&self) -> &Did;
+
+    /// The command being requested, as ability path segments.
+    ///
+    /// Two scopes that differ only in their parameters share a command,
+    /// so this names the access independently of the arguments any one
+    /// invocation carries.
+    fn command(&self) -> &[String];
 }
 
 /// Derive an access scope from an invocable capability.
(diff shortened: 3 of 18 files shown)


rust/dialog-operator/src/operator.rs
```diff
@@ -58,6 +61,10 @@ pub struct Operator<S: Clone> {
 
     /// Network dispatch for fork invocations.
     network: Network,
+
+    /// Chains already proven against `storage`, so that repeated
+    /// authorizations do not walk the certificate store again.
+    memo: ProofMemo,
```
Gozala, line 67: I feel like `memo` is too generic of a name in this context, something like `proofs` or `access` would make a lot more sense.

Gozala (approved):
This is great! Thanks for doing it. I do wonder however if caching could be more deeply integrated inside certificate store perhaps so that separate memo stuff will be unnecessary.

PR #414, opened 2026-07-31
perf(search-tree): single-flight node fetches and read ahead on range scans

jackddouglas:
A branch synced to a remote reads locally-missing tree nodes through
`NetworkedIndex`: one authorized round trip per node (~350ms cold on a wasm
deployment). Two compounding problems:

1. **No single-flight.** Concurrent misses for the same key each fetch. The
   storage cache layer explicitly tolerates this — harmless for local reads,
   expensive for authorized remote fetches. During a cold load, many
   concurrent queries descend the same cold tree and re-fetch the same upper
   nodes.
2. **A strictly serial walk.** The walker awaits one node per iteration;
   there is no concurrency anywhere on the read path, while the write path
   already uploads with `buffer_unordered(UPLOAD_CONCURRENCY = 16)`.

## Changes

**`perf(search-tree): single-flight the node cache's misses`** — the
arbitration lives in `dialog_search_tree::Cache` (the per-branch node
cache), because it is the one layer every read passes through — the walker's
own gets and the prefetch below included. `Fetches` is a registry of
in-flight claims; the claimant runs the caller's fetcher, concurrent misses
await a receiver. The uncontended path is deliberately free of allocation: a
claim starts as a bare map entry, and the `broadcast` outcome channel is
only materialized the moment a second caller actually arrives — so the
common miss with no waiter pays a map insert/remove and nothing else, and
`publish` clones the value only when someone is listening. (An earlier
eager-channel draft cost a measured 6-12% across the native search-tree
benches; the lazy variant benches at parity with the pre-change baseline.)
A claim withdraws itself on `Drop` — so a failed fetch or a cancelled reader
clears the key — and waiters that see a closed channel loop back, re-check
the cache, and fetch themselves if the leader failed. A stored `Shared`
future does not work here: the fetch future borrows the accessor and is not
`'static`. No new dependencies; the mutex guard is never held across an
await, and the `?Send` cfg split is untouched.

**`perf(search-tree): read a range scan's siblings ahead of the walk`** —
when the range stream descends into child N of an index node, it queues
cache-warming reads for children N+1.. up to 16 in flight, drained only
while the walk's own read is pending (`while_warming` poll adapter). The
walk awaits exactly the nodes it awaited before, in the same order — all 126
pre-existing ordering tests pass unmodified. Point lookups do not prefetch:
descent is serial by nature and fan-out would fetch nodes the lookup skips.
Warming goes through the cache, hence through single-flight, so a prefetched
sibling and the walk's own read of it are one backend fetch. Dropping the
stream stops issuing prefetches.

**`perf(repository): probe a branch's root through its node cache`** — the
eager root probe was the one read per query that bypassed the node cache,
so concurrent cold queries duplicated it; 24 lines to route it through.

## Tests

New, all `#[dialog_common::test]`, all native-runnable:

- `it_deduplicates_concurrent_fetches_of_one_node`
- `it_clears_the_inflight_entry_when_a_fetch_fails`
- `it_lets_waiters_retry_a_failed_fetch`
- `it_opens_an_outcome_channel_only_when_a_waiter_arrives`
- `it_warms_sibling_nodes_during_a_range_scan`
- `it_does_not_prefetch_on_point_lookups`

Each was verified to bite: `PREFETCH_CONCURRENCY = 0` fails the overlap
assertion; bypassing single-flight fails the dedupe test (8 backend reads)
and makes prefetch duplicate work; a no-op `Drop` fails the registry-clear
test. `cargo test` green on dialog-search-tree / dialog-storage /
dialog-repository plus dialog-artifacts and dialog-query as read-path
consumers; the search-tree suite ran three times with identical results.
Clippy `--all-targets --all-features` clean on touched crates; fmt clean.

## Review notes

- Prefetch can over-fetch on narrow scans: a scan that stops after two
  siblings may have warmed up to 14 nodes it never reads. If narrow selects
  dominate real workloads, a ramp (1, 2, 4, ...) would cut the waste —
  worth measuring before tuning.
- No timeout on a claim: a leaked, never-polled stream holding a claim makes
  waiters wait with it. Dropping the stream withdraws the claim immediately;
  there is no cross-target timer in the async helpers to do better without a
  new dependency.
- `NetworkedIndex`'s non-tree reads (blob, commit/pull paths) are not
  deduplicated here; out of scope.
- The latency win is not measured on wasm here; the native tests pin the
  mechanics (one fetch per key, overlapping sibling reads, unchanged yield
  order).

### DIFF
--- a/rust/dialog-search-tree/src/accessor.rs
+++ b/rust/dialog-search-tree/src/accessor.rs
@@ -47,6 +47,26 @@ where
         Self { cache, storage }
     }
 
+    /// Loads the node at `hash` into the cache, without decoding it or
+    /// reporting what happened.
+    ///
+    /// This is a read nobody is waiting on: it makes the node local so that a
+    /// later [`get_node`](Self::get_node) is served from the cache, and it is
+    /// deduplicated against any read of the same node already in flight.
+    /// Nothing observes its outcome, so a node that is missing or fails to
+    /// load is left to the read that actually needs it.
+    pub(crate) async fn warm(&self, hash: Blake3Hash) {
+        let _ = self
+            .cache
+            .get_or_fetch(&hash, async |key| {
+                self.storage
+                    .retrieve(key)
+                    .await
+                    .map(|maybe_bytes| maybe_bytes.map(Buffer::from))
+            })
+            .await;
+    }
+
     /// Retrieves a node by its content hash.
     ///
     /// Checks the cache first, then the storage backend. Returns an error if the
--- a/rust/dialog-search-tree/src/accessor.rs
+++ b/rust/dialog-search-tree/src/accessor.rs
@@ -83,3 +83,41 @@ where
             .map(|buffer| PersistentNode::new(buffer))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    #![allow(unexpected_cfgs)]
+
+    use anyhow::Result;
+    use dialog_common::Blake3Hash;
+    use futures_util::future::join_all;
+
+    use crate::{
+        Accessor, Cache, ContentAddressedStorage, PersistentNode, helpers::ObservingBackend,
+    };
+
+    #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+    wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_dedicated_worker);
+
+    #[dialog_common::test]
+    async fn it_deduplicates_concurrent_fetches_of_one_node() -> Result<()> {
+        let backend = ObservingBackend::new();
+        let mut storage = ContentAddressedStorage::new(backend.clone());
+
+        let bytes = b"the bytes of one node".to_vec();
+        let hash = Blake3Hash::hash(&bytes);
+        storage.store(bytes, &hash).await?;
+
+        let accessor = Accessor::new(Cache::new(), storage);
+        backend.reset();
+
+        let reads = join_all((0..8).map(|_| accessor.get_node(&hash))).await;
+
+        for read in reads {
+            let _: PersistentNode<[u8; 4], Vec<u8>> = read?;
+        }
+        assert_eq!(backend.read_log(), vec![hash]);
+
+        Ok(())
+    }
+}
--- a/rust/dialog-search-tree/src/walker.rs
+++ b/rust/dialog-search-tree/src/walker.rs
@@ -1,12 +1,15 @@
 use std::{
+    future::Future,
     marker::PhantomData,
     ops::{Bound, RangeBounds},
+    task::Poll,
 };
 
 use async_stream::try_stream;
 use dialog_common::{Blake3Hash, ConditionalSend, ConditionalSync, NULL_BLAKE3_HASH};
 use dialog_storage::{DialogStorageError, StorageBackend};
 use futures_core::Stream;
+use futures_util::{StreamExt as _, stream::FuturesUnordered};
 use nonempty::NonEmpty;
 use rkyv::{
     Deserialize,
@@ -20,6 +23,14 @@ use crate::{
     Accessor, ArchivedNodeBody, DialogSearchTreeError, Entry, Key, Link, PersistentNode,
     SymmetryWith, Value, into_owned,
 };
+
+/// How many sibling reads a range scan keeps in flight while it walks.
+///
+/// A scan reads a whole run of siblings, one after another, and each read that
+/// misses locally can cost a round trip. Reading ahead of the walk turns a run
+/// of round trips into an overlapping few, and bounding it keeps a scan that
+/// stops early from having fetched much it never looked at.
+const PREFETCH_CONCURRENCY: usize = 16;
 
 /// A traversal mechanism for walking through a tree structure.
 pub struct TreeWalker<Key, Value>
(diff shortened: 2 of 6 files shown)


rust/dialog-search-tree/src/walker.rs
```diff
@@ -21,6 +24,14 @@ use crate::{
     SymmetryWith, Value, into_owned,
 };
 
+/// How many sibling reads a range scan keeps in flight while it walks.
+///
+/// A scan reads a whole run of siblings, one after another, and each read that
+/// misses locally can cost a round trip. Reading ahead of the walk turns a run
+/// of round trips into an overlapping few, and bounding it keeps a scan that
+/// stops early from having fetched much it never looked at.
+const PREFETCH_CONCURRENCY: usize = 16;
```
Gozala, line 33: Seems like this should be a default value for the `TreeWalker` property so it's choice consumer has as opposed to being static.

rust/dialog-search-tree/src/accessor.rs
```diff
@@ -47,6 +47,26 @@ where
...
 
+    /// Loads the node at `hash` into the cache, without decoding it or
+    /// reporting what happened.
+    ///
+    /// This is a read nobody is waiting on: it makes the node local so that a
+    /// later [`get_node`](Self::get_node) is served from the cache, and it is
+    /// deduplicated against any read of the same node already in flight.
+    /// Nothing observes its outcome, so a node that is missing or fails to
+    /// load is left to the read that actually needs it.
+    pub(crate) async fn warm(&self, hash: Blake3Hash) {
```
Gozala, line 58: I would call it `preload` or something, because `warm` is not really a verb and therefor confusing here.

rust/dialog-search-tree/src/walker.rs
```diff
@@ -112,7 +124,21 @@ where
...
+                                // The scan will walk this node's remaining
+                                // children in turn, so start reading them now
+                                // and let them land in the cache while the
+                                // walk descends into the first of them.
+                                for sibling in index.links.iter().skip(child_index + 1) {
+                                    if warming.len() >= PREFETCH_CONCURRENCY {
+                                        break;
+                                    }
+                                    warming.push(accessor.warm(<&Blake3Hash>::from(&sibling.node).clone()));
+                                }
```
Gozala, line 136: This goes against partial query driven replication and will end up preloading adjacent nodes even when if they will never end up being accessed by query. This could be a defensible strategy yielding better performance for various workflows, but it really is a replication strategy that application chooses and not the default which is what current implementation does.

It also would be good idea to perhaps evaluate this against alternative designs like replicating certain blocks on pull instead of doing it on query.

Gozala: Now that I think about it, I'm realizing that our range scans could be made more intelligent without preloads. Specifically we don't have to do a depth first traversal, instead we could fan out and concurrently load all nodes covered by the range, which I suspect will offer same gains but be more precise at what is being loaded as opposed to just loading extra nodes. Maybe this is what PR actually does and I have misunderstood it, in which case I think we should not just stop prefetching we should probably queue up reads we need to do and let the accessor drain it.

Gozala (changes requested):
Consolidating my inline notes, since they add up to one design point rather than four separate nits.

**The win here is real: concurrency.** Loading needed nodes in parallel instead of one-at-a-time depth-first is a clear improvement, and the point-lookup path stays correctly on a single root-to-leaf line (`it_does_not_prefetch_on_point_lookups` proves peak-in-flight stays at 1). I want to keep that.

**But the PR conflates two different things, and only one of them is a pure win:**

- **Concurrency** — loading multiple nodes the scan *provably needs* at the same time. No policy implications.
- **Speculation** — `skip(child_index + 1)` warms *all* trailing siblings at each level, capped at 16, without consulting the scan's upper bound. A bounded or early-stopping range scan therefore dispatches fetches for siblings it never visits, and through a networked store those `preload`/`warm` calls replicate blocks locally as a side effect of a query.

That second half is what goes against partial, query-driven replication: it pulls adjacent nodes into the local store even when the query will never touch them. It's a defensible opt-in strategy for some workloads, but it shouldn't be the default — the current behavior (replicate only what the query reaches) is the right default, and speculation is a policy the application should choose.

**Concrete asks:**

1. Make the prefetch degree a `TreeWalker` field with a default, not a `const`, so a consumer can set it — including `0` to turn speculation off. (`walker.rs:33`)
2. Rename `warm` → `preload`; `warm` doesn't read as a verb here. (`accessor.rs:58`)
3. There's no test bounding over-fetch on a *bounded* range scan. `it_warms_sibling_nodes_during_a_range_scan` scans the whole tree, where every sibling genuinely is needed, so it can't catch over-reading past a range end. A range scan with a tight upper bound would over-read and nothing would flag it.

**The design I'd rather converge on** (my second inline note): don't speculate at all — at each index level select only the child links whose separator range overlaps `[lower, upper]` and load *those* concurrently. Same latency win, zero over-fetch, and every block loaded is one the query provably needs, so it stays within query-driven replication. Concretely: replace the depth-first "descend one, warm blind siblings" loop with a range-directed concurrent fan-out of exactly the covered frontier, drained through the accessor. It's more work than blind sibling warming (the walker has to compute the covered child set per level), but it's precise.

Also worth weighing against replicating certain blocks on *pull* rather than on query — that keeps the read path free of any replication policy at all.

None of this blocks the concurrency mechanism itself; it's about not shipping a speculative-replication default under it.

PR #424, opened 2026-08-12
perf(search-tree): amortized flushing, columnar node format, and read-path memoization

Gozala:
Distills the search-tree engine wins from the performance campaign into the lowest layer of the stack. The whole workspace builds with only this PR applied, so it can be reviewed and measured in isolation.

## What changes

- **Amortized flushing** — deferred inserts accumulate in a byte-capped novelty buffer and settle in bulk, instead of reshaping the tree on every write.
- **Compressed quiet-check node shaping** — the run/quiet split that decides node boundaries now works over a compressed summary of the incoming piece set.
- **Bulk plant** — a batch of ops plants the spine bottom-up in one pass.
- **Slot-tracked reads** — the walker carries pending winners per leaf and merges them, avoiding redundant re-reads on the read path.
- **Memoized decoded keys** — persistent nodes cache their decoded key arena (`DecodedKeys` with `lower_bound` / `binary_search`) so repeated lookups don't re-parse.
- **Node validity as a construction invariant** — `PersistentNode` is built only through `TryFrom<Buffer>` (validates untrusted bytes) or `TryFrom<&PersistentNodeBody>` (serializes typed, valid-by-construction data), so `body()` is infallible and the validation proof rides the type rather than a runtime marker.
- **Validation harness + property suites** — `validate.rs` plus `tests/adversarial.rs` and `tests/program.rs` pin tree-shape convergence.

## Measurement

Criterion A/B, this branch vs `main`, on a quiesced machine (load average low single digits, both sides pre-built, measured back to back). Every row below is `p < 0.05` with a 95% confidence interval; negative = faster.

**Point lookups (`get`)** collapse, and this also clears the large-tree point-lookup regression that #399's 64 KiB leaves had introduced on `main`:

| size | change | 95% CI |
| --- | --- | --- |
| 10 | -43.3% | [-43.7%, -42.9%] |
| 100 | -81.2% | [-81.3%, -81.1%] |
| 1000 | -65.5% | [-65.8%, -65.2%] |
| 10000 | **-97.7%** | [-97.7%, -97.7%] |

**Writes at scale** (the 10 000-row cases, across sequential / random / batched / mixed):

| bench | change | 95% CI |
| --- | --- | --- |
| insert_sequential/10000 | -44.3% | [-44.7%, -44.0%] |
| insert_random/10000 | -54.7% | [-55.0%, -54.4%] |
| insert_batch/sequential/10000 | -53.7% | [-53.9%, -53.4%] |
| insert_batch/batched/10000 | **-92.5%** | [-92.7%, -92.4%] |
| mixed_batch/sequential/10000 | -68.1% | [-68.2%, -67.9%] |
| mixed_batch/batched/10000 | **-96.5%** | [-96.6%, -96.5%] |

**Small-batch tradeoff.** The only regressions are tiny sequential batches below the amortization threshold, where the byte-capped buffer's bookkeeping does not yet pay for itself: `insert_batch/sequential/100` +1.9% [+1.4%, +2.3%], `/1000` +3.9% [+3.3%, +4.6%], `mixed_batch/sequential/100` +1.3% [+0.6%, +1.9%]. Sub-4%, and the same paths turn into 44-96% wins by size 10 000.

Full table across all sizes (get, insert_sequential, insert_random, insert_batch, mixed_batch) confirms the shape: micro-overhead on tiny sequential ops, large wins at scale and on the read path.

This PR is the base of a three-PR stack; the artifacts/query/repository migration and the dev-baseline measurement crate build on top.

### DIFF
--- a/rust/dialog-common/src/buffer.rs
+++ b/rust/dialog-common/src/buffer.rs
@@ -1,4 +1,4 @@
-use std::any::Any;
+use std::any::{Any, TypeId};
 use std::fmt::{Formatter, Result as FmtResult};
 use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::{Arc, OnceLock};
@@ -26,6 +26,7 @@ struct BufferInner {
     bytes: AlignedVec,
     hash: OnceLock<Blake3Hash>,
     decoded: OnceLock<Decoded>,
+    validated: OnceLock<TypeId>,
     touches: AtomicU32,
 }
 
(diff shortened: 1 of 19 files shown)


Gozala:
**Stack** (review bottom-up):
1. #424 search-tree engine ← base of stack (targets `main`)
2. #425 artifacts/query/repository migration (targets #424)
3. #426 dev-baseline + soak + hardening (targets #425)

The tip of #426 is byte-identical to the fully-verified distilled state; full native suite is green (2159 passed, 7 skipped).

PR #425, opened 2026-08-12
perf(artifacts): borrowed materialization, columnar keys, and buffered commits

Gozala:
Stacked on the search-tree engine PR. Migrates the artifact / query / repository stack onto the campaign's zero-copy materialization and key-in-tree encoding.

## Why these three crates move together

`ArtifactView` (borrowed materialization) replaces owned `Artifact` at the crate boundaries shared by artifacts, query, and repository. Attempting to land them separately does not compile: query and repository both consume the view type directly (`Vec<Artifact>: Extend<ArtifactView>`, `Cause: From<&ArtifactView>`, stream item mismatches). They are one atomic API migration.

## What changes

- **Borrowed materialization** — `ArtifactView` lets reads project artifacts without copying out of the tree.
- **Columnar key encoding** — varkey column layout and URI encoding move the datum into the key; the history record footprint shrinks accordingly.
- **Buffered commit path** — `buffered.rs` drives the search-tree `insert_deferred` / `settle` flow so a commit flushes once at the end instead of per-op.
- **Query scan adaptation** — `all` / `only` / `match` / `helpers` consume the borrowed view.
- **Repository branch plumbing** — layer, branch select / subscription / pull / commit adapt to the new artifact and view types.

## Measurement

Criterion, `dialog-query` benches (`--features helpers`). `main`, `#424` (`pr1-search-tree`), and this branch's tip (the full stack: #424 + #425 with the fix and all scan optimizations) were each measured back to back on a quiesced machine (each run gated to load average low single digits). Median times; negative = faster.

| bench | main | #424 | stack | vs main | vs #424 |
| --- | --- | --- | --- | --- | --- |
| query_join/100 | 1.73ms | 1.08ms | 693us | **-59.9%** | -35.8% |
| query_join/1000 | 13.67ms | 11.34ms | 6.80ms | **-50.2%** | -40.0% |
| query_memory/100 | 205us | 189us | 157us | -23.7% | -17.3% |
| query_memory/1000 | 898us | 877us | 635us | -29.3% | -27.5% |
| query_memory/10000 | 8.02ms | 7.56ms | 5.81ms | **-27.5%** | -23.2% |
| query_reads/100 | 198us | 181us | 156us | -21.3% | -13.6% |
| query_reads/1000 | 872us | 811us | 631us | -27.6% | -22.2% |
| query_reads/10000 | 7.80ms | 7.51ms | 6.15ms | **-21.2%** | -18.1% |

The stack is faster than pre-campaign `main` at every size (**-21% to -60%**), and faster than `#424` alone (**-14% to -40%**): the artifacts layer plus the fix and scan optimizations add a further win on top of the search-tree engine.

The 10k full-attribute-scan-and-materialize cases (`query_reads/10000`, `query_memory/10000`) are the ones the raw migration regressed: the URI parse memo it added cost +13% / +23% versus `#424`. With the fix and optimizations they are now faster than **both** `main` and `#424` (-21% / -28% vs main). The regression is not merely recovered but inverted.

### Regression fix and scan optimizations (commits 2-7)

**Root cause.** The raw migration added a thread-local URI parse memo to speed up joins, but it inverts into overhead on large scans: once a scan materializes more distinct entities than the 4096-entry cache holds, every row pays the parse plus the memo's bookkeeping and periodic wholesale clears.

The follow-up commits address it in six steps:

1. **Remove the memo, add `scan_owned`.** Restores a plain parse and routes the materialize-everything path (`SelectOwned`) through `ArtifactTreeExt::scan_owned`, which builds each row from the scan's single key parse (via the existing public `Artifact::from_key_ref_datum_value`). The view-based path is untouched.
2. **Skip the parse on read-back.** `Uri` stores the parser's normalized rendering (`Box<str>`) instead of a `url::Url`; strings read out of the index re-wrap via `from_stored` without re-parsing. Validation stays at the ingest boundaries. Sound because `url` normalization is idempotent (we store the parser's own output).
3. **Validate stored entities, skip corrupt rows.** Read-side validation classifies a non-canonical entity / non-UTF-8 column / invariant-breaking attribute as `CorruptEntry` and skips it (with a `tracing::warn`) instead of failing the whole query. Structural key corruption stays a loud error (keys live in hash-verified blocks). **Semantic change:** reads now carry an implicit "entity is a valid canonical URI" predicate; a future non-normalizing writer's rows would be silently skipped (warned) rather than failing the query.
4. **Canonicality prover.** A byte-scan fast path (`is_provably_canonical`) accepts a string only when the WHATWG parse is provably the identity (covers `did:key:…`, `user:…`, `blob:…`); anything unprovable falls back to the real parse, so strictness is exactly unchanged. Pinned by a 200k-sample generative test against `url`. Recovers the validation cost.
5. **Arena-handle scan keys.** Scan yields each key as a `KeyHandle` (an `Arc<DecodedKeys>` + index into the leaf's decoded-key arena) with no per-row key copy; `ArtifactView` carries the handle. Adds `PersistentTree::stream_range_handles`. This pushes 10k below the #424 baseline.
6. **Head-slot k-way merge.** Materializes the head slot through the merge so the branch overlay no longer re-slices its buffered novelty per step, and threads the borrowed key handles through the repository merge path. Resolves the ranked scan follow-ups.

Net: 10k scan-and-materialize now runs faster than pre-campaign `main` with full read-side validation active (measured above).

### Commit layout

Seven commits, reviewable in order:

1. `perf(artifacts)`: borrowed materialization, columnar keys, buffered commits (the migration)
2. `perf(artifacts)`: fix the 10k scan-and-materialize regression (remove the memo, add `scan_owned`)
3. `perf(artifacts)`: skip the URL parse for strings read back out of the index
4. `feat(artifacts)`: validate stored entities on read, skip corrupt rows
5. `perf(artifacts)`: prove canonicality by byte scan, parse only as fallback
6. `perf(search-tree)`: yield scan keys as arena handles, not per-row copies
7. `perf(repository)`: head-slot k-way merge; resolve remaining follow-ups

Commit 6 touches `dialog-search-tree` (adds `stream_range_handles`) even though it lands in this PR rather than #424; it is the artifact scan path's consumer of the new API, kept with the rest of the scan work for review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -991,6 +991,7 @@ dependencies = [
  "futures-util",
  "getrandom 0.2.16",
  "leb128",
+ "memchr",
  "rand 0.8.5",
  "rand_chacha 0.3.1",
  "rand_core 0.6.4",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -126,6 +126,7 @@ reqwest = { version = "0.12", default-features = false, features = [
     "json",
 ] }
 rexie = "0.6"
+memchr = "2"
 rkyv = "0.8"
 rsa = "0.9"
 s3s = "0.13"
--- a/rust/dialog-artifacts/Cargo.toml
+++ b/rust/dialog-artifacts/Cargo.toml
@@ -39,6 +39,7 @@ leb128 = { workspace = true }
 rand = { workspace = true }
 rand_core = { workspace = true }
 rand_chacha = { workspace = true }
+memchr = { workspace = true }
 rkyv = { workspace = true }
 serde = { workspace = true }
 serde-big-array = { workspace = true }
--- a/rust/dialog-artifacts/src/artifacts.rs
+++ b/rust/dialog-artifacts/src/artifacts.rs
@@ -73,7 +73,10 @@ use crate::tree::{
     ArtifactTree, ArtifactTreeExt, SpillCache, TreeStorageBridge, fetch_spilled_cached, spill_cache,
 };
 use crate::{
-    DialogArtifactsError, HASH_SIZE, Key, State, artifacts::selector::Constrained, make_reference,
+    DialogArtifactsError, HASH_SIZE, Key, State,
+    artifacts::selector::Constrained,
+    buffered::{BufferedBatch, SpineSlot, apply_buffered_reusing},
+    make_reference,
 };
 
 /// An alias type that describes the search tree that is used for each
@@ -114,6 +117,11 @@ where
     /// same large value skips the store fetch. Content-addressed, so it never
     /// serves stale bytes.
     spill_cache: SpillCache,
+    /// Carries the live buffered spine between commits (keyed by root hash,
+    /// so any out-of-band root change simply misses), sparing every commit
+    /// the root-frame decode and sealed-buffer bulk copies of a fresh open.
+    /// Shared across clones, like the index it accelerates.
+    spine: SpineSlot,
 }
 
 impl<Backend> Artifacts<Backend>
@@ -183,6 +191,7 @@ where
             storage,
             index: Arc::new(RwLock::new(index)),
             spill_cache: spill_cache(),
+            spine: SpineSlot::new(),
         })
     }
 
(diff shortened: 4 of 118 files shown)


Gozala:
Part of the perf stack: #424 (base) → **#425** → #426. Stacked on #424; the diff shown against `main` will include #424 until it merges.

PR #426, opened 2026-08-12
test(baseline): dev measurement crate, nightly soak, and hardening suites

Gozala:
Stacked on the artifacts migration PR. Adds the measurement tooling that backs the campaign's numbers. No production code paths change.

## What changes

- **`dialog-baseline`** — a native-only development crate (`publish = false`): sqlite-vs-dialog benches, SE-replay commit profiling, the convergence oracle, scale / census examples, and artifact fuzz. It exists so the perf claims can be re-measured on a quiesced machine.
- **Nightly soak workflow** — `.github/workflows/nightly-soak.yml` with adversarial / program / artifacts / converge arms, plus the matching `flake.nix` soak commands.
- **Artifacts hardening suite** — `dialog-artifacts/tests/hardening.rs` exercises the encoding boundary added in the artifacts PR.

A follow-up commit excludes the native-only `dialog-baseline` from the wasm test archive (its examples pull native-only symbols like tokio and the sqlite harness, which do not compile for wasm32), mirroring how `wbg-pool` is already excluded.

## Measurement

No production code paths change here, so there is no runtime delta to measure for this PR. It is the measurement surface itself: `soak:converge` gates tree-shape convergence (byte-identical roots across commit groupings), and the benches drive the before/after numbers reported on the two perf PRs below it (#424, #425).

### DIFF
--- a/.github/workflows/nightly-soak.yml
+++ b/.github/workflows/nightly-soak.yml
@@ -1,0 +1,84 @@
+name: 'Nightly Soak'
+
+# The standing convergence soak: the same tests CI runs at bounded scale,
+# nightly at the heavy knobs — many-seed adversarial manifests, the
+# artifacts-level fuzz, and a deep SE-log convergence replay. Every arm
+# runs the canonical-form validator at its checkpoints, so a red night
+# localizes to a level and node (see rust/dialog-search-tree/src/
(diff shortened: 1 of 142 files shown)


Gozala:
Top of the perf stack: #424 → #425 → **#426**. Measurement tooling only, no production code paths.

PR #436, opened 2026-08-14
feat(dialog-repository): retain delegations as data

Gozala:
Second slice of [delegations-as-data](https://github.com/dialog-db/dialog-db/blob/feat/delegations-as-data/notes/delegations-as-data.md), stacked on #433. UCAN delegations become synced data: retained into the branch tree, replicated by ordinary push/pull, queryable with the query engine.

## Mechanism

`branch.delegations().retain(chain)` decomposes every certificate into two synced pieces, landing in **one commit** so they cannot drift:

- **Slim facts** under reserved `dialog.ucan/*` attributes (audience, subject, issuer, command, and validity bounds when present) on the entity `blob:<hash>` of the encoded envelope. Ordinary version-controlled facts: all three orderings, history records, observed-remove merge, datalog-queryable. Powerline subjects store `ANY_SUBJECT` so lookups stay equalities.
- **The signed envelope as a blob** addressed by that same hash and recorded in the blob index, so push ships the bytes and replicas hydrate on demand. Policy, meta, nonce, and the signature are never decomposed; proof assembly reads the envelope back byte-identical.

`retract(chain)` mirrors it: fact retractions through the versioned apply path plus the blob-reference tombstone from #433, one commit, bytes untouched. Both operations are idempotent by content address (a re-retain/re-retract mints no revision). `retain_all` is the bulk one-commit form for imports.

## Plumbing

- `write_instructions`/`BufferedBatch::apply` gain a `WriteScope`: `Application` (default everywhere existing) keeps rejecting `dialog.*`; `Machinery` permits it, at the same trust level as the public `record()` path, so the decomposition is the only writer of these facts.
- `Commit` gains crate-private `machinery()` / `with_entries()` so blob-index edits seal in the same revision as the facts (entries also make a commit non-empty).
- `BlobRecord::entry()` / `retract_entry()` let blob-index edits ride a batch instead of editing the tree directly.

## Tests

Unit: decompose-and-read-back (facts via select, envelope via blob read, decode roundtrip), powerline encoding, idempotence, retract semantics, and a pin that application commits still cannot write `dialog.ucan/*`. Integration (`it_replicates_retained_delegations`): site B pulls, finds the delegation by a value-bound audience query, hydrates the envelope byte-identical, and sees the retraction after a second pull.

The certificate store still serves the operator's prove path; the prover switch is the next PR in the stack.

### DIFF
--- a/rust/dialog-artifacts/src/blob_index.rs
+++ b/rust/dialog-artifacts/src/blob_index.rs
@@ -70,6 +70,20 @@ impl BlobRecord {
         })
     }
 
+    /// The tree entry recording this blob reference, for callers appending
+    /// machinery entries to an open batch (see `BufferedBatch::record`) rather
+    /// than editing the tree directly through
+    /// [`put_blob`](BlobIndexExt::put_blob).
+    pub fn entry(self, hash: &Blake3Hash) -> (crate::Key, State<Datum>) {
+        (BlobKey::new(hash).into_key(), self.into_state())
+    }
+
+    /// The tombstone entry retracting a blob reference, the batch-entry
+    /// counterpart of [`retract_blob`](BlobIndexExt::retract_blob).
+    pub fn retract_entry(hash: &Blake3Hash) -> (crate::Key, State<Datum>) {
+        (BlobKey::new(hash).into_key(), State::Removed)
+    }
+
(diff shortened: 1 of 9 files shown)


Gozala:

PR #437, opened 2026-08-14
feat(dialog-repository): prove chains over retained delegations

Gozala:
Third slice of [delegations-as-data](https://github.com/dialog-db/dialog-db/blob/feat/delegation-prover/notes/delegations-as-data.md), stacked on #436. The chain search moves onto the retained-delegation facts, with measurements demonstrating the expected behavior.

## Mechanism

`branch.delegations().prove(principal, scope)` replaces the certificate-store shape of list-and-decode-everything with a slim-candidate walk:

- candidates are filtered on their facts (subject match, command cover, validity window) without touching an envelope;
- a **direct grant** (issuer = subject) is admitted the moment it is seen, so the common case reads and decodes exactly one envelope no matter how many delegations are retained;
- **indirect** candidates are deferred until the hop's direct candidates are exhausted, preserving the certificate-store walk's direct-first preference;
- each queue entry carries the principals on its path, so a cyclic delegation graph terminates by pruning rather than by depth bound alone.

## Parity

Twelve scenarios (direct, powerline, three chain shapes, wrong audience/subject/command, expired, self-auth, cyclic) each prove against **both** the legacy volatile certificate store and the tree walk, asserting agreement on verdict and chain length.

## Measurements

`benches/delegation_prove.rs` populates the legacy filesystem store, the legacy volatile store, and the delegation region with the same N certificates (verdicts asserted every iteration):

| scenario | legacy fs | legacy volatile | tree |
|---|---:|---:|---:|
| direct, N=32 | 751 µs | 31 µs | 23 µs |
| direct, N=512 | 12.2 ms | 492 µs | **21 µs** |
| direct, N=6144 | 210 ms | 6.0 ms | **25 µs** |
| miss (exhaustive), N=32 | 1.5 ms | 43 µs | 780 µs |
| miss, N=512 | 24 ms | 1.8 ms | 13 ms |
| miss, N=6144 | 308 ms | 187 ms | 180 ms |
| chain3 (intended topology) | 174 µs | 4.6 µs | 89 µs |

The accumulation pathology observed in the field (thousands of same-audience grants, every remote effect re-decoding all of them) goes from O(N) to O(1): flat ~25 µs at 6144 certificates vs 210 ms on the filesystem store. The tree walk beats the store native actually persists to in every scenario. Honest caveat the data surfaced: at small-N exhaustive misses the in-memory store wins on per-candidate constant factors (~25 µs of select machinery per candidate; a batching target, irrelevant at sub-ms absolutes).

The certificate store still serves the operator's prove path; switching it over with migration, and the resolved-chain cache that amortizes repeat proves per effect, are the remaining slices.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1382,6 +1382,7 @@ dependencies = [
  "async-stream",
  "async-trait",
  "base58",
+ "criterion",
  "dialog-artifacts",
  "dialog-capability",
  "dialog-common",
--- a/rust/dialog-repository/Cargo.toml
+++ b/rust/dialog-repository/Cargo.toml
@@ -69,6 +69,7 @@ wasm-bindgen-futures = { workspace = true }
 wasm-bindgen-test = { workspace = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+criterion = { version = "0.5", features = ["async_tokio"] }
 reqwest = { workspace = true }
 tempfile = { workspace = true }
 tokio = { workspace = true, features = [
@@ -79,6 +80,10 @@ tokio = { workspace = true, features = [
     "rt",
     "rt-multi-thread",
 ] }
+
+[[bench]]
+name = "delegation_prove"
+harness = false
 
 [lints.rust]
 # This cfg is used by the dialog_common::test proc macro for wasm inner tests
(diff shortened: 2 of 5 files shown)


Gozala:

PR #438, opened 2026-08-14
feat(dialog-repository): serve operator access from the profile branch

Gozala:
Fourth slice of [delegations-as-data](https://github.com/dialog-db/dialog-db/blob/feat/synced-access/notes/delegations-as-data.md), stacked on #437. The operator's authorization switches over to the synced delegation store, with migration from the legacy certificate stores.

## Mechanism

The crate graph forces an inversion: the tree-backed store lives in dialog-repository, which sits **above** dialog-operator, so the operator cannot construct it. Instead:

- **dialog-operator** gains a pluggable `AccessProvider`: by default `Prove`/`Retain` route to the storage certificate provider exactly as before; an installed override routes them elsewhere. `Authorize` now routes its proof through the operator's own `Prove` provider so the override serves the per-effect hot path. The access providers narrow from generic protocols to `Ucan` (the only protocol implemented).
- **dialog-repository** provides the override: `SyncedAccess` serves proofs through the tree walk over a branch's `dialog.ucan/*` facts and lands retains as synced facts. `synced_access(profile, operator)` installs it: opens the profile repository's own branch (always locally present, provable with the locally-minted profile-to-operator delegation, so a fresh device can pull it before holding anything else), migrates the legacy store's certificates into it, and returns the operator with the override installed.
- **Migration** rides a new `Export` access effect: `CertificateStore::export` enumerates a legacy store (filesystem walk, volatile scan, IndexedDb `get_all`), forwarded through Space/Storage/Router like `Prove`. It always reads the legacy store even with an override installed, since it exists to migrate certificates out of it. `retain_all` makes the import one commit, idempotent by content address.

The legacy certificate providers stay: they serve operators without the override and the migration read path. Retiring them means inverting the operator/repository layering, a separate refactor.

## Tests

Migration imports the builder's `allow` delegation and proofs resolve from the branch; retains route to the branch and never touch the legacy store (enumeration stays empty); repeated installs migrate nothing (branch head unchanged); and the full repository flow (create, open, commit, select) runs with every authorize resolving through the tree walk.

### DIFF
--- a/rust/dialog-capability/src/access.rs
+++ b/rust/dialog-capability/src/access.rs
@@ -19,6 +19,7 @@
 use crate::{Ability, Attenuate, Capability, Constraint, Did, Effect};
 use dialog_common::{ConditionalSend, ConditionalSync};
 use serde::{Deserialize, Serialize};
+use std::marker::PhantomData;
 use thiserror::Error;
 
 /// Describes the scope of access being requested or granted.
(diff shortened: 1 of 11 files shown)


Gozala:

PR #439, opened 2026-08-14
feat(dialog-repository): cache resolved delegation chains

Gozala:
Final slice of [delegations-as-data](https://github.com/dialog-db/dialog-db/blob/feat/access-cache/notes/delegations-as-data.md), stacked on #438. The resolved-chain cache removes the per-effect amplification: a sync session authorizes every block get and put, and without this each one re-walks the delegation graph for an answer that cannot have changed mid-session.

## Mechanism

`SyncedAccess` caches a successful walk's chain under `(principal, subject, command)` and **re-verifies it on every hit** against the claim's full access: policy predicates run against the fresh parameters and the chain's effective window must cover the requested duration, so a hit is as sound as a walk — it skips the search, never the checks. The cache epoch is the branch head version: any commit, retain, retract, or pull moves the head and drops the cache, so staleness is impossible by construction. Failures are never cached; a claim the cached chain rejects falls through to a full walk.

## Measurements

`prove_direct` with the same N certificates in every backend (verdicts asserted per iteration):

| N | legacy fs | legacy volatile | tree walk | **tree cached** |
|---:|---:|---:|---:|---:|
| 32 | 753 µs | 32 µs | 33 µs | **769 ns** |
| 512 | 13.4 ms | 516 µs | 24 µs | **737 ns** |
| 6144 | 208 ms | 6.4 ms | 24 µs | **730 ns** |

The steady-state per-effect authorization cost is ~730 ns flat, verify-on-hit included: from 208 ms on the legacy filesystem store at the field-reported accumulation scale, through 24 µs for the uncached tree walk, to sub-microsecond once the session's first prove has run.

## Tests

A repeat claim serves from the cache with the same chain; any head movement invalidates (a retracted chain refuses rather than serving stale); and a hit outside the chain's validity window falls through and refuses.

### DIFF
--- a/rust/dialog-repository/benches/delegation_prove.rs
+++ b/rust/dialog-repository/benches/delegation_prove.rs
@@ -21,8 +21,9 @@ use dialog_capability::access::{CertificateStore, Prove, TimeRange};
 use dialog_credentials::Ed25519Signer;
 use dialog_effects::storage::{Directory, Location};
 use dialog_network::Network;
+use dialog_operator::AccessProvider as _;
 use dialog_operator::{Operator, Profile};
-use dialog_repository::{Branch, RepositoryExt as _};
+use dialog_repository::{Branch, RepositoryExt as _, SyncedAccess};
 use dialog_storage::provider::storage::{Storage, VolatileSpace};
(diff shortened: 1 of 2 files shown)


Gozala:

PR #447, opened 2026-08-16
feat(did-web): resolve did:web identities as an ambient capability

Gozala:
Layers did:web resolution on the algorithm-agnostic authorization foundation (#446, which this stacks on). did:web identities can now be resolved to a verifier, which is what lets a UCAN issued under a `did:web` DID be checked.

## Resolution is an ambient capability

Resolution is a `.perform` capability, mirroring the existing `Identify` / `Attest` ambient effects: callers do `resolve(did).perform(&env)` and the environment decides how the DID is resolved. No subject, no authorization (resolving anyone's DID to check their signature is not a granted privilege), but shaped so a subject could be added later. Because it is a `Provider`, the environment also owns policy like caching.

## New crate `dialog-did-web`

- **`Resolve { did }`** effect returning the agnostic `Verifier`.
- **`DidKeyProvider`** resolves `did:key` locally, no network.
- **`DidWebProvider`** fetches `https://<host>/.well-known/did.json` (or the `did:web:host:path` variant; percent-decoded host handles ports), parses the DID document, and builds the agnostic verifier from a verification method. Supports `publicKeyMultibase` (reusing the credential crate's `did:key` parser, so no duplicated multicodec logic) and `publicKeyJwk` (Ed25519 and P-256). Typed refusals on fetch failure, malformed document, missing or unsupported verification method; no silent fallback.
- **`MethodResolver`** routes by DID method (`key` local, `web` network) — the varsig `CompositeResolver` composes by signature type, not method, so this is the method-dispatch primitive.
- **`CachingResolver`** wraps any provider with a TTL memo and negative cache, so a did:web document is fetched once, not per verification.

## Network layer

The HTTP layer is a small `Fetch` trait (`ReqwestFetch` default, a map-backed double for tests), so the resolver is testable without the network and the client is swappable. `reqwest` is used without the `stream` feature, which compiles for `wasm32` (verified) as well as native, matching how `dialog-remote-s3` avoids the `wasm-streams` link collision.

## Tests

18 tests + a doc-test: ed25519 and P-256 via multibase, ed25519 via JWK, refusals (missing document, no verification method, unsupported key type, non-2xx), method dispatch (did:key resolves without fetching, did:web routes to the fetcher), and caching (second resolution served from cache, negative caching). All without touching the network.

## Not in this PR

Wiring the capability into the `dialog-remote-ucan-s3` authorizer's verify path. That path currently uses the varsig `Resolver` trait; adapting the `.perform` capability into it (or moving the authorizer to `perform` a `Resolve`) is a separate integration choice. did:plc is a later resolver that fits the same `Provider<Resolve>` seam.

## Security review fixes

A security review of this stack found that resolution never bound the fetched document back to the DID being resolved, and that the fetch target was attacker-influenced. Both are addressed here, each as a `fix` commit sitting on the `test` commit that pinned it:

- **Document is now bound to the DID.** The document's `id` must equal the resolved DID, and a verification method whose `controller` names a *different* DID is excluded from the key set. Without the `id` check, a document served at the victim's URL but carrying the attacker's key bound that key to the victim's DID, so the attacker's signatures verified as the victim (a full authentication bypass). Confirmed by a failing test before the fix.
- **`did:web` URL derivation validates the decoded output.** Percent-decoding ran before the URL was assembled with no check on the result, so a crafted DID chose the fetch target: `did:web:good.com%40evil.example` fetched `evil.example` while reading as `good.com`; a decoded `/`/`?` smuggled a path and query past the fixed `/.well-known/did.json`; `..` walked elsewhere; control bytes spliced in. Because the authorizer resolves the issuer DID of any submitted invocation, that was an unauthenticated SSRF primitive. Each decoded segment is now allowlisted (host: letters/digits/`-`/`.` plus one numeric port; path: unreserved only, no dot-segments).
- **The fetch client is configured, not defaulted.** `ReqwestFetch` used `Default`, which follows up to ten cross-origin redirects, has no timeout, and buffers an unbounded body. It now refuses redirects and sets a request timeout (native; the browser owns that policy on wasm32), and caps the response body on every target.


Gozala:
Stacked on #446 (algorithm-agnostic authorization). Review #446 first; this PR's diff includes #446's commits until that merges. did:plc is a planned follow-on resolver on the same `Provider<Resolve>` seam.

PR #449, opened 2026-08-16
feat(did-web): resolve did:plc identities via plc.directory

Gozala:
Stacked on the did:web PR. Adds did:plc resolution.

## Approach

A did:plc DID resolves to a standard W3C DID document (served by `plc.directory`), the same shape did:web already parses. So did:plc reuses the whole machinery: `DidPlcProvider` fetches `https://plc.directory/<did>`, parses the same `DidDocument`, and returns the same `MultiVerifier` over the document's key set. Multi-key any-match, `#fragment` selection, and skipping unsupported key types all come for free. `MethodResolver` routes `did:plc` alongside `did:key` and `did:web`, so the authorizer's default resolver stack now resolves all three.

## Trust model (deliberate)

The plc operation log is an append-only, auditable signature chain. **This resolver does not fetch or validate it** — the document `plc.directory` returns is taken as authoritative for the DID. This is a chosen trust boundary, documented on `DidPlcProvider` so it is not mistaken for an oversight. Validating the operation log (rotation-key chains, recovery windows, genesis-hash-matches-DID) is a larger, separate piece of work that can layer on later if the trust boundary needs tightening.

## Tests

Mock-fetched (no network): ed25519 and p256 did:plc resolve-and-verify; a two-key document verifies a signature by either key; an unsupported key type is skipped as long as a supported one remains; malformed / missing documents and non-plc methods are refused; method dispatch routes did:plc to the network while did:key stays local. Plus URL-derivation unit tests (base32 identifier validation).

## Security review fixes

The security review confirmed that the document-`id` binding gap from the did:web layer applies here too, and pinned a length gap specific to did:plc. The `id` fix lands on the did:web PR and inherits down; this PR adds:

- **did:plc identifier length is enforced.** `did_plc_url` checked the base32 charset but not the length, so a 1- or 25-character identifier passed and became a `plc.directory` request. The charset allowlist already contained the blast radius (no `/`, `?`, `#`, `.` or `%` can appear, so a crafted DID cannot leave the directory's path), which is why this is hardening rather than a bypass: it stops a well-formed-looking but bogus identifier from reaching the network at all. The doc comment already asserted the fixed 24-character length; the code now enforces it. Pinned by a failing test before the fix.

With the did:web `id`/`controller` binding fix inherited from the layer below, a document served for one `did:plc` but naming another is now refused rather than binding the wrong key.

### DIFF
--- a/rust/dialog-did-web/src/lib.rs
+++ b/rust/dialog-did-web/src/lib.rs
@@ -4,8 +4,9 @@
 //! [`Verifier`](dialog_credentials::Verifier). Resolution is expressed as an
 //! ambient [`Resolve`] capability performed against a
 //! [`Provider`](dialog_capability::Provider): `did:key` is parsed locally,
-//! `did:web` fetches the DID document over HTTPS, a [`MethodResolver`] routes by
-//! DID method, and a [`CachingResolver`] layers a TTL cache on top. The caller
+//! `did:web` fetches the DID document over HTTPS, `did:plc` fetches it from the
+//! PLC directory, a [`MethodResolver`] routes by DID method, and a
+//! [`CachingResolver`] layers a TTL cache on top. The caller
 //! only writes `Resolve::new(did).perform(&env)`; where the answer comes from is
 //! entirely a provider concern.
 //!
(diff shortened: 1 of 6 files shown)


Gozala:
Stack: #446 (algorithm-agnostic auth) → #447 (did:web) → **#449** (did:plc). Review bottom-up. did:plc reuses #447's document parser + MultiVerifier; the only new surface is the plc.directory fetch + method routing + the trust-model boundary (log not validated, deliberate).

PR #450, opened 2026-08-16
feat(credentials): WebAuthn passkey signing and verification

Gozala:
Stacked on the did:plc PR. Revives WebAuthn/passkey support (from the closed WebAuthn PR) and wires it through the algorithm-agnostic stack. A passkey becomes a first-class signing identity: register one with `navigator.credentials`, sign a UCAN with it in the browser, verify it server-side through the same path as ed25519 and p256.

## WebAuthn is a wrapper, not a p256 format

Per the [varsig WebAuthn spec PR](https://github.com/ChainAgnostic/varsig/pull/11), WebAuthn wraps an inner signature algorithm. `WebAuthn<Inner>` composes the inner algorithm's varsig tags with the `0x300001` WebAuthn marker (a private-use placeholder, no official multicodec yet). `WebAuthnP256 = WebAuthn<Es256>` is the concrete instance; wrapping another inner algorithm later is additive. The signature body carries `clientDataJSON` and `authenticatorData` alongside the inner ECDSA signature - variable length, which the recently variable-length agnostic signature accommodates. Verification checks the challenge is a hash of the payload and that `authenticatorData || SHA-256(clientDataJSON)` verifies under the inner key.

## A passkey is a did:key variant

No new DID-document machinery. A passkey is `did:key:z<0x300001-multicodec || 33-byte compressed P-256 point>`, which routes through the existing did:key parser to a `WebAuthnVerifier`. So did:web and did:plc documents publish and resolve passkeys as ordinary `publicKeyMultibase` verification methods, and the multi-key resolver handles them like any other key. Proven by an end-to-end did:web passkey resolution test.

## Details

- The inner signature is kept in DER form end to end (what the browser emits), distinct from the raw r||s of a plain p256 signature; the verifier parses DER.
- The agnostic verifier gains a distinct `WebAuthn` arm; header-authoritative decode routes a WebAuthn-tagged signature to it, and the `0x300001` marker disambiguates it from plain Es256.
- Behind an off-by-default `webauthn` feature. ed25519 and p256 paths are untouched and byte-identical; ed25519-only and default builds compile with the WebAuthn arms cfg'd out.
- The browser signer (`navigator.credentials`) compiles for wasm32; its live path needs a real browser, so the verify path is covered by synthetic P-256 tests as the original work intended.

## Tests

Ported the verifier and signer suites (valid / wrong-payload / wrong-key / tampered authenticator-data / tampered client-data / invalid JSON / did roundtrip / cross-key rejection / altered-challenge rejection) and added: a WebAuthn-tagged agnostic signature verifies through the agnostic verifier's WebAuthn arm, the varsig header distinguishes WebAuthn from Es256, a WebAuthn did:key is distinct from a p256-pub did:key, and a did:web passkey resolves end to end.

## Security review fixes

The security review found that `verify_webauthn` validated the challenge and the inner ECDSA signature but nothing else: `authenticatorData` was concatenated into the signed message and never inspected, and `ClientData` deserialized `challenge` alone. A valid inner signature is necessary but not sufficient, because the same key signs both WebAuthn ceremonies and the meaning lives in fields the signature does not interpret. `fix(webauthn): check the assertion` sits on a `test` commit that pinned each gap with a failing test:

- **`clientData.type` must be present and `webauthn.get`.** A registration (`navigator.credentials.create()`) signs the same `authenticatorData || SHA-256(clientDataJSON)` shape, so without this a site prompting "register a passkey" with the challenge set to the hash of a payload it chose could turn a registration gesture into an authorization over that payload. serde ignored the field's absence entirely.
- **`authenticatorData` must be at least 37 bytes** before any field is read.
- **User Present flag must be set**, so a human actually gestured. The signer's `userVerification: "required"` is a signing-time browser hint an attacker assembling an assertion never goes through.
- **`rpIdHash` is checked** against an expected relying party when the caller binds one via the new `WebAuthnVerifier::expecting_rp_id`. It is optional because a `did:key` carries a key and no origin, so a verifier recovered from a DID alone cannot know which origin to expect; a caller that does know should always bind it.
- **The inner signature is normalized to canonical low-S** before verifying, so the two encodings of one authorization collapse to one value and nothing downstream that keys on a signature can be fooled by a duplicate. Normalizing rather than refusing is deliberate: real authenticators emit high-S, and refusing would break genuine passkeys.

ed25519/p256 paths are untouched; the default build is unaffected.


Gozala:
Top of the DID/auth stack: #446 (agnostic auth) → #447 (did:web) → #449 (did:plc) → **#450** (WebAuthn passkeys). Passkeys compose with the whole stack: a passkey is a did:key variant, so it publishes/resolves through did:web and did:plc unchanged. Off-by-default `webauthn` feature.

PR #451, opened 2026-08-16
feat(credentials): RSA signing and verification (RS256 2048/4096)

Gozala:
Stacked on the WebAuthn PR. Completes the algorithm set: RSA (PKCS#1 v1.5, SHA-256, RSA-2048 and RSA-4096) is a first-class signing identity alongside ed25519, p256, and WebAuthn. An RSA key published in a did:web or did:plc document resolves and verifies through the same algorithm-agnostic, multi-key path.

## varsig algorithm

Ported from the earlier RSA work. Headers: RSA-2048 = `[0x1205, sha2-256, 0x0100]`, RSA-4096 = `[.., 0x0200]`. The two key sizes are distinct `AlgorithmTag` arms with 256- and 512-byte signature bodies, which the now variable-length agnostic signature carries (this is what the earlier variable-length change unblocked).

## Credentials + did:key

A new `dialog-credentials/rsa` module (mirroring es256) provides the signer, verifier, and resolver over the `rsa` crate. An RSA did:key is **variable-length**, unlike the fixed curve points: multicodec `0x1205` followed by the PKCS#1 DER public key (~294 bytes for RSA-2048). The did:key parser gained a variable-body arm for it. did:web and did:plc resolution then work unchanged (a passkey-style publicKeyMultibase), and the multi-key resolver handles RSA keys like any other.

**One gotcha:** a ~294-byte RSA key overflows the `base58` crate's fixed 132-byte decode buffer, so the RSA path encodes/decodes via `multibase` (byte-identical `z`-base58btc output). ed25519 and p256 keep their existing `base58` path untouched.

## Details

- Behind an off-by-default `rsa` feature. ed25519, p256, and WebAuthn are untouched and byte-identical; ed25519-only and default builds compile with the RSA arms cfg'd out.
- Compiles for wasm32.
- Tests use a cached 2048-bit key fixture (RSA keygen is slow, so no in-test generation): sign/verify roundtrip, did:key roundtrip, cross-key rejection, agnostic Rsa-arm verify, header distinguishes 2048 from 4096, and an RSA key resolving + verifying through a did:web document.

With this, the algorithm set is complete: ed25519, P-256, WebAuthn passkeys, and RSA all sign and verify through one agnostic, header-authoritative, multi-key path, and all four appear as did:key / did:web / did:plc verification methods.

## Security review fixes

The security review found a denial-of-service reachable from unauthenticated input, since the RSA arm is the only variable-length one:

- **The `did:key` length is capped before the quadratic base58 decode.** `RsaVerifier::from_str` handed the whole string to `multibase::decode` with no bound. base58 decoding is O(n^2), so the caller chose the cost: measured in release, a 100k-character `did:key` took 77ms and the quadratic growth puts 200k at seconds of CPU. Reachable because the authorizer resolves the issuer DID of any submitted invocation and `Verifier::from_did_key` falls through to this arm once the curve arms bail. The curve arms are immune only by accident of the `base58` crate's 132-byte output buffer; a `1024`-character cap is the deliberate equivalent (a 4096-bit key is ~730 characters). Pinned by a failing test; a companion test confirms a real 2048-bit key still fits under the cap.

The length-guard failure now has its own `RsaDidFromStrError::TooLong` variant, distinct from `InvalidBase58`, so a caller can tell "too long" from "bad base58" and the test asserts the variant rather than timing the call (which keeps it meaningful on wasm32, where `Instant` is unavailable).


Gozala:
Top of the stack: #446 → #447 (did:web) → #449 (did:plc) → #450 (WebAuthn) → **#451** (RSA). Completes the algorithm set (ed25519, P-256, WebAuthn, RSA). Off-by-default `rsa` feature; RSA did:key is variable-length (uses multibase to dodge the base58 crate's fixed buffer).

PR #461, opened 2026-08-23
feat(ucan): let a refusal say which question the chain failed

Gozala:
Stacked on #460 — both touch the `ContainerError` match in `authorize`, so they would conflict if opened side by side. Review #460 first; this one retargets to `main` once that lands.

`AuthorizeError` was written to mirror `CheckFailed` variant for variant — its doc comments say so, arm by arm ("Mirrors `CheckFailed::UnprovenSubject`", "Mirrors the `Expired` side of `CheckFailed::TimeBound`"). Nothing ever populated the mirror. `check_failed_to_container_error` rendered every typed reason into `ContainerError::Invocation(String)`, and the authorize boundary then mapped that whole class to `Malformed`.

The consequence is that a caller acting on these variants could not: an expired proof, a proof issued to somebody else, and a command escalation all arrived as "your bytes did not decode". A holder of a lapsed delegation was being told to fix its encoding when what it needed was a fresh proof. `notes/error-audit.md` flagged exactly this — that the client-side matching (`Expired` means refresh, `Revoked` means stop) "cannot fire for real verification failures" — and called the fix out as follow-up scope.

So the decision now travels whole. `ContainerError::Unauthorized` carries the `CheckFailed`, and the boundary names it. `TimeBoundError` grows the bound it failed against and the instant it was judged at, which `Expired { expiration, at }` and `NotValidBefore { not_before, at }` need and previously had no way to obtain — the old variants were unit-like and knew only *that* a bound failed.

Two cases stay descriptive on purpose: a window no instant satisfies is a defect in the chain rather than a clock verdict (no fresh proof at any time would help), and a pending promise is not an access decision.

Three tests pin the variants that were unreachable, each asserting the refusal names its particulars — the lapsed bound, the two principals by identity, both commands. All three fail if the mapping goes back to stringly `Malformed`.

### DIFF
--- a/rust/dialog-remote-ucan-s3/src/authorizer.rs
+++ b/rust/dialog-remote-ucan-s3/src/authorizer.rs
@@ -60,7 +60,9 @@ use dialog_capability::{Capability, Constraint, Did, Policy};
 use dialog_did_web::{CachingResolver, PerformingResolver, Resolve, WebResolver};
 use dialog_effects::{archive, blob, memory};
 use dialog_remote_s3::{Address, Permit, S3Credential, S3Error};
+use dialog_ucan_core::invocation::CheckFailed;
 use dialog_ucan_core::promise::Promised;
+use dialog_ucan_core::time::TimeBoundError;
 use dialog_ucan_core::{Environment, InvocationChain, VerificationContext};
 use ipld_core::ipld::Ipld;
 use serde::de::DeserializeOwned;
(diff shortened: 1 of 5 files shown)


Gozala:
Consumed this downstream in tonk, which turned out to be the useful validation of it.

Tonk's access service carried a hand-rolled screen that re-parsed every presented container to compute its validity window, purely so an expired chain could be answered `401 Expired` instead of the `400 Malformed` the boundary produced. With this change the walk names the refusal itself, and that screen — ~100 lines re-implementing the intersection dialog already computes — deletes outright, with an integration test asserting the status and typed body a client sees.

Also worth reporting: the same flattening existed a second time in tonk's own native server helper, which rendered every authorization failure into `format!("Authorization failed: {e}")`. So the typed reason was being destroyed twice on the way out. Both fixed now, but it's evidence the funnel was costing more than the one hop — `notes/error-audit.md` predicted exactly this ("every denial through this server degrades to `Rejection::Unclassified`").

PR #464, opened 2026-08-24
Present authority good at the moment it presents, and retire lapsed memos

Gozala:
Two clock bugs on the authority path, found from tonk's side while chasing a spot whose sync broke on the second join.

The proof walk has no clock of its own: it admits a certificate whose window covers the REQUESTED one, and an unbounded request is covered by every window, including one that closed yesterday. While one route to a subject exists this is invisible; the moment two are retained side by side, candidate order follows content hashes, so a lapsed certificate is picked about half the time and presented to a responder that does check the clock — which refuses authority the holder still has.

The presign path now asks for a chain good at the instant of presentation. `covers` is inclusive, so this admits exactly what a responder checking `now` admits, and does not bound the resulting authorization: the proof still reports the window its certificates agree on.

The memo behind the walk had the same blind spot one level up: hit-time re-verification judges the cached chain against the claim's requirement and nothing else, and read paths require nothing in particular, so an entry that lapsed mid-epoch was replayed for every claim on its key until the branch head next moved. A lapsed memo now retires, sending the claim back to the walk, which can still find a live route the memo has outlived. Only the lapsed end is checked — an entry that has not started yet may legitimately answer a claim for the window in which it does.

Both carry tests: one records the window the fork actually requests, the other retains a lapsed grant beside a live one and asserts the live route resolves.

### DIFF
--- a/rust/dialog-operator/src/operator/access.rs
+++ b/rust/dialog-operator/src/operator/access.rs
@@ -38,6 +38,7 @@ use dialog_capability::access::{
     Retain, Scope as _, TimeRange,
 };
 use dialog_capability::{Capability, Command, Did, Fork, Policy as _, Provider, Subject};
+use dialog_common::time::{self, UNIX_EPOCH};
 use dialog_common::{ConditionalSend, ConditionalSync};
 use dialog_effects::archive::{Get, Import, Put};
 use dialog_effects::authority::{Attest, Identify};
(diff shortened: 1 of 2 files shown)


Gozala:
Duplicate of #462 — same two commits; that one also carries a third. Closing in its favor.

PR #467, opened 2026-08-25
fix(operator): a proof never waits on the fetch it is proving

Gozala:
`tonk account devices` was parking for good after another device pushed to the account, with no I/O and nothing dropped. Traced with an in-process step recorder, the sequence is:

1. A bare pull adopts the other device's head by root, so the access branch head references nodes the local archive does not hold.
2. The next proof walks the access branch, hits such a node, and fetches it through the walk's reach.
3. That fetch is itself proven by the reach-less inner operator, whose walk over the same branch needs the same node, finds the outer fetch's single-flight claim in the shared node cache, and waits on it. The outer fetch is waiting on the inner proof.

The inner operator was meant to bound this recursion by resolving from what is already local, but "local" read through the shared cache includes waiting on someone else's in-flight remote fetch. It now reads through caches of its own (`Branch::with_own_caches`): the head, upstream, and overlay stay shared, only what it caches is its own. Its walk then resolves from local content or fails, and the fetch it proves proceeds or errors, within a bound either way.

The integration test resets a device's access head to the account's pushed head (the revision, none of its nodes) and requires the next pull to complete within 30s. It fails after 30s on `main` with exactly the recorded trail, and completes in milliseconds with the change.

Verified end to end on tonk's deterministic repro: three `devices` rounds went from 40.00s each to under 10ms.

### DIFF
--- a/rust/dialog-operator/src/operator/builder.rs
+++ b/rust/dialog-operator/src/operator/builder.rs
@@ -144,6 +144,11 @@ impl OperatorBuilder {
             .perform(&operator)
             .await
             .map_err(|e| OperatorError::Delegation(format!("{e}")))?;
+        // The reach-less operator below reads the access branch through
+        // caches of its own: its walk proves the fetches this handle makes,
+        // so sharing a cache would have that proof wait on the fetch it is
+        // proving whenever both need the same node.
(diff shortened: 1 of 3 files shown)


Gozala:
Fixed by #468

PR #477, opened 2026-08-31
fix(wbg-pool): harden pooled browser lifecycle

jackddouglas:
## Summary

- Keep an existing daemon/browser alive when a loaded health probe briefly exceeds 500 ms instead of spawning a replacement and orphaning Chrome.
- Close daemon HTTP/1 responses so tabs do not leave keep-alive sockets behind until the process reaches its descriptor limit.
- Capture the browser harness's reporting `fetch` before test code can replace `window.fetch`.
- Add focused regressions for all three lifecycle contracts.

## Why

Tonk's 1,438-test nextest suite exposed three independent failures while adopting `wbg-pool` as its default Wasm runner:

1. Concurrent tab startup could delay `/api/health` beyond 500 ms. The shim acquired the spawn lock, replaced a daemon whose PID was still alive, and orphaned its browser.
2. Closing a test tab did not close its HTTP/1 keep-alive sockets. On macOS, an unchanged 164-test analyzer archive repeatedly failed after test 123 with the daemon at 254 descriptors against a 256-descriptor soft limit. Sending `Connection: close` reduced the post-run count to 18 and the same archive passed.
3. Application tests that replaced `window.fetch` also replaced the harness control channel, preventing the completed test result from reaching the daemon.

The daemon selection now treats an unhealthy live PID as an existing daemon and gives it the normal readiness window. The server closes browser connections after every response, and the browser harness reports through an early bound fetch reference.

## Verification

- `cargo test -p wbg-pool`: 9 passed
- `cargo clippy -p wbg-pool --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`
- Each new regression was observed failing before its corresponding fix.
- Downstream Tonk pooled CI passed the same 1,438-test inventory in both debug and release with zero retries: [Actions run](https://github.com/tonk-labs/tonk/actions/runs/33366158099)
- Tonk's canonical, unsuffixed pooled commands also passed both complete web jobs after adoption: [Actions run](https://github.com/tonk-labs/tonk/actions/runs/33372091102)

### DIFF
--- a/rust/wbg-pool/src/daemon/harness.rs
+++ b/rust/wbg-pool/src/daemon/harness.rs
@@ -125,3 +125,34 @@ pub fn generate(config: &RunConfig<'_>) -> Result<RunScripts> {
     };
     Ok(scripts)
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn browser_reporting_keeps_fetch_from_before_test_code_runs() {
+        let scripts = generate(&RunConfig {
+            mode: TestMode::Browser,
+            module_js: "/module.js",
+            module_wasm: "/module.wasm",
+            exports: &["__wbgt_test"],
+            include_ignored: false,
+            filtered_count: 0,
+            nocapture: false,
+        })
+        .unwrap();
+
+        let capture = scripts
+            .run_js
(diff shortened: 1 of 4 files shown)


Gozala (approved):
Thanks

PR #478, opened 2026-09-02
dialog-artifacts: gate the wasm-bindgen surface behind a default-on `web` feature

frolic:
A crate that depends on `dialog-artifacts` and targets wasm32 gets every `#[wasm_bindgen]` export in `web.rs`, and the binding on `ValueDataType`, compiled in. If that crate defines its own wasm-bindgen exports, the names collide and the generated `.d.ts` includes both sets.

This PR adds a `web` feature, on by default, and gates `pub mod web` and the `wasm_bindgen` attribute on `ValueDataType` behind it. Default builds, the nix packaging, and the web tests are unchanged. With the feature off, the crate compiles for wasm32 with no exports.

This replaces the `openWith` change this PR carried before. Since the JavaScript bindings may be dropped, a downstream crate can bind its own surface over `Artifacts::open` in Rust instead; the feature gate is the one change that needs. Start at `rust/dialog-artifacts/src/lib.rs`.

### DIFF
--- a/rust/dialog-artifacts/Cargo.toml
+++ b/rust/dialog-artifacts/Cargo.toml
@@ -9,7 +9,10 @@ license.workspace = true
 crate-type = ["cdylib", "rlib"]
 
 [features]
-default = ["csv"]
+default = ["csv", "web"]
+# The wasm-bindgen surface in `web.rs`. Off, the crate compiles for wasm32
+# without exporting anything, so a dependant can bind its own surface.
+web = []
 debug = []
 helpers = ["dep:anyhow"]
 csv = ["dep:csv-async"]
--- a/rust/dialog-artifacts/src/artifacts/value.rs
+++ b/rust/dialog-artifacts/src/artifacts/value.rs
@@ -826,7 +826,7 @@ impl From<Value> for ValueDataType {
 /// [`ValueDataType`] embodies all types that are able to be represented
 /// as a [`Value`].
 #[cfg_attr(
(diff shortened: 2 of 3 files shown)


Gozala:
@frolic I do not mind these change in principal, but I did mention in the other pr that there is very little appetite for taking on maintenance of JS bindings. In fact we were considering dropping the ones we already have.

It is also worth a note that dialog has moved on from `StorageBackend` based interface [to capability based one](https://github.com/dialog-db/dialog-db/tree/main/rust/dialog-storage/src/storage/provider) and we actually have OPFS backend also.  I think only reason `StorageBackend` still lingers is because we have not moved search tree fully onto capability based API and we use an trait based adapter as an interim solution.

Separately it is also worth noting that reason dialog uses IDB / OPFS as backends is because it's designed to partially replicate and cache blocks it did. Network based stores are supported through remotes instead where you effectively need to provide capability to read / write blocks and to do atomic swaps. `dialog-remote-*` crates define providers of those capabilities. Even if you do not wish to persist blocks across sessions, I'd still recommend using volatile store (in memory store) paired with custom remote implementation, because query engine will frequently end up refetching same nodes otherwise making setup impractical.

frolic:
I have replaced the `openWith` change with a smaller one: a default-on `web` feature that gates the wasm-bindgen surface. A downstream crate can then bind its own surface over `Artifacts::open` without colliding with these exports. Default builds are unchanged.

PR #479, opened 2026-09-02
Expose conceal and reveal to JavaScript

frolic:
`dialog-credentials` can conceal bytes to an Ed25519 `did:key` and reveal them with the matching signer, but only Rust callers can reach it. This adds `conceal(recipientDid, context, plain)` and a `Signer` class (`generate`, `import`, `export`, `did`, `reveal(context, sealed)`) to the web bindings in `dialog-artifacts`, the crate wasm-pack publishes.

Why: a sync layer that moves a user's data through a relay it does not control can conceal a payload for the recipient's DID in JavaScript and let the relay carry bytes it cannot read.

Key material crosses the boundary the way `KeyExport` already does: a 32-byte seed, or the opaque `CryptoKey` from `Signer.export`. Context labels arrive as owned strings while `Context` takes a static one, so each distinct label is interned once.

Verified with a new wasm test that conceals for a generated DID and reveals with its signer, run in headless Chrome with the crate's wasm suite (172 passing), plus `cargo check --target wasm32-unknown-unknown`, `cargo test -p dialog-artifacts`, and `cargo fmt`.

### DIFF
--- a/rust/dialog-artifacts/Cargo.toml
+++ b/rust/dialog-artifacts/Cargo.toml
@@ -60,6 +60,7 @@ anyhow = { workspace = true }
 tokio = { workspace = true, features = ["sync", "macros", "io-util", "fs"] }
 
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
+dialog-credentials = { workspace = true }
 getrandom = { workspace = true, features = ["js"] }
 wasm-bindgen = { workspace = true }
 wasm-bindgen-futures = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1051,6 +1051,7 @@ dependencies = [
  "csv-async",
  "dialog-capability",
  "dialog-common",
+ "dialog-credentials",
  "dialog-effects",
  "dialog-search-tree",
  "dialog-storage",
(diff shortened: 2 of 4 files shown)


rust/dialog-artifacts/Cargo.toml
```diff
@@ -60,6 +60,7 @@ anyhow = { workspace = true }
 tokio = { workspace = true, features = ["sync", "macros", "io-util", "fs"] }
 
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
+dialog-credentials = { workspace = true }
```
Gozala, line 63: I am afraid this violates layering of the dependencies. I really don't want artifacts to depend on credentials. In fact artifacts is meant to further shrink in it's scope, just have not got around to doing it yet. It's just meant to be just a thin wrapper around search-tree mapping associative facts onto an search index. Making it aware of credentials would not be ideal.

Gozala:
Hi @frolic thank you for the contribution! I am afraid artifacts is not the right place for exposing cryptographic primitives. However it is true that it's only crate that exposes web bindings which is for historical reasons. Truth is, web bindings have not being used in practice and there is little appetite for their maintenance. If you are interested in taking on that effort, I'd love to discuss the plan for doing so.

rust/dialog-artifacts/Cargo.toml
frolic: Makes sense! This was a purely Claude driven PR, so apologies if it missed this 🙈

frolic:
Closing. You are right that this does not belong in `dialog-credentials`. For web bindings I will follow #478 and bind my own surface in Rust.

PR #480, opened 2026-09-02
Expose the blob index to the wasm bindings

frolic:
The `Artifacts` wasm class had no way to reach the blob index, so a JavaScript caller could see facts but not the blobs a tree references. This adds `putBlob` (record a reference and its size, returns the new revision), `getBlob` (read the record for a reference), and `blobChanges` (iterate the references added and removed between two revisions). They wrap three same-named methods added to `Artifacts` in Rust, and follow the existing `commit`, `select`, and `revision` bindings in how bytes, revisions, and errors cross the boundary.

Why: a backup tool running in a browser can take the revision it last archived, ask which blob references were added since, and upload only those, rather than scanning every fact. Blob bytes are outside this surface: the index carries a reference and a size, and the bytes stay wherever the caller keeps them.

Verified with `cargo test -p dialog-artifacts` (205 tests, including a new one that records a reference, reads it back, and checks the differential names that blob), `cargo check --target wasm32-unknown-unknown`, and a test in `dialog-artifacts-web-tests` that passes in headless Chrome against the built package.

### DIFF
--- a/rust/dialog-artifacts/src/artifacts.rs
+++ b/rust/dialog-artifacts/src/artifacts.rs
@@ -77,7 +77,7 @@ use crate::tree::{
     spill_cache,
 };
 use crate::{
-    DialogArtifactsError, HASH_SIZE, Key, State,
+    BlobChange, BlobIndexExt as _, BlobRecord, DialogArtifactsError, HASH_SIZE, Key, State,
     artifacts::selector::Constrained,
     buffered::{BufferedBatch, SpineSlot, apply_buffered_reusing},
     make_reference,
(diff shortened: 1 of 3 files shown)


Gozala:
Hey @frolic thanks for submitting a pull request, overall this looks clean and simple change. That said I would like to carefully consider implications of what is exposed before accepting this. My intention had being that you'd be able to add blobs through the same transaction assert API as any other concept or fact namely `branch.transaction().assert(blob).commit...` where blob potentially could be web Blob or rather something you can `try_from` it.

As of diffing I would personally prefer to avoid extending surface there also and if needs be add modifier to the `differential` method like `.ony_blobs()`.

My other hesitation is that I am considering to decommission blobs all together. They were added before dialog supported arbitrary sized value that after configured threshold spill into own block. I have not had enough time to consider all the tradeoffs yet, but on surface it seems like a good idea and it shrinks API surface.

If you have a use case for this API under time pressure,  we can discuss this in higher bandwidth channel e.g. I was considering to start community calls to do this kind of decisions.

frolic:
Arbitrary blobs as values would actually simplify a lot on my end! Will close this out in favor of waiting on that :)

Issue #492, opened 2026-09-09
Paralelize query driven replication


Gozala:
Query engine executes sequentially, allowing it to reduce search space, which means on a new replica all the blocks will be fetched sequentially one at a time which makes first load of tonk application very slow.

I think we can improve a lot if query engine could preload blocks ahead of time even if query engine will end up not utilizing them. For example concepts are effectively joins on the same entity and running each attribute scan concurrently would be far more efficient than doing one at a time. It is also worth a note that concepts are most common way to query.

Rules that can have more complicated joins preventing concurrent scans could choose a first scan based on available information while also kicking off preload for joined scans ideally with lower priority that actual loads.

I think this could make number of roundtrips more proportional of tree depth and number of ranges being visited.

PR #502, opened 2026-09-10
feat: the preload queue becomes ambient env state, hints default-on

Gozala:
Stacked on #501. Stage 2 of the v3 replication design (#492, bead dialog-db-82): the preload queue follows the hydration flight into the env, and the per-query staging API retires.

## What changed

- **Vocabulary** (dialog-artifacts, beside `Preload`): `FetchBudget` keeps its shape and gains `ZERO` as the off switch; `PreloadQueue` is the env's ambient queue of speculative preloads (two bounded rank queues plus the budget, pure data, drop-oldest past 1024 pending per rank — hints describe what a running evaluation is about to read, so the freshest are the ones worth keeping); the new `Speculation` command hands out the queue's shared handle. Shared-leaf placement for the same dev-cycle reason as `Hydrate` in #501.
- **The operator provides both halves.** `Provider<Preload>` enqueues into its own queue, refusing without enqueueing at zero budget — which is what keeps evaluators' stop-on-first-refusal and the soak's deterministic demand gate working. `Provider<Speculation>` hands the queue to drivers. The budget is runtime-adjustable through the handle; the soak's unshaped profile sets `ZERO` right after building the operator.
- **`QueryEnv` forwards `Preload` unconditionally** to the underlying env. The plan field, `with_plan`, `SelectQuery::preload`, public `FetchPlan`, and `PreloadHandle` are gone; nothing in the engine used abort/promote, and if M6's speculative ranges want reprioritization it re-enters at the queue.
- **Every evaluation drives.** `Driven` pops from the ambient queue (budget read once per stream, still a per-driver cap) and now wraps every evaluation path: session queries always, both subscription evaluation forms (full recompute and fixpoint continuation), and mid-transaction queries. One evaluation's polling executes another's hints — cross-query and cross-subscription sharing by construction, no per-path wiring.
- **Explicit warming needs no dedicated API**: performing `Preload` against the env enqueues from anywhere, and whichever evaluation polls next replicates it. A fetch.rs test pins exactly this shape (a hint enqueued outside any query, executed by a query that never asked for it), and a second pins the zero-budget off switch.
- The soak gains an `overlap` phase: two independent concepts (the full board and a compact listing sharing three of its five attribute ranges) run concurrently on one fresh cold client.

## Numbers (broadband soak, cold client, 4k entities)

| phase | before (#501) | after |
|---|---|---|
| subscribe (standing query, cold first poll) | 1.03s / 12.9 rounds | **497ms / 6.2 rounds** |
| overlap (two concurrent overlapping concepts) | — | **497ms / 6.2 rounds / 70 blocks, zero dups** |
| concept (plain query) | 497ms / 6.2 rounds | 497ms / 6.2 rounds |
| filtered (selective fold) | 1.56s / 20 rounds | 1.56s / 20 rounds |

The subscription's cold first poll — the path every UI drives, which the per-query staging never reached — now lands at full parity with a plain query. The overlap pair costs the same as one query: the union footprint hydrates once through the env-owned flight and queue, with zero duplicate fetches. The unshaped profile pins the engine's exact demand shape with `FetchBudget::ZERO`, unchanged from the stage-1 baseline.

Design record: `notes/fetch-scheduler.md` ("Stage 2 landed").


Gozala:
A correction to the numbers table: the filtered row (1.56s / 20 rounds) was bimodal at this commit, and the concept/subscribe/overlap rows wobbled with it. Some runs planned the filtered concept (and, on the next branch, rule bodies) as a merge instead of a fold and landed at 8.4s / 105 rounds. Root cause was not in this PR: `range_estimate` read the tree root alone, and near a height boundary the root has few children, so every range fell inside one child and estimated as that child's whole-subtree scale — every scan compared equal and the merge-versus-fold guard called everything balanced. The tree shape (and so the mode) varied per run because the history region's keys derive from signed revisions, which embed the per-run keypair and timestamps.

Fixed in the stacked branch (feat/rule-warm, 117cb33c): the estimate descends to the first level where the range spans siblings. Six consecutive soak runs hold the numbers in this description. Follow-ups filed as beads dialog-db-83 (the value-pinned merge runs near-serial even with range hints) and dialog-db-84 (byte-deterministic soak vault).

PR #505, opened 2026-09-11
Fix the serial block loading: integrate warms a window of lookup paths

Gozala:
The app loads serially: block GETs go out one at a time, so a sign-in that should take a few round trips takes a minute. A throttled HAR shows it plainly, 20 block GETs at peak 1 in flight, while the push of the same tree reaches peak 15.

## Why it took so long to reproduce

Each attempt measured the wrong thing:

- counting block reads instead of round trips (local reads overlap for free, so the number reads healthy while every fetch is serial)
- a replica sharing the source's storage, so nothing crossed the wire
- a replica with *nothing* local, so the merge had no base to diff and the download found the tree already there
- measuring the push alongside the pull, so the push's fan-out hid the pull's serial run

What finally worked was giving up on synthetic fixtures. This PR adds a CARv1 codec to dialog and a profile export route to tonk, captures a real account and a real first-load device from a running app, and replays the exact sign-in: a seeded device adds the account as a remote and pulls.

## Correction: this is not wasm-only

An earlier revision of this PR said the browser serializes because on wasm the request *is* the polling, so a future nobody polls never calls `fetch()`. That was wrong, and so were the numbers it rested on: they compared `peak` on native against `serial_run` on wasm, which are different measurements.

Native serializes identically, 29 to 32 on both targets, same fixtures, same test. There is no wasm/native concurrency difference here, and `spawn` was never the answer.

## What it actually is

Five fixes failed before this because each picked a plausible loop and changed it without proving that loop issued the serial fetches. So instead of guessing again I made the code name itself: a thread-local read-site label that callers set before awaiting reads, printed by the s3 socket probe next to each request boundary.

```
[s3 start #0] 5cUKpPRhzkcD @integrate/get
[s3 end   #0] 5cUKpPRhzkcD @integrate/get
[s3 start #1] D3LdrnknkNLd @integrate/get
[s3 end   #1] D3LdrnknkNLd @integrate/get
```

27 of ~30 serial fetches under `integrate/get`, while the download's `traverse` overlapped 6 to 16 at a time in the same run. The traversal was never the problem; the merge ahead of it was.

It is `TransientTree::integrate`: one point lookup per change, awaited before the next change is even pulled off the stream. On a cold partial replica a lookup costs one round trip per level, with nothing else in flight.

Worth noting I had ruled `integrate` out earlier because it processed 3,680 changes against only ~90 fetches. That ratio says most changes hit warm nodes. It says nothing about whether the misses are concurrent, which was the actual question.

## The fix

Ordering is load-bearing: each change resolves against the effect of the ones before it, so changes cannot be reordered or applied concurrently. But the `get` is a probe, and a probe's cost is fetching blocks. So each pass takes a window of 16 changes, warms every key's path through the stored frontier concurrently, then applies the window in its original order. The applying lookups read blocks the window already pulled.

Warming is advisory: it resolves nothing, swallows absence and failure, and a key it misses is just read the slow way. The result is byte-identical to applying the stream one change at a time.

```
longest serial run   32 -> 3
wire fetches         65 -> 58
```

Fetches going down was a surprise. Real change streams cluster, so warming a window finds shared paths the one-at-a-time walk was re-reading. The synthetic test, where keys are deliberately spread so no two share a path, shows the opposite trade (27 reads serial, 92 warmed): one round trip per level per window instead of one per change. That is the trade worth making, and it is why the test asserts on overlap rather than on a read count.

## What's here

- `TransientTree::integrate` takes a window and warms its lookup paths (`warm_paths` + `boundary_of`)
- `it_integrates_a_differential_concurrently`, measuring peak reads in flight through `ObservingBackend`: 1 before, 16 after
- CARv1 snapshot codec (`repository::snapshot::codec`), and a profile export route in tonk
- real captured fixtures, zstd-compressed (56K + 50K)
- `it_joins_an_account_from_a_seeded_device`, running on both targets, bound at `serial_run < 10`: a few solo reads are inherent (the first fetch has nothing to overlap with) and the floor drifts 3 to 5, while the bug it pins was 29 to 32
- a gauge counting the longest run of fetches with nothing else in flight, which a peak cannot see

Both tests fail without the warming. 2720 tests pass with it.


Gozala:
Closing in favor of #504, which is the same head stacked on #503. This PR's description described the windowed warming that was reverted; #504's body carries the landed design.

PR #520, opened 2026-09-19
fix(operator): derive operator keys by key agreement, not by signing

Gozala:
## Why

An operator key was derived by signing a fixed input and hashing the
signature. That only works where signatures are deterministic, and in
Safari they are not: Apple's CryptoKit hedges Ed25519 nonces, so the same
profile derived a **different operator on every call**. The same profile
also derived different operators on native and in the browser, because the
two arms hashed different material — the extractable arm hashed the seed,
the browser arm hashed a signature.

An operator DID has to be stable: it is the audience of every delegation a
profile mints, so a moving one invalidates the session it just issued.

## Approach

Derive by **key agreement with itself** rather than by signing. X25519
agreement is a pure function of the key, so it is deterministic on every
platform and inside every keystore — including Secure Enclave and TPM
keys, whose ECDSA randomizes `k` and which a signature-based scheme could
therefore never have reached.

The derivation is pinned from both ends: a known-answer vector fixes the
derived secret, and a second test fixes the `did:key` that secret becomes,
asserted by the same test on native and wasm so the two arms cannot drift
apart again.

WebKit is covered directly, since CI otherwise runs wasm tests in Chromium
only. It reaches the same values by a different archive path: WebKit
cannot structured-clone an X25519 `CryptoKey`, so the agreement key must be
archived AES-KW wrapped, and the tests force that path rather than relying
on Chromium's ability to clone. A profile archived before agreement keys
existed fails loudly with `AgreementKeyUnavailable` instead of deriving
something wrong.

## The API this lands with

Derivation returns a **signer**, not key material, so the derived seed is
never a value a caller holds:

```rust
let operator = signer.secret(OPERATOR_DERIVATION_CONTEXT)
    .derive(b"my-app")
    .await?;
```

`Ed25519Signer<E = Sealed>` records whether material can be read back, and
only `Ed25519Signer<Extractable>` has an `export`. A consumer that needs
raw bytes asks for that type in its own signature, so a sealed key cannot
reach it — `derive_extractable` is how one is made, for `iroh::SecretKey`,
which is built from bytes and whose QUIC stack needs the material in
process.

The default carries the cost: every use outside this crate is a `use` or a
constructor call, so all twelve dependent crates compile untouched.

Writing that split `export` in two. Archiving a key and revealing its
material were one method: in the browser a sealed key archives as opaque
`CryptoKey` handles that carry nothing and restore still sealed, which is
how a profile is stored between sessions. `archive` is available in every
state; `export` is not.


rust/dialog-credentials/src/secret.rs
```diff
@@ -123,6 +123,78 @@ impl Secret<'_> {
...
+    /// kind of key decided by the type rather than by the method name. An
+    /// [`Ed25519Signer<Extractable>`] comes back readable; the seed reaches it
+    /// through [`ExtractableKey::import`](crate::key::ExtractableKey::import),
+    /// which asks `WebCrypto` for an extractable `CryptoKey` where the default
+    /// import asks for one that will not give its seed back.
+    ///
+    /// # Errors
+    ///
+    /// As [`Self::derive`].
+    pub async fn derive_as<K>(&self, label: &[u8]) -> Result<K, SecretError>
```
Gozala: Can instead derive have <K=Sealed> that you can override with turbofish ?

Gozala: Or alternatively it could be a trait that Secret implements that has  derive method which returns Ed25519Signer<Extractable> instead. In both cases getting extractable is a deliberate and non-default behavior.

Gozala: Not possible — Rust rejects a default on a method type parameter:

```
error: defaults for generic parameters are not allowed here
  = warning: this was previously accepted by the compiler but is being
    phased out; it will become a hard error in a future release!
```

Defaults are allowed on types, not on method generics. And where the old form still parses it does not drive inference, so a plain `s.derive(x)` fails with `E0283: type annotations needed` — every existing call site would need a turbofish.

Went with your second suggestion instead.

Gozala: Done in 768071f2 — `derive_as` is gone, replaced by `SecretExtractableDerive`.

One thing worth knowing: an inherent method always wins method resolution over a trait method, even with the binding annotated. So `secret.derive(label)` is *always* the sealed one, and the extractable path only exists in fully-qualified form:

```rust
let peer = SecretExtractableDerive::derive(&signer.secret(CTX), b"peer").await?;
```

Which lands where you wanted it: a readable key needs both the trait import and the explicit call.

Gozala: Correcting my framing above — I called the inherent-method shadowing a caveat, which gets it backwards. It is the mechanism, and it is what you were asking for: `.derive()` yields a sealed key whatever the call site expects, so no binding's type can quietly turn one readable. Naming the trait is the deliberate act.

Docs reworded to say that in 8213f48d.

PR #525, opened 2026-09-21
feat(credentials): a signer's extractability is in its type

Gozala:
Stacked on #520.

## Why

`derive` returning credentials closed the incidental leak, but `export`
was still there for the asking: any signer could be asked for its key
material, and the answer was a runtime branch on what the key happened to
be. Whether that succeeds should be visible in the type, not discovered
at the call.

## Approach

`Ed25519Signer<E = Sealed>`, where only `Ed25519Signer<Extractable>` has
an `export`. Reaching for raw bytes is a name that does not resolve rather
than a call that might refuse, and a consumer which genuinely needs them
(`iroh::SecretKey` is built from bytes and its QUIC stack offers no
external-signer hook) declares that in its own signature.

The default carries the cost. Every use of `Ed25519Signer` outside this
crate is a `use` or a constructor call, so all twelve dependent crates
compile untouched; `Signer::Ed25519` holds the sealed form, since a
credential in general circulation should not be exportable.

Sealed is a promise about the API, not always about the platform. In the
browser it is also physical, since a non-extractable `CryptoKey` has no
seed to give. On native every key is readable by `ed25519_dalek`, so what
it buys there is that nothing reads one by accident.

## What writing it found

`export` was two operations under one name. Archiving a key and revealing
its material are different things: in the browser a sealed key archives as
opaque `CryptoKey` handles that carry no material and restore still
sealed, which is how a profile is stored between sessions. That path would
have been broken by requiring `Extractable`, and the typestate is what
surfaced it. `archive` is now available in every state; `export` is not.


Gozala:
Folded into #520. Keeping them apart meant #520 introduced a public `derive_bytes` that this PR immediately made private, so reviewing #520 alone showed an API we had already decided against.

PR #527, opened 2026-09-21
refactor(peer): split Operator into Peer and Session, rename crate to dialog-peer

Gozala:
## Summary

Splits the one `Operator` type into two, with no compatibility alias, and drops `Profile`:

- **`Peer`**: a stored, persistent identity over a `Storage`. Built with `Peer::new().storage(storage).network(..).base(..).branch(..).open(location)`, or `load` and `create` for the failing variants. The peer is itself an env, so `perform(&peer)` is the unconstrained case. What `Profile` fronted sits on the peer: `did`, `credential`, `access`, `secrets` for the site credentials, `space` for named spaces.
- **`Session`**: a scoped delegation from a peer over a `SignerCredential`. `peer.derive(ctx).await?` returns the builder under a key derived for that context, and `peer.session(credential)` starts one over a supplied key. Built with `peer.derive(ctx).await?.allow(..).grant(cert).build()`, where `allow` takes a capability for an unbounded grant or a `Claim` from `peer.access()` that carries its own window.

The crate `dialog-operator` becomes `dialog-peer`. `Operator`, `OperatorBuilder`, `DeriveOperator` and `Profile` are gone. `dialog-identity` keeps the credential loader as `OpenCredential`, yielding a `SignerCredential`, plus the access API, the site-secret handle and `SpaceHandle`; its error is `IdentityError`. The authority attenuation and the schema facts that also carry the name `Profile` are different things and stay.

`access_branch` from #526 lands here as `PeerBuilder::branch`, with the "main holds nothing" test ported, so that PR can close.

Sessions derive by the key agreement #520 introduced, under the same `v2` label, and the pinned-DID fixture asserts the DID #520 pins, so a session derived from a context is the operator that context derived before the rename.

The design, what each piece becomes, the open decisions, the ordering against #524, #526 and #519, and a migration guide with a before/after table live in `notes/peer-and-session.md`.

## Verification

- Tests pass in every migrated crate (`dialog-identity`, `dialog-peer`, `dialog-repository`, `dialog-remote-fs`, `dialog-baseline`, `dialog-soak`, `dialog-query`, `dialog-credentials`), zero failures, on the branch with `main` merged after #520.
- Three doctests in `dialog-repository` fail on `crate::Hydrate`. They fail identically on `main` and #524 records them as pre-existing.
- Clippy passes for the migrated crates at `-D warnings --all-targets --all-features`. Four lints from the local clippy being newer than the repo's last lint, all in files this change does not touch, were allowed locally: `result_large_err` (`dialog-ucan-core`), `unnecessary_sort_by` (`dialog-remote-fs`), `chunks_exact_to_as_chunks` (`dialog-search-tree`), `while_let_loop` (`dialog-query`). CI's pinned toolchain does not raise them.
- The wasm32 check of `dialog-peer` and `dialog-identity` passes.
- Not run locally: the feature-gated S3 and access-service integration tests, which need provisioned servers. They compile under `--all-features` and CI runs them.
- tonk-labs/tonk#990 ports tonk to this branch and is the consumer-side check.

## Follow-ups

- The mechanical migration edits were densest in `dialog-repository`'s tests. Two regex passes touched unrelated files and were reverted after the gate caught them. Worth a skim there.
- #520's seven commits remain in this branch's history because #520 landed as a squash. They add nothing to the diff. Dropping them from the commit list takes a rebase onto `main`.
- Folding `Session` into `Peer` as one struct, with `did()` for the acting key and a second accessor for the replica identity, is the next shape question; the note records it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SyWN8CCDeBinXiRVt8pBYo


notes/peer-and-session.md
```diff
@@ -0,0 +1,359 @@
...
+Everything found by convention reduces to one rule: the peer's credential
+and its own repository share one space at one `Location`. Everything else
+is located by facts in that space.
+
+## Surface
+
+Implemented today (`dialog-peer`):
+
+```rust
+let alice = Peer::new(disk)                // Storage<S>; volatile for tests
```
Gozala: I think it should be .storage(disk) as opposed to new taking disk

notes/peer-and-session.md
```diff
@@ -0,0 +1,359 @@
...
+    .network(net)                          // optional, Network::default()
+    .base(Directory::Current)              // where space names resolve, until the registry
```
Gozala: This probably should be optional ? Or maybe individual spaces are given locations on opening so no base would be required

notes/peer-and-session.md
```diff
@@ -0,0 +1,359 @@
...
+    .branch("main")                        // registry + default proof source
+    .open(Location::profile("alice"))      // or .load(..), or .attach(credential)
+    .await?;
+
+let job = alice.session(b"refactor")       // derived key, deterministic per context
+    .credential(signer)                    // or a supplied SignerCredential
+    .allow(scope)                          // grants minted at build
+    .allow_until(scope, t)                 // bounded
```
Gozala: I don't like `.allow_until` and I would expect `scope` to capture it instead. Perhaps better option would be something like

```rs
let job = alice.session(b"refactor")       // derived key, deterministic per context
    .credential(signer)                    // or a supplied SignerCredential
    .allow(|audience| {
        alice.access().claim(...).delegate(audience)
    })
```

But then again allow supposed to be given everything until .delegate

notes/peer-and-session.md
```diff
@@ -0,0 +1,359 @@
...
+    .grant(certificate)                    // pre-minted, when the audience was known
+    .build()
+    .await?;
+
+alice.space("notes").open().perform(&job).await?;   // named space under the peer
```
Gozala: Maybe trying to use names here is unnecessary complexity, and instead it could be something like 

```rs
alice.repository('did:key:...').open().at(location).perform(&job).await?;
```

Original intention was to make repo opening universal across web and native, but I don't think it panned out anyhow

notes/peer-and-session.md
```diff
@@ -0,0 +1,359 @@
...
+`Peer::new` takes the storage rather than a `.storage()` step because the
+storage fixes the space type parameter and Rust cannot infer it from a
+later call.
+
+Planned (steps 4 and 5 below):
+
+```rust
+let tonk = alice.connect(did).await?;      // address book lookup; NoAddress otherwise
+let job = alice.session(b"refactor")
+    .using(branch)                         // extra proof layers, repeatable
```
Gozala: I feel like `alice.connect` perhaps should return a session so you would not need to derive one ?

Gozala: Relatedly maybe Peer could be an enum so session is just another constrained delegate peer as opposed to being it's own thing.

notes/peer-and-session.md
```diff
@@ -0,0 +1,359 @@
...
+    .build().await?;
+Repository::open("notes").perform(&job).await?;            // registry lookup, mounts
+let there = job.at(&tonk);                                 // bound env
```
Gozala: Also not sure why we need job.at(&tonk) as opposed to just `.perform(&tonk)`. Note that we can't even really derive peer for the remote peer who's keys we don't have

rust/dialog-peer/src/session/builder.rs
```diff
@@ -0,0 +1,197 @@
...
+use dialog_varsig::Principal as _;
+#[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+use dialog_varsig::Signer;
+
+const SESSION_DERIVATION_CONTEXT: &str = "dialog-db operator derivation";
+
+/// Where a session's acting key comes from.
+enum SessionKey {
+    /// Derived from the peer key and this context.
+    Derived(Vec<u8>),
```
Gozala: I think they both should be wrapping SignerCredential instead of using raw secret

notes/peer-and-session.md
```diff
@@ -0,0 +1,400 @@
...
+let peer = Peer::new()
+    .storage(Storage::<NativeSpace>::default())
+    .base(Directory::At(root))             // was OperatorBuilder::base
+    .network(Network::default())           // was OperatorBuilder::network
+    .branch("main")                        // was OperatorBuilder::access_branch
+    .open(Location::new(Directory::Profile, name))
+    .await?;
+let session = peer
+    .derive(b"app")                        // was profile.derive(b"app")
+    .await?
```
Gozala: I still think peer.session(b"app") would make more sense or maybe peer.attenuate(b"app") or maybe even peer.worker("b"app") and have it use builder syntax.

Gozala:
On the pending thread at `notes/peer-and-session.md` line 339 (`peer.session(b"app")` with builder syntax): done in 6a28135. `peer.session(key)` takes a `SessionKey`, made from bytes or a string for a derived key, or from a `SignerCredential` or bare signer for a supplied one, so `peer.session(signer)` is the same entry point. Derivation runs at `build`, or earlier through the builder's async `did` when a certificate has to be issued to the session before build. `Peer::derive` is gone and every call site here and in tonk reads `peer.session(ctx).allow(..).build()`.

Given the builder sketch discussed since, this method disappears again once a worker is built through the peer builder with `.credential(..)` and `.grant(..)`. It landed so the branch matches the review; the rework is the fold.

---

Gozala:
Ported onto the peers stack as #538, which also moves peer routing to the host's contacts.

